PR #93, opened 2021-02-09
Add Airflow BashOperator orchestration documentation

vijaykiran:


tombaeyens (approved):

tombaeyens:
As the next step, it would be great to let people choose between this bash operator and a python operator.  Would a python operator also be able to work in its own virtual env?

vijaykiran:
That’s something I’m investigating, the plan is to isolate as much as possible so we don’t clash with airflow dependencies

tombaeyens:
👌

PR #96, opened 2021-02-12
integration of soda CLI

milanaleksic:
This pull request allows Soda CLI to be tested as a container image. To do that we need to choose any possible http port and host, and turn off https on demand for the soda client.

Additionally, the time seems not to be included into the cloud API which was exposed during tryouts.

### DIFF
--- a/sodasql/cli/cli.py
+++ b/sodasql/cli/cli.py
@@ -8,12 +8,14 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import datetime
 import logging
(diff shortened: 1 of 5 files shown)


sodasql/cli/cli.py
```diff
@@ -195,7 +197,7 @@ def init(warehouse_file: str):
               help='Variables like -v start=2020-04-12.  Put values with spaces in single or double quotes.')
 @click.option('-t', '--time',
               required=False,
-              default=None,
+              default=datetime.datetime.now().isoformat(timespec='seconds'),
```
tombaeyens, line 200: We might postpone resolving the default value until back on the server.

tombaeyens (approved):

Issue #97, opened 2021-02-14
[bug] AWS Athena - error "TODO override and implement this abstract method"


lgov:
**Describe the bug**
I am trying to get soda-sql connect to some AWS Athena tables and stumble into an issue:

```
/Users/lgo/dev/soda-sql/venv/bin/python /Users/lgo/dev/soda-sql/venv/bin/soda init
  | 2.0.0b10
  | Initializing warehouse.yml ...
  | Querying warehouse for tables
  | Directory tables already exists
  | Exception: TODO override and implement this abstract method
Traceback (most recent call last):
  File "/Users/lgo/dev/soda-sql/venv/lib/python3.8/site-packages/sodasql/cli/cli.py", line 171, in init
    scan_initializer.initialize_scan_ymls()
  File "/Users/lgo/dev/soda-sql/venv/lib/python3.8/site-packages/sodasql/cli/scan_initializer.py", line 45, in initialize_scan_ymls
    self.warehouse.dialect.sql_tables_metadata_query())
  File "/Users/lgo/dev/soda-sql/venv/lib/python3.8/site-packages/sodasql/scan/dialect.py", line 109, in sql_tables_metadata_query
    raise RuntimeError('TODO override and implement this abstract method')
RuntimeError: TODO override and implement this abstract method

Process finished with exit code 0
```

Not sure why I end up here, AWS Athena support is supposed work in this beta right? Besides this stack track, it would be good if the cli return a non-zero exit code in case of a fatal error like this.

**Context**
My warehouse.yml:
```
name: athena
connection:
    type: athena
    database: xxxxx_reporting_db
    access_key_id: env_var(AWS_ACCESS_KEY_ID)
    secret_access_key: env_var(AWS_SECRET_ACCESS_KEY)
    role_arn:
    region: us-east-1
    staging_dir: s3://xxxxx-reporting-db/db/
```

My env_vars.yml:
```
athena:
  AWS_ACCESS_KEY_ID: 'xxxxxxxxxxxxxx'
  AWS_SECRET_ACCESS_KEY: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
```

**OS**: Mac OS X
**Python Version**: Python 3.8.6
**Soda SQL Version**: 2.0.0b10
**Warehouse Type**: athena

lgov:
Works now, so issue can be closed.

PR #98, opened 2021-02-14
In the AWS Athena dialect, implement function sql_tables_metadata_query

lgov:
This fixes issue #97 for me.


tombaeyens:
Hey Igov, Tnx for the suggestion!  We'll first check if we can do it consistent with other dbs using the information_schema tables.  If that doesn't work we ll apply your proposal.

lgov:
The branch conflicts because on master apparently the information_schema approach was already implemented. Not sure if I tried master HEAD first or not, will check.

lgov:
This change has been implemented in a different way already, so this PR is not needed anymore.

PR #112, opened 2021-02-17
Update docs to link to example files

vijaykiran:


dirkgroenen:
You might want to use full paths to the Github Repo, since relative paths will throw 404 errors on the docs website.

vijaykiran:
@dirkgroenen  good point, I was only thinking about GitHub README, I'll update the links

tombaeyens (approved):

PR #121, opened 2021-02-24
Validation warehouses

milanaleksic:
Add first iteration of Connection and Authentication validation checks for all warehouses

### DIFF
--- a/sodasql/dialects/postgres_dialect.py
+++ b/sodasql/dialects/postgres_dialect.py
@@ -8,6 +8,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import re
 
 import psycopg2
 
@@ -44,8 +45,8 @@ class PostgresDialect(Dialect):
         }
 
     def sql_tables_metadata_query(self, limit: str = 10, filter: str = None):
-        return (f"SELECT table_name \n" 
-                f"FROM information_schema.tables \n" 
+        return (f"SELECT table_name \n"
+                f"FROM information_schema.tables \n"
                 f"WHERE lower(table_schema)='{self.schema.lower()}'")
 
     def sql_connection_test(self):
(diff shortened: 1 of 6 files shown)


sodasql/dialects/postgres_dialect.py
```diff
@@ -94,10 +92,17 @@ def is_connection_error(self, exception):
         if exception is None:
             return False
         error_message = str(exception)
-        return error_message.find('Operation timed out') != -1
+        return error_message.find('Operation timed out') != -1 or \
```
vijaykiran, line 95: Are you sure you want to use string based error interpretation? This will fail if the locale is set to anything other than EN, I think. I would guess checking the proper sql state error code should be better. Assuming we are using PEP

milanaleksic: We will be checking this only in managed scan flow, where we are in control of the locale. 

It is a good remark, can you point me to the error code listing, I tried to find something which is a cross-driver standard but failed to find it

vijaykiran: The drivers should return the SQLSTATE error codes according to the standard. (https://en.wikipedia.org/wiki/SQLSTATE).

In the case of psycopg2 the exceptions/errors are in https://www.psycopg.org/docs/errors.html

milanaleksic: No they don't unfortunately, I am aware of that standard and already looked at this previously, I assumed there is something you could recommend that would work for psycopg2. Here is an example of raw exception (not a wrapper) of the psycopg2 error for connection/authentication (every time it is `OperationalError` without any properties/attributes - the only difference is in error message).

image

milanaleksic: BTW, the place the exception is thrown is in binary library `_psycopg.cpython-<python_version>m-<arch>.so` so I can't get any more information about it

milanaleksic:
@vijaykiran I am going to close this to not block further work and merging of integration tests, but let's revisit this again if opportunity for improvement comes

PR #130, opened 2021-03-01
Fixes #126 Make sure columnName is added to measurement

### LINKED ISSUE #126: [bug] Column isn't set when flushing sql_metric measurements
**Describe the bug**

I'm having this scan-yml which contains custom SQL metrics on a column level. I'm expecting those custom SQL metrics to generate Custom Metric Measurements which are associated with the belonging column, but when measurements are flushed the column isn't send to the platform.

See the following snippet, where you can see that `sodaSQlScanMeasurements` doesn't contain the column, even though it does belong to `O_ORDERSTATUS`. 

```
 | Executing SQL query: 
SELECT
  COUNT(CASE WHEN O_ORDERSTATUS = 'P' THEN 1 END) as ORDER_P,
  COUNT(CASE WHEN O_ORDERSTATUS = 'O' THEN 1 END) as ORDER_O
FROM ORDERS

  | SQL took 0:00:00.107177
  | SQL metric O_ORDERSTATUS.sql_metric[0] ORDER_P -> 38543
  | Query measurement: ORDER_P = 38543
  | SQL metric O_ORDERSTATUS.sql_metric[0] ORDER_O -> 732044
  | Query measurement: ORDER_O = 732044
  | > /api/command {
  "type": "sodaSqlScanMeasurements",
  "scanReference": "92276c91-1afe-4b68-861f-6b4bba046ddf",
  "measurements": [
    {
      "metric": "ORDER_P",
      "value": 38543
    },
    {
      "metric": "ORDER_O",
      "value": 732044
    }
  ]
}
  | < 200 {
  "added": 2
}
```

**To Reproduce**
Execute a `soda scan` using the below `orders.yml` on a snowflake sample dataset. Make sure it's connected to the Soda Cloud so measurements are send. 

**Context**
`orders.yml`

```yml
table_name: ORDERS
metrics:
  - row_count
  - sum
  - missing_count
  - stddev
  - variance
  - avg_length
  - valid_count
  - distinct
  - histogram
  - mins
  - maxs
  - frequent_values
columns:
  O_TOTALPRICE:
    sql_metrics:
      - sql: |
          SELECT SUM(O_TOTALPRICE) as OPEN_TOTALPRICE FROM ORDERS WHERE O_ORDERSTATUS = 'O'
        tests:
          open_orders_price: OPEN_TOTALPRICE < 2000000

  O_ORDERSTATUS:
    valid_values:
      - O
      - F
    sql_metrics:
      - sql: |
          SELECT
            COUNT(CASE WHEN O_ORDERSTATUS = 'P' THEN 1 END) as ORDER_P,
            COUNT(CASE WHEN O_ORDERSTATUS = 'O' THEN 1 END) as ORDER_O
          FROM ORDERS
        tests:
          open_orders: ORDER_O == 0
          processing_orders: ORDER_P > 0

tests:
  dataset_size: row_count == 0

```

**OS**: Linux
**Python Version**: Python 3.8
**Soda SQL Version**: 2.0.0b15
**Warehouse Type**: Snowflake

vijaykiran:
- The columnName is missing in JSON which is sent to Soda Cloud API

### DIFF
--- a/sodasql/scan/scan.py
+++ b/sodasql/scan/scan.py
@@ -145,10 +145,9 @@ class Scan:
         # eg { 'colname': {'missing': 2, 'invalid': 3}, ...}
         column_metric_indices = {}
 
-        for column_name_lower in self.scan_columns:
+        for column_name_lower, scan_column in self.scan_columns.items():
             metric_indices = {}
             column_metric_indices[column_name_lower] = metric_indices
-            scan_column: ScanColumn = self.scan_columns[column_name_lower]
             column_name = scan_column.column_name
 
             if scan_column.is_missing_enabled:
@@ -227,8 +226,7 @@ class Scan:
             row_count_measurement = next((m for m in measurements if m.metric == Metric.ROW_COUNT), None)
             if row_count_measurement:
                 row_count = row_count_measurement.value
-                for column_name_lower in self.scan_columns:
-                    scan_column = self.scan_columns[column_name_lower]
+                for column_name_lower, scan_column in self.scan_columns.items():
                     column_name = scan_column.column_name
                     metric_indices = column_metric_indices[column_name_lower]
                     non_missing_index = metric_indices.get('non_missing')
@@ -259,10 +257,9 @@ class Scan:
         self._flush_measurements(measurements)
 
     def _query_group_by_value(self):
-        for column_name_lower in self.scan_columns:
+        for column_name_lower, scan_column in self.scan_columns.items():
             measurements = []
 
-            scan_column: ScanColumn = self.scan_columns[column_name_lower]
             column_name = scan_column.column_name
 
             if scan_column.is_any_metric_enabled(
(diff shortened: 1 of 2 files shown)


sodasql/scan/scan.py
```diff
@@ -440,12 +440,17 @@ def _run_sql_metric_default_and_run_tests(self, sql_metric: SqlMetricYml, resolv
 
         test_variables = self._get_test_variables(column_name_lower)
 
+        column_name = None
+        if column_name_lower:
+            scan_column = self.scan_columns[column_name_lower]
+            column_name = scan_column.column_name
```
tombaeyens, line 446: This looks like it is the right fix indeed.  2 comments:

1) There is a method below `_run_sql_metric_with_groups_and_run_tests` that would have to be updated in similar style/

2) I think it's better to pass the case sensitive column name or the scan_column as a parameter into the function.  If the scan column has both the column_name and the column_name_lower pecomputed than that might be best. But just passing the column_name (case sensitive) and recomputing the column_name_lower is good too.  In any case you will then have to revisit the places where the method is called.

tombaeyens (changes requested):

tombaeyens (approved):

Issue #52, opened 2021-01-20
Split soda-sql in separate modules


tombaeyens:
Folder structure to be discussed

Libraries use dashes in the names, no underscores.  Folder names as indicated below all use underscores.

```
+ src
    + sodasql (why was this again?)
        + setup.py
    + sodasql_cloud
        + setup.py
        + sodasql_cloud
        + tests
    + sodasql_scan
        + setup.py
        + sodasql_scan
        + tests
    + sodasql_cli
        + setup.py
        + sodasql_cli
        + tests
    + sodasql_airflow
    + connectors
        + sodasql_postgres
            + setup.py
            + sodasql_postgres
            + tests
        + sodasql_snowflake
        + sodasql_athena
        + sodasql_redshift
        + sodasql_bigquery
        + sodasql_spark
```

Also, before starting I would like to understand how we'll deal with versioning the individual libs and how this impacts the release process.

TODO investigate if dialects can be extracted.  Passing data from core lib to dialects is no prob.  But The dialects should not invoke methods on core stuff.  That would cause circular dependencies.  To be investigated before starting this.

dirkgroenen:
>   + sodasql (why was this again?)

Because you can then refer to it as a _meta package_ in for example your `5 min tutorial`. It would basically install a default subset of modules, like eg the postgres connector.  

You can go without the package but then your `5 min tut` would state something like: 

> Install `pip install soda-sql-base soda-sql-postgres-connector soda-sql-snowflake-connector ...`

vs

> Install `pip install soda-sql`, which installs the `base` package along with for example the `postgres` connector.

vijaykiran:
I think the DX should be more like:
`pip install soda-sql[all]` to install everything and `pip-install soda-sql[posgres]` or `pip install soda-sql[snowflake]' etc. The AFAIK we can do this using using setuptools extras.

milanaleksic:
My 2 cents: I am asking myself who would need ever to download all the warehouse supports? I thinks that’s a degenerate case that we should maybe not optimise for by doing it with `pip install soda-sql`. I agree with Vinjay on this: perhaps better to leave soda-sql as core dependency only and as people to fetch warehouse they wish to use.

dirkgroenen:
As discussed during Engineering Meeting:

- API plug-able approach (inspiration can be taken from Mopidy or many other packages)
- Keep versions aligned in each library as long as all code is in the same repository. Any release creates a new release for each package in the repo.

[...]

PR #133, opened 2021-03-02
Pyodbc add new requirements

milanaleksic:
Updates the list of expected components on the Linux systems before soda-sql installation would be tried out.

@dirkgroenen I think you've added previously `openssl-dev` and `libffi-devel` but I am not sure if they are needed since docker image can execute scans (on postgresql database at least) without having these installed. Is there a way to quickly verify they are still needed? I didn't even have to install `libpq-dev` on docker image `python:3.8.5-slim-buster` and again soda-sql seems to be working OOTB.

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -21,8 +21,8 @@ The soda-sql CLI requires the following dependencies to be installed on your sys
 - Pip >=21.0
 - postgresql-libs (`libpq-dev` in Debian/Ubuntu, `libpq-devel` in CentOS, `postgresql` on MacOSX)
 - _Linux only:_ 
-  + On Debian Buster: `g++`, `unixodbc-dev`, `python3-dev`, `libssl-dev` and `libffi-dev`
-  + On CentOS 8: `gcc-c++`, `unixODBC-devel`, `python38-devel`, `libffi-devel` and `openssl-devel`
+  + On Debian Buster: `apt-get install g++ unixodbc-dev python3-dev libssl-dev libffi-dev`
+  + On CentOS 8: `yum install gcc-c++ unixODBC-devel python38-devel libffi-devel openssl-devel`
 
 To check your version of python, run the `python` command
 ```
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -20,7 +20,7 @@ The soda-sql CLI requires the following dependencies to be installed on your sys
 - Python >=3.7 <3.9
 - Pip >=21.0
 - postgresql-libs (`libpq-dev` in Debian/Ubuntu, `libpq-devel` in CentOS, `postgresql` on MacOSX)
-- _Linux only:_ `libssl-dev` and `libffi-dev` (`libffi-devel` and `openssl-devel` in CentOS)
+- _Linux only:_ `g++`, `unixodbc-dev`, `python3-dev`, `libssl-dev` and `libffi-dev` (`libffi-devel` and `openssl-devel` in CentOS)
 
 To check your version of python, run the `python` command
 ```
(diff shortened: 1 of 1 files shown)


dirkgroenen:
> @dirkgroenen I think you've added previously openssl-dev and libffi-devel but I am not sure if they are needed since docker image can execute scans (on postgresql database at least) without having these installed. Is there a way to quickly verify they are still needed? I didn't even have to install libpq-dev on docker image python:3.8.5-slim-buster and again soda-sql seems to be working OOTB.

I remember having to install those in order to successfully install soda-sql (since it was part of some of the library's dependencies). I'm using Arch as a reference, which is way neater than debian/ubuntu - maybe it's not required to be installed since it's already there by default? 

Maybe we can check if it's required by performing in installation in a slimmer image, like arch of alpine?

milanaleksic:
> arch of alpine

Do we want to support arch as a non-mainstream release? 

I would definitely not support alpine since we depend on many c/cpp deps which will be a nightmare to guarantee would work (https://pythonspeed.com/articles/base-image-python-docker-images/)

docs/getting-started/installation.md
```diff
@@ -20,7 +20,9 @@ The soda-sql CLI requires the following dependencies to be installed on your sys
 - Python >=3.7 <3.9
 - Pip >=21.0
 - postgresql-libs (`libpq-dev` in Debian/Ubuntu, `libpq-devel` in CentOS, `postgresql` on MacOSX)
-- _Linux only:_ `libssl-dev` and `libffi-dev` (`libffi-devel` and `openssl-devel` in CentOS)
+- _Linux only:_ 
+  + On Debian Buster: `apt-get install g++ unixodbc-dev python3-dev libssl-dev libffi-dev`
+  + On CentOS 8: `yum install gcc-c++ unixODBC-devel python38-devel libffi-devel openssl-devel`
```
vijaykiran, line 26: We probably need to point to https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/install-microsoft-odbc-driver-sql-server-macos?view=sql-server-ver15 for macOS here, unixodbc wasn't working with SQLServer

dirkgroenen:
> Do we want to support arch as a non-mainstream release?

@milanaleksic I woud tend to say _no_. People using Arch imo are able to understand how to resolve missing dependencies and connect the dots when a lib doesn't install because of one.

docs/getting-started/installation.md
milanaleksic: I think that page is a must read if they are actually trying out MSSQL in runtime, right? I think this above is enough to do a `pip install soda-sql`, but MSSQL will not work until instructions on that page are followed?

dirkgroenen:
This discussion really triggers me to favor #52 to be implemented. @vijaykiran @milanaleksic If there's any time left we should see if it makes sense to start thinking about https://github.com/sodadata/soda-sql/issues/52...

tombaeyens:
#52 if after the cloud launch the most important one.

I still need to digg deeper to understand the full details of this.  But it has me concerned.  If there is any way we can simplify installation and getting started, that is crucial.  And these libraries can be a pain.  So if at all possible we should ensure that users don't have to digg into these details.  But I guess we're already on the same page on that topic.

vijaykiran:
> This discussion really triggers me to favor #52 to be implemented. @vijaykiran @milanaleksic If there's any time left we should see if it makes sense to start thinking about #52...

I agree - better to start sooner than later on this, but it would also mean design review.

docs/getting-started/installation.md
vijaykiran: Yes, the install will run but scan/analyze will fail. Not sure if that gives a reasonable error message. For me it didn't work until I installed odbc driver, even though pyodbc claimed otherwise.

dirkgroenen:
> I agree - better to start sooner than later on this, but it would also mean design review.

Maybe you @vijaykiran and @tombaeyens can get in touch to see if you can start working on this in between other things you're doing?

docs/getting-started/installation.md
milanaleksic: Alright, so if it’s indeed runtime then we should add here “for MSSQL you need to follow this link”, and not ask each user to install those things needs for the MSSQL, agreed?

PR #149, opened 2021-03-04
bugfix: filter at the wrong place of histogram query

mmigdiso:
In case of histogram queries, filter is already passed to the CTE.
Adding the filter to the query using the CTE as input cause an error as non of the columns given in the filter may exists in select statement.

### DIFF
--- a/sodasql/scan/scan.py
+++ b/sodasql/scan/scan.py
@@ -391,9 +391,6 @@ class Scan:
                            f'  {fields} \n'
                            f'FROM group_by_value')
 
-                    if self.filter_sql:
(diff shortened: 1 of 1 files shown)


mmigdiso:
Integration tests are failing bc of the credentials.. (as I don't have the credentials as a github secret on my fork)

vijaykiran:
Thank you, I think we still need to update GitHub workflow to make sure that CI works on forked repo PRs.

tombaeyens:
Thanks, Murat!

vijaykiran:
@all-contributors please add @mmigdiso  for code

Issue #155, opened 2021-03-08
Soda should support DQ on Hive Tabls


mmigdiso:
**Is your feature request related to a problem? Please describe.**
Today most of our transformations are happening on Hive and I believe that this is common in most companies. 
The existing codebase does not support any DQ on Hive tables. 

**Describe the solution you'd like**
There are a couple of approaches for this. The easiest would be using  pyHive. 
This would enable companies using Presto/Hive run DQ checks on their data lake. 
It seems pretty straightforward to implement this in the existing code structure. A HiveDialect(with couple of overwritten methods) class would probably be enough. 

Another approach would be using spark as the computation engine. There might be some cases where the companies might want to run custom pyspark code against their data for verification(ex: Running a ml model against the dataframe). 
This can be done by using apache Livy which supports creating interactive spark sessions and run pyspark code. 

I believe that submitting spark jobs directly to the cluster would not be an efficient approach as it would spin a spark app for each query and would require reserving containers (which might take more time than the DQ check itself)

The current version of the livy does not fully support spark on k8s but there is an open PR https://github.com/apache/incubator-livy/pull/249
 

**Additional context**

**Soda SQL Version**:

PR #165, opened 2021-03-10
#155 - Adding Hive dialect

ilhamikalkan:
Hive dialect supports hive version < 3.x. It would be good to improve it for the next versions in the future.


vijaykiran:
Thank you for this contribution @mmigdiso and @ilhamikalkan ! Can you both please confirm if you signed the CLA?

sodasql/scan/scan.py
```diff
@@ -112,7 +112,7 @@ def execute(self) -> ScanResult:
 
     def _query_columns_metadata(self):
         sql = self.warehouse.dialect.sql_columns_metadata_query(self.scan_yml.table_name)
```
mmigdiso, line 114: @vijaykiran I think we have a design decision to make here. The existing code assumes that there will be no need to do any transformation for getting the columns. 
This might not always be true. I believe that it should be the dialects responsibility to get the columns. The dialect can do anything that is needed to get the columns and convert the result to the coulmn tuples.

mmigdiso (changes requested):

sodasql/dialects/hive_dialect.py
```diff
@@ -0,0 +1,123 @@
...
+        if exception is None:
+            return False
+        return isinstance(exception, Error)
+
+    def is_authentication_error(self, exception):
+        if exception is None:
+            return False
+        return isinstance(exception, TTransportException)
+
+    def try_to_raise_soda_sql_exception(self, exception: Exception) -> Exception:
```
vijaykiran, line 117: This override is probably not necessary - right?

sodasql/scan/scan.py
vijaykiran: Indeed, I can see Hive is a special case :) This would change the `Dialect`s responsibility a bit. For the first cut, I think this approach you have (below) seem okay to me.

sodasql/dialects/hive_dialect.py
```diff
@@ -0,0 +1,123 @@
...
+                configuration={ key: str(value) for key, value in self.configuration.items() }, #https://github.com/jaegertracing/jaeger-client-python/issues/151
+                auth=None)
+            return conn
+        except Exception as e:
+            self.try_to_raise_soda_sql_exception(e)
+
+    def sql_columns_metadata(self, table_name:str):
+        # getting columns info from hive which version <3.x needs to be parsed
+        column_tuples = []
+        cur = self.create_connection().cursor()
```
vijaykiran, line 61: Can you confirm if this connection will be closed after the query? If not, can you call close at the end of this function to make sure that it is cleaned up?

vijaykiran (changes requested):

vijaykiran:
@mmigdiso  and @ilhamikalkan  Thanks a lot for this contribution! I left a few comments for you to review and a few more notes:

- Can you please add the experimental connection settings in https://github.com/sodadata/soda-sql/blob/main/docs/documentation/warehouse_types.md ?
- Also, can you format the files using PEP8?

sodasql/dialects/hive_dialect.py
ilhamikalkan: done

sodasql/dialects/hive_dialect.py
ilhamikalkan: yes. I removed it from hive_dialect.py

vijaykiran (approved):

PR #199, opened 2021-03-20
closes #198: bump urllib3 version, remove jinja2

### LINKED ISSUE #198: Update version of urllib3
dependabot indicates that we have a vulnerability in urllib3 https://github.com/sodadata/soda-sql/security/dependabot/.github/scripts/requirements.txt/urllib3/open

i ve added `urllib3>=1.26.4` to the requirements.in

that leads to problems when recreating the venv:

```
Could not find a version that matches urllib3!=1.25.0,!=1.25.1,<1.26,>=1.20,>=1.21.1,>=1.26.4 (from -r requirements.in (line 12))
Tried: 0.3, 1.0, 1.0.1, ..., 1.26.4, 1.26.4
There are incompatible versions in the resolved dependencies:
  urllib3>=1.26.4 (from -r requirements.in (line 12))
  urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1 (from requests==2.23.0->snowflake-connector-python==2.4.0->-r requirements.in (line 4))
  urllib3<1.26,>=1.20 (from botocore==1.18.18->PyAthena==2.2.0->-r requirements.in (line 6))
```

requests has newer version 2.25.1
snowflake-connector-python has a newer version 2.4.1
botocore has a newer version 1.20.33

but

PyAthena only has 2.2.0

So for now I don't know how to fix this.  Anyone a suggestion?

milanaleksic:


tombaeyens (approved):
@milanaleksic Do you have any idea why this seems to have worked for you and that it didn't work for me?

Afaict I did just the same, but ended up with conflicts and I could not recreate the virtual env

milanaleksic:
> @milanaleksic Do you have any idea why this seems to have worked for you and that it didn't work for me?

I think you have been fixing it in `soda-sql/requirements.txt`, and I fixed it in `.github/scripts/requirements.txt`. If you need to change github CI Actions workflows then you need to create separate venv for `.github/scripts/requirements.txt` because those requirements are kept separate from the main application's requirements

PR #220, opened 2021-03-30
Updated How it Works with more monitor history detail.

janet-can:


docs/documentation/concepts.md
```diff
@@ -99,7 +99,7 @@ Additionally, you can integrate Soda SQL with a **Soda Cloud** account. This fre
 Though you do not have to set up and ingrate a Soda Cloud account in order to use Soda SQL, the web app serves to complement the CLI tool, giving you a non-CLI method of examining data quality. Use Soda Cloud to:
 
 - collaborate with team members to review details of scan results that can help you to diagnose data issues
-- store [metrics]({% link documentation/sql_metrics.md %}) over time
+- use monitors to view stored [scan output]({% link documentation/scan.md %}#scan-output) as a line graph that represents the volume of failed tests in each scan
```
tombaeyens, line 102: Not 100% sure, but I think it's good if we can highlight a bit more the storing of metrics over time.  

For engineers which are imo the primary target for this, they get the scan and testing pretty quickly.  Then the next thing they want to do is store their metrics over time.  We should make sure they understand that Soda Cloud is giving this to them.  Just today we had someone explain how they want to build up their own database of the test results over time.  The main mental model that we should plant in readers heads is a) Soda SQL scans calculate a set of metrics after each batch of data is produced.  and b) Soda Cloud can store and visualize all the information for my engineering use cases.  I don't need to build anything myself to store metrics over time and calculate change over time derivative metrics etc and c) Soda Cloud allows me even to collaborate with others.

The b) part is something that engineers might know how to build themselves on top of Soda SQL.  And they might be not expecting that a company offer this out of the box.  That's why I think it deserves some extra highlighting.  Also alternative solutions don't have this service so all the more reason to put some emphasis on it.

So that part should maybe even deserve a more prominent place than this bullet item.  Just throwing out the idea...  Dunno what the best way is to include it into the docs

tombaeyens (approved):

PR #224, opened 2021-03-31
Start using new sodaSql endpoints for some calls on the cloud API

milanaleksic:


Antoninj (approved):
Are we going to deprecate the old endpoints on the backend?
It's probably time to start implementing soda-sql version checks on both soda-sql and the backend.

milanaleksic:
> Are we going to deprecate the old endpoints on the backend?
> It's probably time to start implementing soda-sql version checks on both soda-sql and the backend.

Yes, but I can't do it immediately, we will have to wait for a month or two (or until we publish non-beta), since the current users of the API calls would otherwise be blocked

PR #226, opened 2021-03-31
Type qualification

tombaeyens:


sodasql/dialects/athena_dialect.py
```diff
@@ -61,7 +61,7 @@ def create_connection(self):
         return conn
 
     def is_text(self, column_type: str):
-        return not self.is_complex(column_type) and super().is_text(column_type)
+        return column_type.upper() in ['CHAR', 'VARCHAR', 'STRING']
```
vijaykiran, line 64: Postgres also has TEXT type (http://sqlfiddle.com/#!17/a31724/2)

PR #230, opened 2021-04-01
reword output of `create` to point to `analyze`

bastienboutonnet:
# Description
Since we renamed the `init` task to `analyse` we might as well incentivise people to use that one directly instead to prevent them getting any future bumps with deprecations and whatnot


vijaykiran:
@all-contributors please add @bastienboutonnet for code

Issue #136, opened 2021-03-02
[enhancement] It would be helpful to provide a filter(or custom query) to analyze a table


mmigdiso:
Currently soda analyze works on a small dataset to create the analyzer templates and it does it by applying a LIMIT clause to the table. 
Running analysis on a limited dataset might not always be a good idea. Let's say that I have a table containing data from 2010 and over 11 years some columns have changed the format. For example, the user_id was a hash code and we changed it to uuid.
Limit 1000 will always give the earliest 1000 rows (not guaranteed as a RDBMS standard but it is the way how it works on redshift). Extracting patterns based on the data that was generated 1000 years ago might not be accurate.

It would be very helpful to provide a way to let the data owner provide a query to be used for sampling. 
something like the following option in warehouse.yml would do the job. 

```
connection:
  type: redshift
  host:  HOST
  username: env_var(REDSHIFT_USERNAME)
  password: env_var(REDSHIFT_PASSWORD)
  region:  XXXX
  database: db
  schema: schema1
  analyze_templates:
    table1: "select * from table1 where id > 1000 limit 10"
```

tombaeyens:
Some databases also have sampling capabilities in SQL like snowflake https://docs.snowflake.com/en/sql-reference/constructs/sample.html  
Eg `select * from testtable tablesample bernoulli (20.3);`
IIRC there are other warehouses supporting the same or similar.

tombaeyens:
If we support table level configurations they should go in the scan YAML file.
A generic analyze filter or such config should go in the warehouse file.

I suggest we make an inventory first of which databases we know that still do a full table scan when `LIMIT 1000` is specified.  And for each of those (if we know) what the best strategy to deal with this.

tombaeyens:
Pasting discussion for reference:

> murat migdisoglu Yesterday at 9:52 AM
> I got another question related to the analyze phase: When I run soda analyze, the DatasetAnalyzer.analyze (dataaset_analyzer.py) runs some queries using a limit 10000 statement.
> But when I check the query plan for the query being executed, It executes the limit on top of the result of the query. In other words, I executes  COUNT(CASE WHEN) queries using a full scan, calculates the result and applies limit on top of it.
> This is a very costly operation for big tables
> XN Limit  (cost=243595200.05..243595200.10 rows=1 width=31)
>   ->  XN Aggregate  (cost=243595200.05..243595200.10 rows=1 width=31)
[...]

PR #242, opened 2021-04-06
#136 Support providing a filter (or custom query) which is used when analyzing a table

vtascioglu:
Supported the feature with adding configuration to warehouse.yml file analyze_templates section. Its totally optional and analyze template script has defaults for all dialects.

https://github.com/sodadata/soda-sql/issues/136

@vijaykiran


tombaeyens:
@vtascioglu @mmigdiso 

Thanks for sharing!  Good input

I m curious to learn a bit more about background of this use case:

Do you plan to run this analyze multiple times on a warehouse?  If so, did you consider a merge strategy on how the previous scan YAML would be merged with the newly produced analyze results?

I do understand the need to add additional filtering for the analyze in order to speed the analyze for larger columns.  Apart form performance, is there another reason that you want to apply this template in your situation?  I'm asking because I saw in the test this filter `where COUNTRY=\'US\' limit 10"` which seems to be about more than just performance.  I m wondering if that is a real use case and if so how it's linked to the overall scans that will be done on it later.  Wouldn't it imply that the generated scan YAMLs are only applicable to US and not to other countries?

What would you think of a multi-step strategy?  
 * First generate the scan YAMLs
 * Then allow user to specify filters in the scan YAMLs
 * Then run the analyze on the filters in the scan YAMLs
That way the filters would be local to the tables and reusable.

We still have to think through the next steps.   Especially merging.  I think it makes sense to rerun the analyze command later on and merge the new results into the existing scan YAMLs.  Ideally the scan YAMLs are in source control. Then as a user you can review the changes before checking them in.  Any ideas, comments, thoughts on that are welcome

mmigdiso:
Hey @tombaeyens thank you for your comments. For us the most important use case is to run analyze on a recent set of data. The nature of the data changes over time. A customer id can become a uuid over time. they are both still a string but the  semantic is not same anymore, therefore the data quality rules to be applied. 
The biggest headache of the teams responsible of defining the dq rules is to manually define them. It is very time consuming to define dq checks for a table with 100 columns. Our surveys show that our stakeholders are mostly interested in auto generated rules, a smart data analyzer. But the first step of a smart data analyzer is to use a meaningful set of data as the input. 
With this PR, we are addressing this issue. 

Multi step approach can work but I'm not sure about the UX part of it. It will add extra complexity. Let me think it thoroughly and come up with a more sophisticated answer to it.

PR #246, opened 2021-04-06
Changes to add nav files to documentation

janet-can:
Rather than try to manage the navigation side-bar using numbers in the headers of docs files, I want to shift to using a nav.yml file in conjunction with a nav.html file.  This PR gets us most of the way there, but there are still a couple of adjustments needed.
- Adjust so that folder levels (Get Started, Documentation) expand and collapse when clicked, instead of doing nothing
- Widen the nav sidebar a bit to properly pad the contents and account for longer page titles
- Adjust any CSS to better match the current side-nav actions, styling, and font.

Any comments or feedback welcome! This will be a WAAAY better experience for organizing docs content.


janet-can:
For reference: http://planetjekyll.github.io/snippets/nav-with-data

docs/_data/nav.yml
```diff
@@ -0,0 +1,46 @@
...
+    subhref: "/soda-sql/documentation/monitors"
+  - subtitle: "Integrate with Slack"
+    subhref: "/soda-sql/documentation/integrate-slack"
+  - subtitle: "Troubleshoot"
+    subhref: "/soda-sql/documentation/troubleshoot"
+  - subtitle: "Glossary"
+    subhref: "/soda-sql/documentation/glossary"
+
+- title: "Community"
+  href: "/soda-sql/community"
```
vijaykiran, line 46: You might want to change your editor to add empty new lines at the end of the file :)

dirkgroenen (commented):
@janet-can I went ahead and resolved some of your "todo" items as per requested in the email. Some changes I made in https://github.com/sodadata/soda-sql/pull/246/commits/95460d9b63c51b7eb4ee7dcb57857b106f4ea407:

- Use references to pages (`page`) instead of hard-coded links in your `nav.yml`. This should prevent any dead links, as well as prevent us from having to go through each link when making a structural change.  (We can still support 'absolute links' as well, if you would like).
- Adjusted the rendered template to re-use the template's classes. This gives the look and feel you're looking for. 
- Widen up the nav bar a little to make sure all items stay on a single line.
- Toggle the sub-nav when one of its children is active (or when clicking on the array icon)

dirkgroenen:
Closes https://github.com/sodadata/soda-sql/issues/205

vijaykiran:
@janet-can  I guess this PR on this repo is not needed anymore right ? Since we moved docs to https://github.com/sodadata/docs ?

PR #285, opened 2021-04-27
#284: validate scan time ISO 8601 compliance

### LINKED ISSUE #284: Scan time should be ISO-8601 compliant
There isn't any validation done on the scan time property at the moment.  As such, we can not be sure the cli user input is ISO-8601 compliant.

We should probably try to parse and validate it so that non compliant dates are rejected.

**Soda SQL Version**:
v2.1.0b2

Antoninj:
closes #284


core/sodasql/common/time_helper.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class TimeHelper:
+
+    @staticmethod
+    def datetime_valid(date: str):
+        try:
+            datetime.fromisoformat(date)
+        except ValueError as e:
+            logging.error(f'The scan time is not in ISO8601 format: {str(e)}')
```
milanaleksic, line 22: To avoid confusion about which date we are talking about here, can you include the input value `date`? Or is it maybe already included in the error message?

Antoninj: It's part of the error message indeed. 

Example: `ERROR:root:The scan time is not in ISO8601 format: Invalid isoformat string: '2021-04-15T09:00:00+0200'`

core/sodasql/cli/cli.py
```diff
@@ -349,6 +350,8 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple = None, t
         scan_builder = ScanBuilder()
         scan_builder.warehouse_yml_file = warehouse_yml_file
         scan_builder.scan_yml_file = scan_yml_file
+        if not TimeHelper.datetime_valid(time):
```
vijaykiran, line 353: Are you sure we need an extra class for this functionality here? Since we catch all the exceptions anyway, this can be just 
```python
datetime.fromisoformat(time)
```

In case of exception it will print exception:
```
soda scan -t blah warehouse.yml tables/demodata.yml
  | 2.1.0b2
  | Scan failed: Invalid isoformat string: 'blah'
Traceback (most recent call last):
  File "/Users/vijay/.pyenv/versions/3.9.4/envs/soda-sql-main/lib/python3.9/site-packages/sodasql/cli/cli.py", line 353, in scan
    datetime.fromisoformat(time)
ValueError: Invalid isoformat string: 'blah'
  | Exiting with code 1
```

vijaykiran (changes requested):
I think we can get away without adding an additional class for this - can you please check my comment?

core/sodasql/cli/cli.py
vijaykiran: Oh, also `datetime.isoformat` doesn't support `Z` postfix. 

Either we need to either update the help value that says `The scan time in ISO8601 format like eg 2020-12-31T16:48:30Z`  or  update the code to treat `Z` as `+00:00`

Antoninj: @vijaykiran I've just pushed your suggested changes. I went with updating the help text for now, we can probably decide later if we want to treat `Z` as `+00:00`.

vijaykiran (approved):

PR #342, opened 2021-06-09
Adding event dispatch for two depending repo's

jmarien:
Dispatch an event to two internal repositories which may then decide to start a new build with latest soda-sql.


milanaleksic (approved):

vijaykiran:
@all-contributors please add @jmarien for infra

PR #357, opened 2021-06-18
Add docker compose for spark

JCZuurmond:
Adds a docker-compose file for running Spark. To be used for testing Soda spark.


JCZuurmond:
I was able to run it locally with the script. It is intended to be used in the CI for testing the spark dialect.

PR #360, opened 2021-06-22
Fix is_number detection in Athena

Tonkonozhenko:
The next query returns REAL type even if a table has float/double type.

```SELECT column_name, data_type, is_nullable 
FROM information_schema.columns 
WHERE table_name = '...' 
  AND table_schema = '...';
  ```


vijaykiran:
Thank you @Tonkonozhenko !

PR #381, opened 2021-07-07
make the recreate_venv.sh script work again

milanaleksic:
Script `scripts/recreate_venv.sh` stopped working at some moment in time.

I have also noticed that the `clean_install.sh` is not operational any more as well:

```
➜ scripts/clean-install.sh
Found existing installation: soda-sql-core 2.1.0b11
Uninstalling soda-sql-core-2.1.0b11:
  Successfully uninstalled soda-sql-core-2.1.0b11
python: can't open file 'setup.py': [Errno 2] No such file or directory
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
```

But I am not sure I understand in which flow should it be used so I decided not to try to fix that one as well

### DIFF
--- a/scripts/recreate_venv.sh
+++ b/scripts/recreate_venv.sh
@@ -13,8 +13,6 @@ python3 -m venv $VENV_DIR
 source $VENV_DIR/bin/activate
 pip install --upgrade pip
 pip install "$(cat dev-requirements.in | grep pip-tools)"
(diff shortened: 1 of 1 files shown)


vijaykiran:
@all-contributors please add @milanaleksic  for infra

PR #390, opened 2021-07-13
Update default athena duration

vijaykiran:


dirkgroenen:
Be aware that this change causes issues in some environments, where the request duration exceeds the max duration. See https://github.com/sodadata/platform-feedback/issues/44#issuecomment-879178849

PR #397, opened 2021-07-15
Update athena_dialect.py

Tonkonozhenko:


packages/athena/sodasql/dialects/athena_dialect.py
```diff
@@ -23,14 +23,13 @@ def get_aws_credentials_optional(parser: Parser):
         access_key_id = parser.get_str_optional_env('access_key_id')
         role_arn = parser.get_str_optional_env('role_arn')
         profile_name = parser.get_str_optional_env('profile_name')
-        if access_key_id or role_arn or profile_name:
```
vijaykiran, line 26: @Tonkonozhenko  Since we need at least some of these, isn't it better to leave the if condition, and log error on else?

vijaykiran (changes requested):
Thank you for the PR, can you check my comment?

packages/athena/sodasql/dialects/athena_dialect.py
Tonkonozhenko: We don't need any of these if we use IRSA, instance profile or default profile, that's why I removed it

Tonkonozhenko:
@vijaykiran could you check my comment above please?

vijaykiran:
@all-contributors please add @Tonkonozhenko for code

vijaykiran:
@all-contributors please add @Tonkonozhenko for code

PR #407, opened 2021-07-29
Issue/402

fakirAyoub:
Being able to validate numerics and arrays:
'valid_min' can only have an int as a value, or else an exception is raised before the scan.
'valid_values' can only be an array, or else an exception is raised before the scan.

### DIFF
--- a/core/sodasql/common/yaml_helper.py
+++ b/core/sodasql/common/yaml_helper.py
@@ -14,10 +14,10 @@ class YamlHelper:
 
     @staticmethod
     def validate_numeric_value(column_name, key, value):
-        if column_name is None:
+        if value is None:
             print("column none")
             logging.info(f'There is no {column_name} key to be validated')
-        elif value is None or not isinstance(value, int):
+        elif value is not isinstance(value, int):
             logging.error(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
             raise Exception(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
         else:
--- a/core/sodasql/common/yaml_helper.py
+++ b/core/sodasql/common/yaml_helper.py
@@ -17,7 +17,7 @@ class YamlHelper:
         if value is None:
             print("column none")
             logging.info(f'There is no {column_name} key to be validated')
-        elif value is not isinstance(value, int):
+        elif value is isinstance(value, int):
             logging.error(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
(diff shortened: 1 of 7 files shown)


core/sodasql/common/yaml_helper.py
```diff
@@ -11,3 +11,23 @@ def parse_yaml(yaml_str: str, description: str = None):
             return yaml.load(yaml_str, Loader=yaml.SafeLoader)
         except Exception as e:
             logging.error(f'Parsing YAML failed: {str(e)}: ({description if description else yaml_str})xWW4')
+
+    @staticmethod
+    def validate_numeric_value(column_name, key, value):
+        if value is None:
+            logging.info(f'There is no {key} key of this column name {column_name} to be validated')
```
vijaykiran, line 18: Small style change here:
I'd write it as : `There is no value specified for {key} for column {column_name}`

core/sodasql/common/yaml_helper.py
```diff
@@ -11,3 +11,23 @@ def parse_yaml(yaml_str: str, description: str = None):
...
+        elif value is isinstance(value, int):
+            logging.error(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
+            raise Exception(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
+        else:
+            return value
+
+    @staticmethod
+    def validate_array_value(column_name, key, value):
+        if value is None:
+            logging.info(f'There is no {key} key of this column name {column_name} to be validated')
```
vijaykiran, line 28: Please see the comment above for the text

core/sodasql/common/yaml_helper.py
```diff
@@ -11,3 +11,23 @@ def parse_yaml(yaml_str: str, description: str = None):
...
+        elif value is isinstance(value, int):
+            logging.error(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
+            raise Exception(f'{column_name} could not be parsed: {key}-{value} is not of a numeric type.')
+        else:
+            return value
+
+    @staticmethod
+    def validate_array_value(column_name, key, value):
```
vijaykiran, line 26: maybe name this `validate_list_value` ? Since python nomenclature prefers `list` than `array`?

vijaykiran (changes requested):
Thank you @fakirAyoub ! Looking good, just two suggestions on naming and log message. LGTM!

core/sodasql/common/yaml_helper.py
fakirAyoub: you're right, sure!

vijaykiran (approved):

PR #411, opened 2021-07-30
Draft: dbt dialect

NielsZeilemaker:
Made this DBT dialect POC, and wanted to get your feedback on it. Basically, it uses the information in your dbt profile to make a connection to the database of your choice (only tested with postgres). But you could extend this approach to also start parsing the dbt models etc.

Again, just had some time, so made this +/- working, and would really appreciate feedback on the direction.


vijaykiran:
Hi Niel! Thank you for the proposal, I'll think about it a bit and get back to you soon :)

PR #416, opened 2021-08-03
Issue/405

fakirAyoub:
Fixed common tests between MySQL and PostgreSQL
More Generic Regex without specific syntax
MySQL Dialect fixed
Some Unit tests are now embracing both mysql and postgresql specificities


vijaykiran:
@all-contributors please add @fakirAyoub for code

PR #439, opened 2021-08-10
sql_test_connection for all warehouses

fakirAyoub:
- Implement the `sql_test_connection` method for all warehouses
- Unit tests
- Small code refacto

### DIFF
--- a/packages/athena/sodasql/dialects/athena_dialect.py
+++ b/packages/athena/sodasql/dialects/athena_dialect.py
@@ -74,14 +74,6 @@ class AthenaDialect(Dialect):
             catalog_name=self.catalog)
         return conn
 
-    def __query_table(self, table_name):
-        query = f"""
-        SELECT *
-        FROM {table_name}
-        LIMIT 1
-        """
-        return query
-
     def sql_test_connection(self) -> Union[Exception, bool]:
         conn = self.create_connection()
         cursor = conn.cursor()
--- a/packages/athena/sodasql/dialects/athena_dialect.py
+++ b/packages/athena/sodasql/dialects/athena_dialect.py
@@ -9,6 +9,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 import re
+import logging
+from typing import Union
 from datetime import date
 import pyathena
 from sodasql.scan.dialect import ATHENA, KEY_WAREHOUSE_TYPE, Dialect
@@ -71,6 +73,29 @@ class AthenaDialect(Dialect):
             role_arn=self.aws_credentials.role_arn if self.aws_credentials else None,
             catalog_name=self.catalog)
         return conn
+
+    def __query_table(self, table_name):
+        query = f"""
+        SELECT *
+        FROM {table_name}
+        LIMIT 1
+        """
+        return query
+
+    def sql_test_connection(self) -> Union[Exception, bool]:
+        conn = self.create_connection()
+        cursor = conn.cursor()
+        tables = cursor.fetchall()
+        if tables:
+            for (table_name,) in cursor:
+                test_query = self.__query_table(table_name)
+                try:
+                    cursor.execute(test_query)
+                except Exception as e:
+                    raise Exception(f'Unable to query table: {table_name} from the database: {self.database}. Exception: {e}')
+        else:
+            logging.warning(f'{self.database} does not contain any tables.')
+        return True
 
     def is_text(self, column_type: str):
         column_type_upper = column_type.upper()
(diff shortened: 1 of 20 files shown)


packages/athena/sodasql/dialects/athena_dialect.py
```diff
@@ -72,6 +74,23 @@ def create_connection(self):
...
+    def __query_table(self, table_name):
+        query = f"""
+        SELECT *
+        FROM {table_name}
+        LIMIT 1
+        """
+        return query
+
+    def sql_test_connection(self) -> Union[Exception, bool]:
+        conn = self.create_connection()
```
vijaykiran, line 86: I think this should also be in try clause, because this can raise an exception

packages/bigquery/sodasql/dialects/bigquery_dialect.py
```diff
@@ -70,10 +70,13 @@ def __query_table(self, table_name):
         """
         return query
 
-    def sql_test_connection(self, dataset_id) -> Union[Exception, bool]:
+    def sql_test_connection(self) -> Union[Exception, bool]:
         logging.info(f'Listing tables to check connection')
```
vijaykiran, line 74: can your remove `f` from here :) following boy scout rule ;)

packages/bigquery/sodasql/dialects/bigquery_dialect.py
```diff
@@ -70,10 +70,13 @@ def __query_table(self, table_name):
...
+        project_id = self.account_info_dict['project_id']
+        dataset_id = f'{project_id}.{self.dataset_name}'
         try:
             tables = self.client.list_tables(dataset_id)
+            logging.info(f'dataset_id = {dataset_id}')
```
vijaykiran, line 79: Maybe you want the log before list_tables call or maybe even before try, because it won't print if list_tables raises an exception

packages/hive/sodasql/dialects/hive_dialect.py
```diff
@@ -58,6 +59,29 @@ def create_connection(self, *args, **kwargs):
...
 
+    def __query_table(self, table_name):
+        query = f"""
+        SELECT *
+        FROM {table_name}
+        LIMIT 1
+        """
+        return query
+
+    def sql_test_connection(self) -> Union[Exception, bool]:
```
vijaykiran, line 70: The return type doesn't look like what the function does, raising exceptions don't return them. So maybe you mean just bool here?

vijaykiran: This applies to other dialects too, btw :)

packages/redshift/sodasql/dialects/redshift_dialect.py
```diff
@@ -117,6 +118,33 @@ def __get_cluster_credentials(self):
...
 
+    def __query_table(self, table_name):
+        query = f"""
+        SELECT *
+        FROM {table_name}
+        LIMIT 1
+        """
+        return query
+
+    def sql_test_connection(self) -> Union[Exception, bool]:
```
vijaykiran, line 129: Do we need these functions here? RedshiftDialect extends PostgresDialect anyway 🤔

packages/athena/sodasql/dialects/athena_dialect.py
```diff
@@ -72,6 +74,23 @@ def create_connection(self):
             catalog_name=self.catalog)
         return conn
 
+    def __query_table(self, table_name):
```
vijaykiran, line 77: Looks like this can go into the  Dialect class, and we don't need to repeat it in every dialect.

vijaykiran (changes requested):
@fakirAyoub Looking nice! Added a few comments, please take a look.

packages/hive/sodasql/dialects/hive_dialect.py
vijaykiran: Some more context: https://www.python.org/dev/peps/pep-0484/#exceptions

fakirAyoub: right, it's either we return boolean or we raise an exception, so bool is enough!

packages/redshift/sodasql/dialects/redshift_dialect.py
fakirAyoub: good catch!

packages/athena/sodasql/dialects/athena_dialect.py
fakirAyoub: create_connection method takes care of this already, no?

fakirAyoub: All `create_connection` methods catch the connection exception: 

`except Exception as e:
            self.try_to_raise_soda_sql_exception(e)`

so if the exception is taken into account already

vijaykiran (approved):

PR #442, opened 2021-08-16
Adjust logging level with env var (#399)

### LINKED ISSUE #399: [core] Logging in Soda-sql
Currently Soda just import logging directly without logger name. So it's hard to changing logging level.

m1n0:
- Use env var to adjust the logging level.
- No tests added for this minor change

Fixes #399

### DIFF
--- a/core/sodasql/common/logging_helper.py
+++ b/core/sodasql/common/logging_helper.py
@@ -10,14 +10,26 @@
 #  limitations under the License.
 
 import logging
+import os
 import sys
 
 
 class LoggingHelper:
+    log_format = "  | %(message)s"
 
     @classmethod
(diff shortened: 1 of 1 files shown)


vijaykiran (approved):

toddy86:
Longer term, it might be also useful to name space the loggers in each module. That way, as a soda-sql user I can have fine grained control over which modules emit which log levels. So if I want to suppress the logs from just one module (which is the issue we have encountered), I can suppress those messages, but others still bubble up.  

```python 
# In module X
import logging

...
...
logger.info("I'm logged in the namespace")
```

vijaykiran:
@toddy86 Agreed, added an issue:  #445

PR #448, opened 2021-08-18
Add --offline flag to scan cli to skip sending results to Soda Cloud

### LINKED ISSUE #192: Add --offline flag to the CLI
Introduce a new _--offline_ or (_--dryrun_) optional flag on the CLI.

**Additional context**
Allow to run scans strictly offline for users that do not want to push data to Soda Cloud.

m1n0:
- Added offline flag with default False
- No test added as it would only work on the builder, which would be a pointless test testing that passed parameter is used
- Removed default values for the `scan` cli method as we are using click default values for those params anyway
- Fixes #192

### DIFF
--- a/core/sodasql/cli/cli.py
+++ b/core/sodasql/cli/cli.py
@@ -328,7 +328,12 @@ def analyze(warehouse_file: str, include: str, exclude: str):
               required=False,
               default=datetime.now(tz=timezone.utc).isoformat(timespec='seconds'),
               help='The scan time in ISO8601 format like eg 2021-04-28T09:00:00+02:00')
-def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple = None, time: str = None):
+@click.option( '--offline',
+              required=False,
+              is_flag=True,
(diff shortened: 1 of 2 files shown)


core/sodasql/cli/cli.py
```diff
@@ -340,6 +345,9 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple = None, t
     """
     logger.info(SODA_SQL_VERSION)
 
+    if offline:
+        logging.info('Running in offline mode, scan results will NOT be pushed to Soda Cloud.')
```
vijaykiran, line 349: Should be `logger.info`

vijaykiran (changes requested):
Just one minor change needed, otherwise LGTM

core/sodasql/cli/cli.py
m1n0: ah yes thanks for the catch, fixed

vijaykiran (approved):

PR #455, opened 2021-08-26
#359 add --non-interactive flag

fakirAyoub:
* When using --time, user will be asked for confirmation
* The flag defaults to False

### DIFF
--- a/core/sodasql/cli/cli.py
+++ b/core/sodasql/cli/cli.py
@@ -331,7 +331,7 @@ def analyze(warehouse_file: str, include: str, exclude: str):
               is_flag=True,
               required=False,
               help='Use this flag if you want to skip confirmations and run the scan.')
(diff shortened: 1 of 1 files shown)


janet-can:
Hi @vijaykiran and @fakirAyoub– 
I'm documenting non-interactive and when I test it locally, I'm getting a message in the CLI that I don't expect:
```shell
soda scan --non-interactive warehouse.yml tables/orders.yml
  | 2.1.0b15
  | You are using the --time option with the following value: 2021-09-01T20:57:30+00:00, meaning that the actual date of the scan is being altered manually.
Are you sure you wish to continue with the --time option? Press 'y' to continue... 
```
I didn't include a `--time` argument in my command so I suppose I'm not really sure what --non-interactive is doing. Could you advise, please and thank you?

PR #466, opened 2021-09-01
Implement option to limit the number of tables analyzed

### LINKED ISSUE #398: [snowflake] make the analyse query limit configurable
Snowflake analyse SQL query has a limit of 10 - https://github.com/sodadata/soda-sql/blob/main/packages/snowflake/sodasql/dialects/snowflake_dialect.py#L128

Change this needs to be configurable

vijaykiran:
Closes #398

### DIFF
--- a/core/sodasql/cli/cli.py
+++ b/core/sodasql/cli/cli.py
@@ -189,7 +189,12 @@ def matches_table_exclude(table_name: str, table_exclude_pattern):
 @click.option('-e', '--exclude',
               required=False,
               help='Table name exclusion filter, case insensitive, comma separated list, use * as a wild card')
-def analyze(warehouse_file: str, include: str, exclude: str):
+@click.option('-l', '--limit',
+              type=int,
+              required=False,
+              help='Limit the number of tables analyzed. This option is ignored for Hive and Spark dialects'
+              )
+def analyze(warehouse_file: str, include: str, exclude: str, limit: int):
     """
     Analyzes tables in the warehouse and creates scan YAML files based on the data in the table. By default it creates
     files in a subdirectory called "tables" on the same level as the warehouse file.
@@ -230,7 +235,7 @@ def analyze(warehouse_file: str, include: str, exclude: str):
         if hasattr(dialect, "sql_tables_metadata"):
             rows = dialect.sql_tables_metadata()
         else:
-            tables_metadata_query = dialect.sql_tables_metadata_query()
+            tables_metadata_query = dialect.sql_tables_metadata_query(limit=limit)
             rows = warehouse.sql_fetchall(tables_metadata_query)
 
         table_include_regex = create_table_filter_regex(include)
--- a/core/sodasql/scan/dialect.py
+++ b/core/sodasql/scan/dialect.py
@@ -47,6 +47,7 @@ ALL_WAREHOUSE_TYPES = [ATHENA,
 
 logger = logging.getLogger(__name__)
 
+
 class Dialect:
     data_type_varchar_255 = "VARCHAR(255)"
     data_type_integer = "INTEGER"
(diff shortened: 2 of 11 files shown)


core/sodasql/cli/cli.py
```diff
@@ -189,7 +189,13 @@ def matches_table_exclude(table_name: str, table_exclude_pattern):
 @click.option('-e', '--exclude',
               required=False,
               help='Table name exclusion filter, case insensitive, comma separated list, use * as a wild card')
-def analyze(warehouse_file: str, include: str, exclude: str):
+@click.option('-l', '--limit',
+              default=10,
```
m1n0, line 193: Why do we want a global default limit of 10? This sounds like a breaking change that will produce unexpected results for users using everything else than snowflake

vijaykiran: Good point.

This is not done yet - I'm adding limit option to all the dialects, and remove the default 10 on snowflake as well.

core/sodasql/scan/dialect.py
```diff
@@ -133,7 +134,7 @@ def create_connection(self):
     def sql_columns_metadata_query(self, table_name: str) -> str:
         raise RuntimeError('TODO override and implement this abstract method')
 
-    def sql_tables_metadata_query(self, limit: str = 10, filter: str = None):
+    def sql_tables_metadata_query(self, limit: int = None, filter: str = None):
```
m1n0, line 137: Shouldnt we mark these params with `Optional[X]` in case of passing `None` as default? 🤔 This is repeating in many places

packages/athena/sodasql/dialects/athena_dialect.py
```diff
@@ -98,12 +98,15 @@ def is_number(self, column_type: str):
...
+    def sql_tables_metadata_query(self, limit: int = None, filter: str = None):
         # Alternative ( https://github.com/sodadata/soda-sql/pull/98/files )
         # return (f"SHOW tables IN `{self.database.lower()}`;")
-        return (f"SELECT table_name \n"
-                f"FROM information_schema.tables \n"
-                f"WHERE lower(table_schema) = '{self.database.lower()}';")
+        sql = (f"SELECT table_name \n"
+               f"FROM information_schema.tables \n"
+               f"WHERE lower(table_schema) = '{self.database.lower()}'")
+        if limit is not None:
```
m1n0, line 107: `if limit` should be enough

packages/bigquery/sodasql/dialects/bigquery_dialect.py
```diff
@@ -84,9 +84,12 @@ def sql_test_connection(self) -> bool:
...
             raise Exception(f'Unable to list tables from: {dataset_id}. Exception: {e}')
 
-    def sql_tables_metadata_query(self, limit: str = 10, filter: str = None):
-        return (f"SELECT table_name \n"
-                f"FROM `{self.dataset_name}.INFORMATION_SCHEMA.TABLES`;")
+    def sql_tables_metadata_query(self, limit: int = None, filter: str = None):
+        sql = (f"SELECT table_name \n"
+               f"FROM `{self.dataset_name}.INFORMATION_SCHEMA.TABLES`")
+        if limit is not None:
+            sql += f"\n LIMIT {limit}"
```
m1n0, line 91: this limit part is repeating, perhaps it could be a Dialect method and we could do 
```
sql = ... " from ... {self.get_metadata_query_limit();"}
```
wdyt?
and the method could return None

packages/athena/sodasql/dialects/athena_dialect.py
vijaykiran: I'd prefer to be more 'explicit' :)

core/sodasql/scan/dialect.py
vijaykiran: Yup, I'll update the type hint

packages/bigquery/sodasql/dialects/bigquery_dialect.py
vijaykiran: I think for the first version we probably don't need an extra method in the base class .. yet. My thoughts are we can also make the entire query configurable, so the API is not finalised yet - there is also 'unused' filter option.

vijaykiran: some context : https://github.com/sodadata/soda-sql/pull/242

m1n0 (approved):

PR #475, opened 2021-09-04
Fix MySQL dialect issues

sintezcs:
- change the warehouse type `SQLSERVER` -> `MYSQL`
- fix escape sequences in regex
- fix the weird mysql-connector-python error `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 1: invalid start byte`
- fix the debug log (don't pass the exc_info object into the log message templated string)
- fix the test suite class name


vijaykiran (approved):

vijaykiran:
@all-contributors please add @sintezcs for code

PR #476, opened 2021-09-06
Add equality methods to data classes

JCZuurmond:
This equality method helps me to define tests in [`soda-spark`](https://github.com/sodadata/soda-spark/blob/main/tests/test_scan.py) like:

``` python
expected_measurement = Measurement(metric="row_count', column_name=None, value=6)

scan.execute()

assert any(expected_measurement == measurement for measurement in scan.scan_result.measurements
```

Where do I put tests for these methods?

### DIFF
--- a/core/sodasql/scan/measurement.py
+++ b/core/sodasql/scan/measurement.py
@@ -8,11 +8,15 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from dataclasses import dataclass
+from dataclasses import dataclass, astuple
 from typing import Optional, List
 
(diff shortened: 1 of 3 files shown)


core/sodasql/scan/measurement.py
```diff
@@ -8,13 +8,17 @@
...
-from dataclasses import dataclass
+from dataclasses import dataclass, astuple
 from typing import Optional, List
 
 from sodasql.common.json_helper import JsonHelper
 from sodasql.scan.group_value import GroupValue
 
 
+def is_equal_or_both_none(left, right):
+    return (left == right) or (left is None and right is None)
```
vijaykiran, line 19: I think  `left == right` should suffice to test for both `None` .. is there a reason for `or (left ...)` ?

JCZuurmond: works! so no need for this

PR #481, opened 2021-09-09
Add pyodbc connection method, with Databricks connection

JCZuurmond:
Heavily inspired on https://github.com/dbt-labs/dbt-spark/blob/master/dbt/adapters/spark/connections.py

### DIFF
--- a/core/sodasql/dataset_analyzer.py
+++ b/core/sodasql/dataset_analyzer.py
@@ -66,7 +66,7 @@ class DatasetAnalyzer:
                     qualified_regex = dialect.qualify_regex(format_regex)
                     regexp_like = dialect.sql_expr_regexp_like(
                         qualified_column_name, qualified_regex)
-                    count_field = f'COUNT(CASE WHEN {regexp_like} THEN 1 END)'
+                    count_field = f'COUNT(CASE WHEN {regexp_like} THEN 1 END) AS {validity_format}'
                     validity_format_count_fields.append(count_field)
 
                 row = warehouse.sql_fetchone(
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -9,12 +9,13 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import pyodbc
 from enum import Enum
 from pyhive import hive
 from pyhive.exc import Error
 from thrift.transport.TTransport import TTransportException
 import logging
-from typing import Optional
+from typing import Any, Optional
 
 from sodasql.exceptions.exceptions import WarehouseConnectionError
 from sodasql.scan.dialect import Dialect, SPARK, KEY_WAREHOUSE_TYPE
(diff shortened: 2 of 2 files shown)


packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -9,31 +9,175 @@
...
+        AuthMech=3,
+        SparkServerType=3,
+        ThriftTransport=2,
+        SSL=1,
+        UserAgentEntry=user_agent_entry,
+        LCaseSspKeyName=0 if server_side_parameters else 1,
+        **server_side_parameters,
+    )
+    connection = pyodbc.connect(connection_str, autocommit=True)
+    connection.add_output_converter(pyodbc.SQL_VARCHAR, decode_sketchy_utf8)
```
JCZuurmond, line 150: I am getting the following error when this is not set: 

```
UnicodeDecodeError: 'utf-16-le' codec can't decode bytes in position 718-719: illegal UTF-16 surrogate
```

See the github discussion in the source string where I found this solution. 

I want to think about this for a moment to decide if we should use this solution

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -58,21 +202,35 @@ def sql_tables_metadata_query(self, limit: Optional[int] = None, filter: str = N
     def sql_tables_metadata(self, limit: str = 10, filter: str = None):
         # TODO Implement limit
         with self.create_connection().cursor() as cursor:
-            cursor.execute(f"SHOW TABLES FROM {self.database};")
+            cursor.execute(f"SHOW TABLES FROM {self.database}")
```
JCZuurmond, line 205: Pyodbc does not like the trailing `;`. I think hive requires it though. In dbt spark they implemented a connection specific [wrapper](https://github.com/dbt-labs/dbt-spark/blob/3980e5ce4bce11e96dccf47b89f48e4ba9c2de1d/dbt/adapters/spark/connections.py#L257) to handle this. Maybe we should do the same.

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -9,31 +9,175 @@
...
+    ODBC = "odbc"
+
+
 class SparkDialect(Dialect):
     data_type_decimal = "DECIMAL"
 
     def __init__(self, parser: Parser):
         super().__init__(SPARK)
         if parser:
+            self.method = parser.get_str_optional('method', 'hive')
```
JCZuurmond, line 165: I would prefer to make this connection specific.

JCZuurmond (commented):
I was able to run this against a databricks spark cluster and send the output to the cloud. All in all very happy that it works! Some things to consider:

- in databricks databases are like schemas, the analyze/scan does not have to be database specific. Can we change this in the warehouse implementation?
- can we create a databricks instance for testing this connection in the CI

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: This issue is resolved by adding an aliases to the regexes

core/sodasql/dataset_analyzer.py
```diff
@@ -66,7 +66,7 @@ def analyze(self, warehouse: Warehouse, table_name: str):
                     qualified_regex = dialect.qualify_regex(format_regex)
                     regexp_like = dialect.sql_expr_regexp_like(
                         qualified_column_name, qualified_regex)
-                    count_field = f'COUNT(CASE WHEN {regexp_like} THEN 1 END)'
+                    count_field = f'COUNT(CASE WHEN {regexp_like} THEN 1 END) AS {validity_format}'
```
JCZuurmond, line 69: An encoding error was thrown some times. As [this issues](https://github.com/mkleehammer/pyodbc/issues/328) explained, it could happen due to special characters in the column name. The regexes contain special characters and without an alias they are present in the column name. so this problem was resolved by adding an alias (which do not contain special characters).

JCZuurmond:
All issues are resolved. To be more confident about these changes it would be nice to have a Databricks for testing this in the CI. Is that possible?

Improvements that can be handled in separate PRs:
- In Databricks databases are like schemas, the analyze/scan does not have to be database specific, change the implementation to allow a scan to run over multiple databases.
- Make the YML parser specific to the connection method. Now we get *all* attributes even though they are not all used by the different connection methods.

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: Let's tackle this in another Pr

JCZuurmond:
Can someone have a look why the non-Spark suits fail?

fakirAyoub:
hi @JCZuurmond 

after rebase from main, works fine, can you please rebase and reapply changes before re-actioning the workflow? 

Thanks

fakirAyoub (commented):
LGTM

Just need a rebase so other suites pass.

JCZuurmond:
@fakirAyoub : I did a rebase, tests still do not pass. I think it has to do with that some Github secrets can not be accessed by a PR from a cloned repo.

vijaykiran:
squashed as https://github.com/sodadata/soda-sql/pull/498

Issue #479, opened 2021-09-07
Do not use the `warehouse.connection.cursor` but use one of the sql methods on the warehouse


JCZuurmond:
In the [sampler](https://github.com/sodadata/soda-sql/blob/674b4b8f2352d0691eba8ac99c9d41fce13be274/core/sodasql/scan/sampler.py#L189) the `warehouse.connection.cursor` is used directly. We should use the methods defined on the [`Warehouse`](https://github.com/sodadata/soda-sql/blob/3e21084832d6308f6a6e6869c0147383cf561c98/core/sodasql/scan/warehouse.py).

In `soda-spark` I can overwrite the methods on the warehouse, but since these are not used here, I have to mock a cursor and connection object to make this work.

JCZuurmond:
I have thought about this with the `soda-spark` in mind. I think we should enforce not using the `warehouse.connection` attribute directly. When we stick to a defined API for executing SQL on/in a warehouse, I can implement that API for `soda-spark`. If the code accesses the underlying connection - and cursor on a connection - I do not know which methods will be accessed and I think we should not implement all of them.

The work that should be done is roughly:
- rename `warehouse.connection` to `warehouse._connection` to create a visual cue that this is an attribute which should not be accessed.
- rewrite all references to `warehouse.connection` to use on of the warehouse methods
- narrow the use of the `cursor.description`, I think it is only used to get column name or data type, while it returns 7 things.

PR #483, opened 2021-09-10
Allow spark dialect to have no database specified

JCZuurmond:
In `soda-spark` we register the data frame as global view, we could use a non-global temporary view if we do not need to specify the database. This makes sql metrics (https://github.com/sodadata/soda-spark/pull/78) more intuitive, as you need to defined `global_temp.<table name>` in your `FROM` clause.

### DIFF
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -32,7 +32,7 @@ class SparkDialect(Dialect):
             self.port = parser.get_int_optional('port', '10000')
             self.username = parser.get_credential('username')
             self.password = parser.get_credential('password')
-            self.database = parser.get_str_optional('database', 'default')
+            self.database = parser.get_str_optional('database')
             self.auth_method = parser.get_str_optional('authentication', None)
             self.configuration = parser.get_dict_optional('configuration')
 
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -58,6 +58,9 @@ class SparkDialect(Dialect):
 
     def sql_tables_metadata(self, limit: str = 10, filter: str = None):
         # TODO Implement limit
+        if self.database is None:
+            raise NotImplementedError("Can not query for tables when database is not given.")
+
         with self.create_connection().cursor() as cursor:
             cursor.execute(f"SHOW TABLES FROM {self.database};")
             return [(row[1],) for row in cursor.fetchall()]
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -94,7 +94,8 @@ class SparkDialect(Dialect):
 
     def sql_columns_metadata(self, table_name: str):
         with self.create_connection().cursor() as cursor:
-            cursor.execute(f"DESCRIBE TABLE {self.database}.{table_name}")
+            qualified_table_name = self.qualify_table_name(table_name)
+            cursor.execute(f"DESCRIBE TABLE {qualified_table_name}")
             return [(row[0], row[1], "YES") for row in cursor.fetchall()]
 
     def sql_columns_metadata_query(self, table_name: str) -> str:
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -59,7 +59,7 @@ class SparkDialect(Dialect):
     def sql_tables_metadata(self, limit: str = 10, filter: str = None):
         # TODO Implement limit
         if self.database is None:
-            raise NotImplementedError("Can not query for tables when database is not given.")
+            raise NotImplementedError("Cannot query for tables when database is not given.")
 
         with self.create_connection().cursor() as cursor:
             cursor.execute(f"SHOW TABLES FROM {self.database};")
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -116,7 +116,11 @@ class SparkDialect(Dialect):
             'FLOAT', 'DOUBLE', 'DOUBLE PRECISION', 'DECIMAL', 'NUMERIC']
 
     def qualify_table_name(self, table_name: str) -> str:
-        return f'{self.database}.{table_name}'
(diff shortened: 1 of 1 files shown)


packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -57,6 +57,9 @@ def sql_tables_metadata_query(self, limit: Optional[int] = None, filter: str = N
 
     def sql_tables_metadata(self, limit: str = 10, filter: str = None):
         # TODO Implement limit
+        if self.database is None:
```
JCZuurmond, line 60: I think we should fix this together with [this issue](https://github.com/sodadata/soda-sql/issues/485). If not database is given, then we first query for all schemas/database and then we return all tables within each of the schemas/databases here.

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, parser: Parser):
             self.port = parser.get_int_optional('port', '10000')
             self.username = parser.get_credential('username')
             self.password = parser.get_credential('password')
-            self.database = parser.get_str_optional('database', 'default')
+            self.database = parser.get_str_optional('database')
```
JCZuurmond, line 34: We should test how this works out for the `hive` connector. As suggested in [this PR](https://github.com/sodadata/soda-sql/pull/481), it would be nice to have a parser per type of connection, so that we can have different requirements on the user input.

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -114,7 +118,11 @@ def is_number(self, column_type: str):
             'FLOAT', 'DOUBLE', 'DOUBLE PRECISION', 'DECIMAL', 'NUMERIC']
 
     def qualify_table_name(self, table_name: str) -> str:
-        return f'{self.database}.{table_name}'
+        if self.database is None:
```
fakirAyoub, line 121: Given the fact that `sql_tables_metadata` raises an Exception if database is None, we might never get here if both methods are used sequentially

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -57,6 +57,9 @@ def sql_tables_metadata_query(self, limit: Optional[int] = None, filter: str = N
 
     def sql_tables_metadata(self, limit: str = 10, filter: str = None):
         # TODO Implement limit
+        if self.database is None:
+            raise NotImplementedError("Can not query for tables when database is not given.")
```
fakirAyoub, line 61: Just a small typo here, can not => cannot / can't

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -57,6 +57,9 @@ def sql_tables_metadata_query(self, limit: Optional[int] = None, filter: str = N
...
+
         with self.create_connection().cursor() as cursor:
```
fakirAyoub, line 63: Might need to add a Try / Exception here as the connection might fail for multiple reasons

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -92,7 +95,8 @@ def sql_test_connection(self) -> bool:
 
     def sql_columns_metadata(self, table_name: str):
         with self.create_connection().cursor() as cursor:
```
fakirAyoub, line 97: Same here, Try / Exception for `create_connection`

We might want to create the connection once in a separate method and return a cursor that can be used in others when necessary?

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -92,7 +95,8 @@ def sql_test_connection(self) -> bool:
...
-            cursor.execute(f"DESCRIBE TABLE {self.database}.{table_name}")
+            qualified_table_name = self.qualify_table_name(table_name)
+            cursor.execute(f"DESCRIBE TABLE {qualified_table_name}")
             return [(row[0], row[1], "YES") for row in cursor.fetchall()]
```
fakirAyoub, line 100: What is the semantics behind "YES"? Is it some debugging value or is it passed somewhere else?

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: Correct. In `soda-spark` we will not run `sql_tables_metadata`, but use the `qualify_table_name` directly. And as mentioned when we implement #485, we can use the `schema` name here and allow to reference multiple schemas instead of one database (In spark databases are schemas, and vice versa)

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: Fixed

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: Sure, I think we solve this with #479, then we always use the `Warehouse` class to query the ware house. In that class we can implement this properly with a `try-except`.

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: See comment for `sql_tables_metadata`.

packages/spark/sodasql/dialects/spark_dialect.py
JCZuurmond: This is something that we copied from other dialects (e.g.  [hive](https://github.com/sodadata/soda-sql/blob/aa11659c0d197c976e4011f9cc22ef81856083e8/packages/hive/sodasql/dialects/hive_dialect.py#L107)). "YES" means that the column is nullable.

We should create a data class to make more clear what is meant with this. I would do this outside of this PR

JCZuurmond:
@fakirAyoub we have the same problem with the test as in #481

fakirAyoub:
@m1n0  can you please help with this as I'm not able to look into it these 3 days?

Thank you!

JCZuurmond:
> @m1n0 can you please help with this as I'm not able to look into it these 3 days?
> Thank you!

@m1n0 : The trick is to create a branch on the `soda-sql` repo (not a clone) and recreate this PR, like Vijay did in #498.

vijaykiran:
Squashed with https://github.com/sodadata/soda-sql/pull/499

PR #497, opened 2021-09-20
Force use of warehouse class

### LINKED ISSUE #479: Do not use the `warehouse.connection.cursor` but use one of the sql methods on the warehouse

JCZuurmond:
Resolves #479

### DIFF
--- a/core/sodasql/scan/sampler.py
+++ b/core/sodasql/scan/sampler.py
@@ -114,27 +114,71 @@ class Sampler:
                               if sample_name == 'dataset' else f'{self.scan.scan_yml.table_name}.{column_name}.{sample_name}')
 
         logger.debug(f'Sending sample {sample_description}')
+
+        sample_columns, sample_rows, total_row_count = self._get_query_results_with_limit(sql, limit or 5)
+
+        return {'sample_name': sample_name,
+                'column_name': column_name,
+                'test_ids': test_ids,
+                'sample_columns': sample_columns,
+                'sample_rows': sample_rows,
+                'sample_description': sample_description,
+                'total_row_count': rows_total}
+
+    def send_samples_to_soda_cloud(self, samples_context):
+
+        sample_name = samples_context.get('sample_name')
+        column_name = samples_context.get('column_name')
+        test_ids = samples_context.get('test_ids')
+        sample_columns = samples_context.get('sample_columns')
+        sample_description = samples_context.get('sample_description')
+        total_row_count = samples_context.get('total_row_count')
+        sample_rows = samples_context.get('sample_rows')
+        stored_row_count = len(sample_rows)
+
+        scan_folder_name = (
+            f"{self._fileify(self.scan.warehouse.name)}"
+            + f"-{self._fileify(self.scan.scan_yml.table_name)}"
+            + (f"-{self._fileify(self.scan.time)}" if isinstance(self.scan.time, str) else "")
+            + (
+                f'-{self.scan.time.strftime("%Y%m%d%H%M%S")}'
+                if isinstance(self.scan.time, datetime)
+                else ""
+            )
+            + f'-{datetime.now().strftime("%Y%m%d%H%M%S")}'
+        )
+        file_path = (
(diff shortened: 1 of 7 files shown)


JCZuurmond:
I am stuck because some methods in a dialect use the connection/cursor, while there is no reference to the Warehouse. I think it makes sense to add this reference to the dialect, but we maybe should decide first on what exactly should the `Warehouse` represent and what the `Dialect`. Now these two are mixed e.g.: a connection to a warehouse is created with the `create_connection` method on the dialect.

tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import os
+import sys
+
+# It is not preferred to change the python sys.path like this, however it is
+# required since we are importing from the tests folder like:
+# from tests.common.sql_test_case import SqlTestCase
+sys.path = [os.getcwd()] + sys.path
```
JCZuurmond, line 7: Some test files contain an import like:

``` python
from tests.common.sql_test_case import SqlTestCase
```

This only works if you run pytest as a module from python, like so:

``` bash
python -m pytest
```

The `pytest` cli command will not work:

``` bash
pytest
```

This is because when you run `python` the current working directory is added to the sys path. This does not happen when you run a command like `pytest`. 

I would recommend using the `pytest` command, since this tests the code in isolation of the current working directory. This is also the command most editors run. However, I did not want to change too much in this PR. Therefore this _trick_ always adds the current working directory to your path, independent of how you run it.

tests/common/sql_test_case.py
```diff
@@ -119,7 +118,7 @@ def teardown_close_warehouses(cls):
         SqlTestCase.warehouse_fixtures_by_target = {}
 
     def sql_update(self, sql: str) -> int:
-        return sql_update(self.warehouse.connection, sql)
+        return self.warehouse.sql_update(sql)
 
     def sql_updates(self, sqls: List[str]):
```
JCZuurmond, line 123: I did not have to update this one - like `sql_update` above - to make the local tests pass.

JCZuurmond (commented):
The local tests pass. Points for discussion:

- Some tests circumvent the warehouse API, for example by creating a connection within the test. Do we want to change this too? If yes, let's do this in a separate PR.
- There is confusion about what goes in the `Warehouse` and what in the `Dialect.` For example, the `Dialect` contains a function to connect to a warehouse. What do we want to do with this?

JCZuurmond:
Some of the suites fail because the secrets are hidden from PRs from clones

core/sodasql/scan/sampler.py
```diff
@@ -150,66 +150,32 @@ def __serialize_file_upload_value(self, value):
...
-
-            delta = datetime.now() - start
-            logger.debug(f'SQL took {str(delta)}')
-
-        finally:
-            cursor.close()
-
-        return stored_rows, sample_columns
+    def save_sample_to_local_file(self, sql, temp_file, *, limit = None):
+        logger.debug('Executing SQL query: \n%s', sql)
```
JCZuurmond, line 161: I think we should add this to the methods on the warehouse instead of doing this logging here.

vijaykiran:
@JCZuurmond  Is this still required? if so, can you please rebase on main and push ?

JCZuurmond:
It still relevant for me. Soda should decide if they agree that the `Warehouse` should be the API for sending queries to a warehouse and retrieving results, instead of accessing the underlying connection directly.

vijaykiran:
Closing this and leaving the option open for next version.

PR #499, opened 2021-09-20
[spark-sql] Allow spark dialect to have no database specified

vijaykiran:
- This allows us to have views without a schema/database
- Make default database None


JCZuurmond:
@vijaykiran : Could you wait with merging this?

PR #503, opened 2021-09-23
Filter columns in spark dialect

### LINKED ISSUE #502: Spark dialect: describe table returns partitioning column for delta tables
**Describe the bug**
For delta tables `DESCRIBE TABLE` returns an empty column, `# Partitioning` and all partitioning columns. These columns show up in Soda cloud as if these are "normal" columns.

**To Reproduce**
1. Run `soda scan ...` against a delta table

**Context**
It probably depends on the table type (what comes after `USING`). I see this for a delta table, it might also happen with other table types.

**Python Version**: 3.7

**Soda SQL Version**: 2.1.0b17 

**Warehouse Type**: Spark dialect

JCZuurmond:
Resolves #502. Unfortunately I was not able to test this with a delta table.

### DIFF
--- a/packages/spark/sodasql/dialects/spark_dialect.py
+++ b/packages/spark/sodasql/dialects/spark_dialect.py
@@ -16,7 +16,7 @@ from pyhive import hive
 from pyhive.exc import Error
 from thrift.transport.TTransport import TTransportException
 import logging
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from sodasql.exceptions.exceptions import WarehouseConnectionError
 from sodasql.__version__ import SODA_SQL_VERSION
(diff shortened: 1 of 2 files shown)


packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -230,11 +232,69 @@ def sql_test_connection(self) -> bool:
...
+        ----------
+        table_name : str
+            The table to show columns in.
+
+        Returns
+        -------
+        out : List[str]
+            The column names.
+        """
+        qualified_table_name = self.qualify_table_name(table_name)
```
JCZuurmond, line 249: Something for a refactor once: let all methods on the dialect expect the qualified table name instead of creating this table name within the methods.

tests/warehouses/spark_suite.py
```diff
@@ -7,3 +9,15 @@ class SparkSuite(SqlTestSuite):
     def setUp(self) -> None:
         self.target = TARGET_SPARK
         super().setUp()
+
+    def test_sql_columns_metadata(self):
```
JCZuurmond, line 13: I think this could be a test applied to **all** dialects. However, then we need to integrate `ColumnMetadata` with the core.

fakirAyoub (approved):
LGTM! I'll let @vijaykiran merge to main when available.

PR #505, opened 2021-09-28
Make bigquery `account_info_json` optional

JCZuurmond:
Log error instead of throw exception

When no account info credentials are found then log error. This is more
consistent with the other behaviour like being unable to decode the json.


JCZuurmond:
In the new soda sql version we would like against for each of the dialects. Throwing the an error when the `account_info_json` is not found by the big query dialect blocks us from using the dialect for testing, see example in [this PR](https://github.com/sodadata/soda-sql-v3/pull/2).

PR #507, opened 2021-09-29
fix timeout validation

fakirAyoub:
* Temporary fix timeout validation for datawarehouses


vijaykiran:
@fakirAyoub  I think the function  should return True now with just `logger` it will return None.

I think it is better to only remove `for tables ...` than the entire logic of getting list of tables, so we do another connection validation to list tables check here.

vijaykiran (changes requested):
Please check my comment on the PR

PR #540, opened 2021-10-26
[core] Add link to Soda Community Slack

fakirAyoub:
Add link as an info after errors, when applicable


vijaykiran:
@fakirAyoub  Looking at all the places, I think the best option is to add the message at scan end, I see that the join slack is printed several times in case of exceptions or metrics issues which can get annoying pretty quickly.

vijaykiran (changes requested):
Please see my comment on the discussion

PR #545, opened 2021-10-28
Add warning message when user is using an old version

fakirAyoub:


core/sodasql/cli/cli.py
```diff
@@ -31,7 +31,10 @@
 
 LoggingHelper.configure_for_cli()
 logger = logging.getLogger(__name__)
-
+latest_version = luddite.get_version_pypi('soda-sql')
```
vijaykiran, line 34: There is no soda-sql anymore, it should be soda-sql-core. https://pypi.org/project/soda-sql-core/

dev-requirements.txt
```diff
@@ -151,6 +151,5 @@ websocket-client==0.57.0
     # via docker
 zipp==3.5.0
     # via importlib-metadata
-
```
vijaykiran, line 154: can you delete this change/file from commit?

core/setup.py
```diff
@@ -23,7 +23,8 @@
     'cryptography==3.3.2',
     'pyyaml>=5.4.1, <6.0',
     'requests>=2.23.0, <3.0',
-    'Deprecated>=1.2.13, <1.3'
+    'Deprecated>=1.2.13, <1.3',
+    'luddite==1.0.2'
```
vijaykiran, line 27: It is better to use version ranges, maybe pin it <1.1 - so we can get latest patch versions.

core/sodasql/cli/cli.py
vijaykiran: This only warns when CLI is invoked, the version check should also happen for programmatic scans using ScanBuilder.

vijaykiran (changes requested):

core/sodasql/cli/cli.py
fakirAyoub: Added a __call__ method to ScanBuilder for this purpose. Will add the method to cli_helper class once the other branch is merged.

fakirAyoub: Yes for the commit, please check it out (I can see the changes in the branch)

Tested offline in my computer and worked fine, maybe because of some cache? I will add an exception to this right now.

core/sodasql/scan/scan_builder.py
```diff
@@ -69,7 +72,15 @@ def __init__(self):
         self.assert_no_warnings_or_errors = True
         self.soda_server_client: SodaServerClient = None
 
+    def __call__(self, *args, **kwargs):
```
vijaykiran, line 75: I don't see how `__call__` helps, it is better to add the version check to `__init__`

fakirAyoub: Update done. Thanks

core/sodasql/scan/scan_builder.py
```diff
@@ -68,8 +72,17 @@ def __init__(self):
         self.parsers: List[Parser] = []
         self.assert_no_warnings_or_errors = True
         self.soda_server_client: SodaServerClient = None
+        latest_version = ""
+        try:
+            latest_version = luddite.get_version_pypi('soda-sql-core')
+        except:
+            logger.warning("Cannot check Soda SQL version.")
+
+        if SODA_SQL_VERSION != latest_version:
```
vijaykiran, line 81: This shouldn't trigger if we are not able to get the latest version (in case of running behind a firewall or network issue) otherwise even though people are using latest version the logger will still report that it is not latest.

core/sodasql/cli/cli.py
```diff
@@ -31,7 +31,14 @@
 
 LoggingHelper.configure_for_cli()
 logger = logging.getLogger(__name__)
-
+latest_version = ""
+try:
+    latest_version = luddite.get_version_pypi('soda-sql-core')
+except:
+    logger.warning("Cannot check Soda SQL version.")
+if SODA_SQL_VERSION != latest_version:
```
vijaykiran, line 39: Check my comment below, this should't log a warning when we can't determine the latest version.

vijaykiran (changes requested):

core/sodasql/cli/cli.py
fakirAyoub: changes done

PR #563, opened 2021-11-12
[WIP] Open telemetry #543

### LINKED ISSUE #543: [Telemetry] Add anonymous usage and profiling statistics collection
## Goal
In order to understand the usage, adoption, and performance of our tool better we should collect usage and performance statistics. What we want to achieve with those are the following goals:
- Pro-actively track bugs and find and fix errors to get ahead of issues faster
- Manage performance to keep our tools fast and secure,
- Troubleshoot issues, speed up resolution, and minimize any downtime for users
- Understand how you user use our tools to improve user flow and experience 

Those statistics must be **strictly anonymous**, will not pertain to any private information, and will be limited to tracking events that emanate from the tool itself.

**What do we consider private and will never track?**
- Personally identifiable information (names, email addresses, credentials, geolocation, IP, etc.)
- Organizations' or Users' Proprietary information (e.g. name of checks, datasets, test data, or test results content)

## Requirements
- Users should be able to opt-out at any point via an easy to configure switch (e.g. YAML variable in `config.yaml`)
- Documentation pages are updated with reference to this GitHub issue along with goal and rationale digest and how to opt-out of the tool.
- Telemetry events should not significantly hurt the performance of the tool (i.e. we need to make sure we use an async or non-blocking framework/API for that).

### Event Collection Framework
Two generally accepted frameworks come to mind: 
- [**Open Telemetry**](https://opentelemetry.io/): built on open standards, easy to configure, suitable for tracing of performance (traces and spans) with high granularity.

- [**Snowplow**](https://snowplowanalytics.com/): open source option, used by a lot of organizations and reliable, little heavier setup, only made for behavioral events so tracing around performance would have to be shoehorned in their data models.

### Event Schema & Attributes
Names and implementations are just ideas. Let's challenge those in this issue and we can update this once we're set.
- `user_cookie_id`: we could generate a UUID and store it in `~/.soda/config.yaml` or in a different file in that config folder.
- `command(_ran)`: we could get it as a trace name or just add it as an attribute in the attributes key, value object. We should be able to capture this from `click` or API method name.
- `command_options`: these should be flags used if we're talking about a CLI invocation. Can be captured from `click` but we might want to find a more internal way of capturing it from method signatures of class attributes so that we are ready for API-driven usage in the future.
- `version`: which version of soda_sql is this coming from: https://github.com/sodadata/soda-sql/blob/main/core/sodasql/__version__.py or whatever else makes sense.
- `datasource_type`: other names could be `dialect_id` but it lends itself less to `soda-spark` or other data source types so maybe datasource_type is fine.
- `datasource_id`: given the importance of knowing how many projects and teams are using the tool, we should add a `dataserouce_id` which should be a hash of the connection parameter that best uniquely identifies a group of people working on the same project.
    - Postgres, Redshift, Spark, Presto: host
    - Snowflake: account
    - BigQuery: project
    - etc. to be determined on a dialect/supported data source basis.
    - see a similar rationale from [dbt](https://github.com/dbt-labs/dbt-core/issues/3713)
   
~- `run_status`: the alternative name could be `invocation_status` and would allow us to know if a `scan`, `analyze` etc. executed successfully or not (exception types should be logged as well).~
~- `metrics_count`: number of metrics collected/derived in a scan.~
~- `metrics_invoked`: which metrics are people using. NOTE: we should not log custom metrics derivation code (i.e. the SQL query of a custom metric or any formula as this is considered proprietary information). When uses use `custom_sql` we should just know that they used a custom metric.~
~- `tests_count`: count of tests evaluated.~
~- `tests_invoked`: same reasoning as metrics list, if the tests are built-in and do not reveal any proprietary logic we should have a list, if not let's leave it out.~

- `sql_metrics_count`: see comment about `metrics_count`. SQL metrics contain proprietary business logic so we absolutely do not want to track the SQL query itself but just the fact that a custom metric was used.❓
- `historic_metrics_count` ❓
- `architecture`: we can use the `platform` library of `os` library to get at which architecture we're running on
- `operating_system`: see above.
- `python_version`
- `python_implementation`
- `invocation_start`: span start/end in the OT framework
- `invocation_end`: same as above.

m1n0:
First iteration of Open Telemetry tracing.

Contains transparent anonymous basic usage tracing. Fixes #543 (see that issue for much more info)

Remains to be done:
- improve documentation and comments
- document skipping telemetry
- test creating datasource hash for all warehouse types (only postgres tested at the point of creating the MR)
- polish the ConfigHelper


core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+        return str(uuid.uuid4())
+
+    def __ensure_basic_config(self) -> None:
+        for key, value in self.DEFAULT_CONFIG.items():
+            if key not in self.config:
+                self.upsert_value(key, value)
+
+    @property
+    def skip_telemetry(self) -> bool:
+        return self.config.get("skip_telemetry", False)
```
bastienboutonnet, line 121: just a naming question here. I would probably go for a different name such as `send_anonymous_usage_stats` with a True default. I think from a user perspective it's more friendly to ask them for a positive question rather than negative. That does mean reversing our boolean logic but shouldn't be too complex.

This also is totally identical to the dbt approach (https://docs.getdbt.com/reference/profiles.yml#send_anonymous_usage_stats) and should be more familiar to users.

packages/bigquery/sodasql/dialects/bigquery_dialect.py
```diff
@@ -50,6 +50,15 @@ def default_connection_properties(self, params: dict):
             'dataset': params.get('database', 'Eg your_bigquery_dataset')
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.dataset_name,
+            self.account_info_dict['type'] or None,
+            self.account_info_dict['project_id'] or None,
```
bastienboutonnet, line 58: `project_id` is what we want to push to `datasource_id` for BQ

packages/athena/sodasql/dialects/athena_dialect.py
```diff
@@ -56,6 +56,14 @@ def default_connection_properties(self, params: dict):
             'catalog': 'AwsDataCatalog'
         }
 
+    def safe_connection_data(self):
+        return [
```
bastienboutonnet, line 60: is there a concept of host or something in athena? if so I think that is what we want to push into `datasource_id` for this warehouse.

packages/hive/sodasql/dialects/hive_dialect.py
```diff
@@ -47,6 +47,15 @@ def default_connection_properties(self, params: dict):
             'database': params.get('database', 'your_database')
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.host,
```
bastienboutonnet, line 53: we will push this into `datasource_id`

packages/mysql/sodasql/dialects/mysql_dialect.py
```diff
@@ -41,6 +41,15 @@ def default_connection_properties(self, params: dict):
             'database': params.get('database', 'your_database')
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.host,
```
bastienboutonnet, line 47: we will push this into `datasource_id`

packages/postgresql/sodasql/dialects/postgres_dialect.py
```diff
@@ -45,6 +45,16 @@ def default_connection_properties(self, params: dict):
             'schema': 'public'
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.host,
```
bastienboutonnet, line 51: we will push this in `datasource_id`

packages/snowflake/sodasql/dialects/snowflake_dialect.py
```diff
@@ -56,6 +56,17 @@ def default_connection_properties(self, params: dict):
             'schema': 'PUBLIC'
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.database,
+            self.account,
```
bastienboutonnet, line 63: we will push this into `datasource_id`

packages/spark/sodasql/dialects/spark_dialect.py
```diff
@@ -169,6 +169,17 @@ def default_connection_properties(self, params: dict):
             'database': params.get('database', 'your_database')
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.host,
```
bastienboutonnet, line 175: we will push this into `datasource_id`

packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
```diff
@@ -66,6 +66,17 @@ def default_connection_properties(self, params: dict):
             'schema': 'public'
         }
 
+    def safe_connection_data(self):
+        return [
+            self.type,
+            self.host,
```
bastienboutonnet, line 72: we will push this into `datasource_id`

core/sodasql/scan/dialect.py
```diff
@@ -109,9 +111,31 @@ def create_for_warehouse_type(cls, warehouse_type):
                     KEY_WAREHOUSE_TYPE: warehouse_type}))
 
     def default_connection_properties(self, params: dict):
-        # to be overriden by subclass
+        # to be overridden by subclass
+        pass
+
+    def safe_connection_data(self):
```
bastienboutonnet, line 117: Judging by the concrete implementations below, unless I'm mistaken, we end up sending usernames and passwords which I think doesn't reflect the name or docstring.

core/sodasql/scan/warehouse.py
```diff
@@ -14,16 +14,18 @@
...
 
-
 class Warehouse:
 
     def __init__(self, warehouse_yml: WarehouseYml):
         self.name = warehouse_yml.name
         self.dialect: Dialect = warehouse_yml.dialect
         self.connection = self.dialect.create_connection()
+        soda_telemetry.set_attribute('datasource_type', self.dialect.type)
+        soda_telemetry.set_attribute('datasource_hash', soda_telemetry.obtain_datasource_hash(self.dialect))
```
bastienboutonnet, line 28: In #543 (haha it's almost the default postgres port 😂 ) there is no attribute named `datasource_hash`. Did you mean `datasource_id`? If so, then I think we're sending the wrong thing to it.

Sorry for the stupid question. What is the difference between `dialect.type` and `dialect`?

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __setup(self):
+        session
+
+        provider = TracerProvider(
+            resource=Resource.create(
+                {
+                    'os.architecture': platform.architecture(),
+                    'python.version': platform.python_version(),
+                    'python.implementation': platform.python_implementation(),
```
bastienboutonnet, line 63: would it be possible to use _s here when I flatten this into a datawarehouse table it'll be painful probably. Unless I'm missing some reasoning?

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+        # provider.add_span_processor(otlp_processor)
+        trace.set_tracer_provider(provider)
+
+    def set_attribute(self, key: str, value: str) -> None:
+        if not self.__skip:
+            current_span = trace.get_current_span()
+            current_span.set_attribute(key, value)
+
+    def obtain_datasource_hash(self, connection: Dialect):
+        return connection.generate_hash_safe()
```
bastienboutonnet, line 86: this also ties into my question about whether this is what we mean by `datasource_id` and in that case I think we're sending too much data here to be hashed. Unless we're adding this for other purposes but in that case let's discuss why we send the credentials.

If not, I would change to `obtain_datasource_id` and change the list in the dialect files to only give us the info that is most unique but not personally identifying or proprietary business logic identifying (e.g. schema names etc).

I've added comments for each adapter file below about what I think this would be but you and @vijaykiran please double check that it makes sense.

bastienboutonnet (commented):
Thanks for this work @m1n0 . It's looking in pretty good shape from my perspective. I glossed over some of the core implementation details as I think @vijaykiran is better suited for this than me.

I left comments mainly about naming and making sure that we're tracking the correct info in the attributes as defined in #543 and more specifically this comment: https://github.com/sodadata/soda-sql/issues/543#issuecomment-966298328

vijaykiran (changes requested):

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self,  skip: bool = False):
+        self.__skip = skip
+        if self.__skip:
+            logger.info("Skipping usage telemetry.")
+        else:
+            logger.info("Sending usage telemetry.")
+            self.__setup()
+
+    def __setup(self):
+        session
```
vijaykiran, line 56: is this a typo?

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+                    'platform': platform.platform(),
+                    ResourceAttributes.SERVICE_VERSION: SODA_SQL_VERSION,
+                    ResourceAttributes.SERVICE_NAME: 'soda',
+                    ResourceAttributes.SERVICE_NAMESPACE: 'soda-sql',
+                }
+            )
+        )
+        console_span_processor = BatchSpanProcessor(ConsoleSpanExporter())
+        provider.add_span_processor(console_span_processor)
+        # otlp_exporter = OTLPSpanExporter(endpoint=ENDPOINT)
```
vijaykiran, line 75: Can you make this (later) overridable from commandline or config?

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+        # otlp_processor = BatchSpanProcessor(otlp_exporter)
+        # provider.add_span_processor(otlp_processor)
+        trace.set_tracer_provider(provider)
+
+    def set_attribute(self, key: str, value: str) -> None:
+        if not self.__skip:
+            current_span = trace.get_current_span()
+            current_span.set_attribute(key, value)
+
+    def obtain_datasource_hash(self, connection: Dialect):
```
vijaykiran, line 85: This doesn't look like an instance method may be just make it static?

vijaykiran: And also I'd rather name it dialect since it is a dialect object and not connection :)

core/setup.py
```diff
@@ -23,7 +23,10 @@
     'cryptography==3.3.2',
     'pyyaml>=5.4.1, <6.0',
     'requests>=2.23.0, <3.0',
-    'Deprecated>=1.2.13, <1.3'
+    'Deprecated>=1.2.13, <1.3',
+    'opentelemetry-api>=1.6.2,<1.7',
+    'opentelemetry-sdk>=1.6.2,<1.7',
```
vijaykiran, line 28: looks like we can remove opentelemetry-api and -sdk as they seem to be transitive reps from exporter below.

core/sodasql/cli/cli.py
```diff
@@ -50,6 +53,8 @@ def main():
 @click.option('-p', '--password', required=False, default=None,
               help='The password to use for the connection, through env_var(...)')
 @click.option('-w', '--warehouse', required=False, default=None, help='The warehouse name')
+
```
vijaykiran, line 56: No blank lines between decorators :) 

https://www.flake8rules.com/rules/E304.html

core/sodasql/cli/cli.py
```diff
@@ -62,6 +67,19 @@ def create(warehouse_type: str,
 
     WAREHOUSE_TYPE is one of {postgres, snowflake, redshift, bigquery, athena}
     """
+
+    soda_telemetry.set_attribute('cli_command_name', 'create')
```
vijaykiran, line 71: Just aside: it is annoying that we can't get the current function name within the function thanks to rejected https://www.python.org/dev/peps/pep-3130/

core/sodasql/cli/cli.py
```diff
@@ -62,6 +67,19 @@ def create(warehouse_type: str,
...
+    soda_telemetry.set_attribute('datasource_type', warehouse_type)
+
+    span_setup_function_args(
+        locals(),
```
vijaykiran, line 75: This is a bit risky code and it might raise eyebrows because we end up sending password (which is part of locals) to the span_setup_function. 

I'd rather make the args arg (for span_setup_function) more explicit. 

WDYT?

core/sodasql/cli/cli.py
```diff
@@ -153,6 +171,8 @@ def create(warehouse_type: str,
         logger.info("If you think this is a bug in Soda SQL, please open an issue at: "
                     "https://github.com/sodadata/soda-sql/issues/new/choose")
         sys.exit(1)
+    finally:
+        trace.get_current_span().end()
```
vijaykiran, line 175: I don't think we need this anymore .. right ? traces should be set to auto-end when we exit ...

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+class ConfigHelper:
+    """Helper class for handling global Soda config.
+    """
+    DEFAULT_CONFIG = {
+        'skip_telemetry': False,
+        'user_cookie_id': str(uuid.uuid4())
+    }
+    LOAD_PATHS = ["~/.soda/config.yml", ".soda/config.yml"]
+    __instance = None
+    __config: Dict = None
```
vijaykiran, line 31: can we make it just empty dict ? None is not a type of `Dict`

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+    file_system = FileSystemSingleton.INSTANCE
+
+    @staticmethod
+    def get_instance(path: Optional[str] = None):
+        if ConfigHelper.__instance == None:
```
vijaykiran, line 36: I think the preferred comparison operator is `is None` here

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+            ConfigHelper()
+        return ConfigHelper.__instance
+
+    def __init__(self, path: Optional[str] = None):
+        if ConfigHelper.__instance != None:
```
vijaykiran, line 41: same as above - `is not None` is preferable

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+    def reload_config(self) -> Dict:
+        for path in self.LOAD_PATHS:
+            logger.info(f"Trying to load Soda Config file {path}.")
+
+            if self.file_system.file_exists(path):
+                self.__config = yaml.load(
+                    self.file_system.file_read_as_str(path),
+                    Loader=yaml.SafeLoader
+                )
+                break
```
vijaykiran, line 76: This is not returning the config dict .. is that intentional ? If so, can you update the function return type signature?

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,121 @@
...
+        self.file_system.file_write_from_str(
+            self.config_path,
+            yaml.dump(
+                config,
+                default_flow_style=False,
+                sort_keys=False
+            )
+        )
+
+    def generate_user_cookie_id(self) -> str:
```
vijaykiran, line 111: This can be a class method

vijaykiran (changes requested):

packages/athena/sodasql/dialects/athena_dialect.py
bastienboutonnet: Do you know what information we can use to uniquely identify an account or host @vijaykiran

core/sodasql/scan/warehouse.py
bastienboutonnet: Ok thanks for the clarification. I was 99% sure of that after seeing more of the code.

So I think for me, it's a case of making sure we don't hash the entire dialect object into the `datasource_hash` that will become the `datasource_id` and hopefully my comments on each dialect about what should be hashed form the dialect object into `datasource_id` should get us in a really good shape.

core/sodasql/telemetry/soda_telemetry.py
```diff
@@ -0,0 +1,93 @@
...
+        return connection.generate_hash_safe()
+
+    @property
+    def user_cookie_id(self) -> str:
+        return self.soda_config.get_value('user_cookie_id')
+
+# Global
+soda_telemetry = SodaTelemetry(skip=soda_config.skip_telemetry)
```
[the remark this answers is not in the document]
bastienboutonnet: since soda-sql is open source we can even use this: https://pre-commit.ci/ which is a CI version of pre-commit (by same author) that will auto fix the code even if the dev does not have pre-commit set up locally. And pre-commit definitely has an end of file fixer (i use it in all my projects)

core/sodasql/common/config_helper.py
bastienboutonnet: of if you want to still use mypy type annotation classes you can do `Optional[Dict[Any, Any]]` or set up a placeholder dict indeed. I never know about these and do it very inconsistently...

core/sodasql/common/config_helper.py
bastienboutonnet: I think flake8 would surface this. Might be a good idea to set it up as a pre-commit hook

core/sodasql/telemetry/soda_telemetry.py
m1n0: It seems that the only use case for overriding here would be for development of OT for now, so I just added a quick optional config file entry for this for now

core/sodasql/telemetry/soda_telemetry.py
m1n0: changed

packages/athena/sodasql/dialects/athena_dialect.py
m1n0: indeed, Athena is challenging to create a unique-enough hash without sensitive info

packages/bigquery/sodasql/dialects/bigquery_dialect.py
m1n0: I wasnt sure whether project id is globally unique so added more attributes, but now I found this

> Each project ID is unique across Google Cloud. Once you have created a project, you can delete the project but its ID can never be used again.
(https://cloud.google.com/docs/overview?hl=id) 

so yes it is safe to only use project id, changed

Note - we will not be sending the actual project id as `datasource_id`, but the hash of it.

packages/hive/sodasql/dialects/hive_dialect.py
m1n0: not sure what you mean, but this list is what goes to our hashing function that _generates_ the `datasource_id`. The hashing is (arguably) irreversible, so no one will be able to determine host/user/schema or any of these `safe_connection_data` values from the hash. We add all this info as input to the hashing so that we have enough unique values coming into it to ensure uniqueness of the hash. Otherwise we would generate conflicting `datasource_ids` (hashing is deterministic, i.e. we would end up with same hashes for what would actually be different warehouses)

core/sodasql/cli/cli.py
m1n0: agreed, changed to explicit dicts. This was unnecessarily complex, doing it explicitly simplified the logic a lot as well.

core/sodasql/telemetry/soda_telemetry.py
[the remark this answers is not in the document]
m1n0: indeed!

core/sodasql/telemetry/soda_telemetry.py
bastienboutonnet: I meant whether it would be possible to use underscores instead of dots in the key names so that it is easier when we flatten the json to table columns. I'm not entirely sure how well snowflake copes with dots as column names.

packages/athena/sodasql/dialects/athena_dialect.py
bastienboutonnet: do you have a sample of what it looks like so I can see?

packages/hive/sodasql/dialects/hive_dialect.py
bastienboutonnet: I think we can be ok with the following:
- host
- port

we will have the type already as `datasource_type` anyway and it doesn't make sense to me to include `database` (since our definition of core system is that they use some host, if things happen in separate dbs and what not we don't really care to differentiate.

In the unlikely event that someone was table to reverse it they would have some credentials + some proprietary info. It's not uncommon for database to carry some info about what is in there and users may not want us to be able to know.

From an analytical point of view host is enough host and port is also fine, and I would argue we don't even need port.

packages/mysql/sodasql/dialects/mysql_dialect.py
bastienboutonnet: I would leave it at hashing the `host` here

packages/postgresql/sodasql/dialects/postgres_dialect.py
bastienboutonnet: I would leave it at `host`

packages/snowflake/sodasql/dialects/snowflake_dialect.py
bastienboutonnet: I would leave it at `account`

packages/spark/sodasql/dialects/spark_dialect.py
bastienboutonnet: I would leave it at `host` unless for spark people always use the same host even between teams?

packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
bastienboutonnet: `host` should be sufficient unless host is potentially often the same across different organisations.

bastienboutonnet (commented):
I had another look. I still think we're including too much information in the `datasource_id` hash. Normally the account or the host should be sufficient, unless host is something that several orgs would use. Are there dialects in this repo that tell people to use a certain host name always?

packages/spark/sodasql/dialects/spark_dialect.py
m1n0: Same as with snowflake, I do not know whether host is unique enough.  Keeping `type`, `host`, `port` and `database` for now.

packages/hive/sodasql/dialects/hive_dialect.py
bastienboutonnet: I can buy that @m1n0! I totally understand where you were coming from but wanted to push to the point where we think that we're tracking the correct components. Your point about potential generic host names is very good. I actually made a comment where I was voicing that same concern in the form of a question somewhere.

And I agree that if we think the host is likely to be often the same then type could add some entropy.

The username, however, besides being potentially a security risk, may have added too much entropy in the case where several users connect to the same host and type but with a different username. I think this scenario is likely and it would have actually prevented us of getting an idea of a "team, using a core system" do you see what I mean?

So the decision to remove it solves those two issues and is a good decision in my opinion.

core/sodasql/cli/cli.py
```diff
@@ -62,6 +66,23 @@ def create(warehouse_type: str,
...
+
+    span_setup_function_args(
+        {
+            'command_argument': {'datasource_type': warehouse_type},
+            'command_option':
+                {
+                    'file': file,
+                    'warehouse': warehouse,
+                    'database': database,
+                    'username': username
```
bastienboutonnet, line 81: Is this the username or whether they pass a username?

core/sodasql/telemetry/soda_telemetry.py
bastienboutonnet: Thanks for checking on snowflake @m1n0 

Regarding the second idea. At this point, I'm not too sure, as I have not looked enough into how I will flatten. My idea would be I'd flatten as simple columns yes, but I prefer the idea of the structured approach from a "type" perspective, however, I don't know if that makes tings easy or not.

At this point let's do the following:
- leave the dots
- I need to play around with it which I should be able to do within a day or so of setting up a pipe.
- if it's annoying, we can choose to break the best practice or I place some aliasing in Transform stage or maybe the structured approach works.

Since I don't have a good answer, let's add to the current scope and go with that. If we need to make a change it will be small I imagine.

packages/athena/sodasql/dialects/athena_dialect.py
bastienboutonnet: That looks very good to me!

packages/snowflake/sodasql/dialects/snowflake_dialect.py
bastienboutonnet: Account is unique enough for the purposes, so we can just have account here.

core/sodasql/cli/cli.py
vijaykiran: it is the username you can pass that will be added to the warehouse.yml

core/sodasql/common/config_helper.py
```diff
@@ -0,0 +1,125 @@
...
+            self.config_path,
+            yaml.dump(
+                config,
+                default_flow_style=False,
+                sort_keys=False
+            )
+        )
+
+    @classmethod
+    def generate_user_cookie_id() -> str:
```
vijaykiran, line 115: I think Sonar cloud is shouting at this because the `cls` arg is missing

packages/snowflake/sodasql/dialects/snowflake_dialect.py
m1n0: removed account

core/sodasql/common/config_helper.py
m1n0: actually it can be static as well, changed.

core/sodasql/cli/cli.py
m1n0: yup it was that, I changed warehouse, database and username to be just booleans whether the param has been passed and renamed it to `using_X`

PR #564, opened 2021-11-15
[sqlserver] Support for SQLServer Dialect

fakirAyoub:
* redefinition of regex expressions
* add a `column name` parameter to some function in the dialect


packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
```diff
@@ -86,13 +86,13 @@ def sql_connection_test(self):
...
         try:
             conn = pyodbc.connect(
-                ('Trusted_Connection=YES;' if self.trusted_connection else '') +
-                'DRIVER={' + self.driver +
-                '};SERVER=' + self.host +
-                ';PORT=' + self.port +
-                ';DATABASE=' + self.database +
-                ';UID=' + self.username +
-                ';PWD=' + self.password)
+                ('Trusted_Connection=NO;' if True else '') +
```
vijaykiran, line 89: Please revert these .. trusted connection is required to make sure that AD SSO works.

packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
```diff
@@ -62,7 +62,7 @@ def default_connection_properties(self, params: dict):
             'port': '1433',
             'username': 'env_var(SQLSERVER_USERNAME)',
             'password': 'env_var(SQLSERVER_PASSWORD)',
-            'database': params.get('database', 'your_database'),
+            'database': params.get('database', 'sodasqldb'),
```
vijaykiran, line 65: I guess you didn't intend to leave it here?

vijaykiran (changes requested):
I'm not sure if we want to skip the tests for all other warehouses, because SQL Server isn't working .. I need to think about a better strategy. May be parametrise the tests and skip only for SQLServer with a message?

packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
fakirAyoub: Yes I will be reversing these, hence the WIP

packages/sqlserver/sodasql/dialects/sqlserver_dialect.py
fakirAyoub: this is just my local db that I used for testing

tests/local/warehouse/validity/test_date_and_time_validity_formats.py
```diff
@@ -21,7 +22,7 @@ class TestDateAndTimeValidityFormats(SqlTestCase):
     def test_date_eu(self):
         self.sql_recreate_table(
             [f"name {self.dialect.data_type_varchar_255}"],
-            ["('21-01-2021')",
```
vijaykiran, line 24: Any reason why this is removed? I think we support date validation with `-` in it

PR #566, opened 2021-11-18
Dialect - use abstract methods instead of exceptions fixes#565

### LINKED ISSUE #565: Use abstract Dialect class and methods instead of RuntimeError
Dialects are controlling the mandatory abstract methods using `RuntimeError` - it would be neater to use abstract classes and methods, and it would prevent IDEs complaining about unreachable code after `dialect.create_connection()`.

m1n0:
Fixes #565

### DIFF
--- a/core/sodasql/scan/dialect.py
+++ b/core/sodasql/scan/dialect.py
@@ -10,6 +10,7 @@
 #  limitations under the License.
 from __future__ import annotations
 
+import abc
 import re
 from datetime import date
 from numbers import Number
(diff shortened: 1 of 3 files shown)


m1n0:
Hm this fails on non-specific dialect [tests](https://github.com/sodadata/soda-sql/blob/main/tests/local/independent/test_sql_expressions.py#L17) which now cannot be instantiated. I set up a mock dialect just to get around that, which is a bit silly, wdyt?

PR #568, opened 2021-11-18
Let scan have different exit codes

### LINKED ISSUE #555: Have an exit code `2` for when tests / monitors fail
I would like to differentiate between failing tests and other kind of failures (bugs, connections not working, etc.). This would allow me to have a CI running that does not fail if the test / monitors fail (we get notifications for this via the cloud), but the CI **does** fail if the Soda scan fails for another reason.

Exit with code `2` if tests / monitors fail, exit with code `1` for any other failures.

JCZuurmond:
Resolves #555 

Exit codes:
- 0 : all tests passed
- 1 : an exception occured
- 2 : at least one test failed


JCZuurmond:
@vijaykiran : Could we merge this in? I miss out on some errors in the CI since I can not make a difference between test that fail and scan logic that fails

Issue #570, opened 2021-11-19
Allow users to ingest dbt test results into Soda Cloud via soda-sql


bastienboutonnet:
# Context
Users who use dbt have spent time crafting tests in their transformations. Because we don’t want users to repeat themselves to benefit from Soda Cloud’s visibility and incident management we plan to:
- Ingest dbt test results into Soda Cloud.
- Provide clear indications of whether the tests are maintained in soda-sql or ingested from dbt
- Allow collaboration and reporting to be done on those results just as if they were maintained from soda-sql

# User Flow
Users should be able to call a command such as `soda ingest --tool dbt`, provide soda-sql with paths to the dbt artifacts, and have their tests results show up in the Soda Cloud UI. We will have to be clear in our documentation that ingestion should be orchestrated once dbt has been invoked and the transformations have completed. Soda-sql (at least in this first version) will not integrate with the core dbt API. This means it will not orchestrate a dbt run or be called by dbt. Those scenarios may be revisited in further iterations of our integration.

We may also consider creating a new `soda` command such as `soda capture` which would run `soda scan` and follow it with `soda ingest` or a flag on `soda scan` such as `--ingest dbt`. Up for discussion and TBD once minimal isolated approach is working. This is sugar coating at this point.

# Overall Approach
In this v1, we want to keep it simple. dbt generates two valuable artifacts: `manifest.json` and `run_results.json`. When it comes to test results, we can parse the results from the `run_results.json` and use the info in the `manifest.json` to gain knowledge about the table (dataset in Soda's language) that it maps on to.

Once we have parsed all the info we need (see below) we can compose a `TestObject` similar (but not exactly) to the soda-sql `TestObject` that is then exposed to the Soda Cloud backend API. The UI will show the ingested tests on the dataset view, those tests will be clearly identified as `external` and display a dbt logo or whatever UI choice makes sense.

# Specific Requirements
We will want to parse and represent the following information from dbt test results:
- `test_run_at`: timestamp of when the test was run.
- `test_result`: whether the test is passing or failing.
- `severity`: if this info is available (I know it is for freshness but maybe not for all).
- info that allows us to generate a fully qualified table name:
    - `database`
    - `schema`
    - `table`
- `tool_name`: in this case it would be `dbt` (this information will be used by the Soda Cloud UI to display the test provenance and govern available attributes and details)
[...]

PR #572, opened 2021-11-19
Add dbt package to parse the dbt `manifest.json` and `run_results.json`

### LINKED ISSUE #571: Add dbt package to parse dbt artifacts
Add a package to parse the dbt artifacts. Part of resolving #570.

JCZuurmond:
Resolves #571. This PR adds a dbt package to parse the `manifest.json` and the `run_results.json`. 

I will create a separate PR with changes to the core, following the following structure:
1. Create soda cli command to ingest dbt test results.
2. Parses the artifacts
3. Maps them to soda [test results](https://github.com/sodadata/soda-sql/blob/75248e2a4b4bb4b05f3597f5b69f22a5073dad70/core/sodasql/scan/test_result.py#L21)
4. Flushes them to the cloud, like [in the soda scan](https://github.com/sodadata/soda-sql/blob/75248e2a4b4bb4b05f3597f5b69f22a5073dad70/core/sodasql/scan/scan.py#L879)


bastienboutonnet (commented):
I'll leave it to @vijaykiran to review this thoroughly but as far as I can see, I really like the approach and also the fact that you're using dbt's internals to parse and enforce validation which is really nice. 

I think it will also make our life a lot easier if they change the schemas since they seem to provide json schemas we can leverage which is even better.

Might be a good idea to have a quick peek at whether those objects have or are going to change in `v1.0.0` which is in `rc` candidate now.

JCZuurmond:
Thanks @bastienboutonnet . 

note to self and before we merge: add `dbt` to requirements.txt so that it is picked up by the testing suite

packages/dbt/sodasql/dbt.py
```diff
@@ -0,0 +1,90 @@
...
+    NotImplementedError :
+        If the dbt schema is not equal to the V3 manifest
+
+    Source
+    ------
+    https://docs.getdbt.com/reference/artifacts/manifest-json
+    """
+    dbt_v3_schema = "https://schemas.getdbt.com/dbt/manifest/v3.json"
+    if manifest["metadata"]["dbt_schema_version"] != dbt_v3_schema:
+        raise NotImplementedError("Dbt manifest parsing only supported for V3 schema.")
```
vijaykiran, line 41: Maybe it's better to assert than raising NotImplementedError

packages/dbt/sodasql/dbt.py
```diff
@@ -0,0 +1,90 @@
...
+    NotImplementedError :
+        If the dbt schema is not equal to the V3 run results.
+
+    Source
+    ------
+    https://docs.getdbt.com/reference/artifacts/run-results-json
+    """
+    dbt_v3_schema = "https://schemas.getdbt.com/dbt/run-results/v3.json"
+    if run_results["metadata"]["dbt_schema_version"] != dbt_v3_schema:
+        raise NotImplementedError(
```
vijaykiran, line 83: Same comment as above

vijaykiran (changes requested):
I'm not sure what we are testing in the tests is valuable enough, it looks like we test the python list comprehension. Otherwise, it LGTM!

packages/dbt/sodasql/dbt.py
JCZuurmond: Typically `raise` is used for user facing checks and `assert` for internal checks. The manifest and run_results are provided by the user, so I would keep a `raise`

packages/dbt/sodasql/dbt.py
JCZuurmond: see above

JCZuurmond:
Agreed on the testing, will keep it in for now so that I can use it in a PR for #570

vijaykiran (approved):

PR #577, opened 2021-11-25
Add ingest dbt tests

### LINKED ISSUE #570: Allow users to ingest dbt test results into Soda Cloud via soda-sql

JCZuurmond:
Fixes #570


core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+        tests_in_run_results = [
+            (run_result, test_nodes[run_result.unique_id])
+            for run_result in parsed_run_results
+            if run_result.unique_id in test_nodes.keys()
+        ]
+
+        tests = [
+            Test(
+                id=Parser.create_test_id(
+                    test_expression=test_node.compiled_sql if isinstance(test_node, CompiledSchemaTestNode) else None,
```
JCZuurmond, line 586: I do not understand how/why I get non-compiled nodes. I would expect that for tests being executed - and thus part of the run results - that they need to be compiled... I will have a further look at this.

core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+                    test_name=test_run_results.unique_id,
+                    test_index=index,
```
JCZuurmond, line 588: What is the test index exactly?

core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+                    context_column_name=test_node.column_name,
+                    context_sql_metric_name=None,
```
JCZuurmond, line 590: We do not have the sql matric name and index

core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+                    context_sql_metric_index=None,
+                ),
+                title=Parser.create_test_title(
```
JCZuurmond, line 593: I am thinking to change the title in something like: `f"dbt test - {test_run_results.unique_id}"`

core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+                expression=test_node.compiled_sql if isinstance(test_node, CompiledSchemaTestNode) else None,
+                metrics=None,
+                column=test_node.column_name,
+            )
+            for index, (test_run_results, test_node) in enumerate(tests_in_run_results)
+        ]
+
+        test_results = {
+            test_run_results.unique_id:
+            TestResult(
```
JCZuurmond, line 610: As value we can add the number of failed rows, not _really_ a value, but makes most sense (for v1)

core/sodasql/cli/cli.py
```diff
@@ -488,3 +498,143 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+            for (test_run_results, _), test in zip(tests_in_run_results, tests)
+        }
+
+        for model_unique_id, test_unique_ids in models_with_tests.items():
+            start_scan_response = soda_server_client.scan_start(
+                warehouse_yml.name,
+                warehouse_yml.dialect.type,
+                table_name=model_nodes[model_unique_id].alias,
+                scan_yml_columns=None,
+                scan_time=datetime.now().isoformat(),
```
JCZuurmond, line 624: the run results contain time per test, where we have a time per dataset. This also works, but is not _really_ the scan results: when the SQL is executed.

core/sodasql/scan/scan_builder.py
```diff
@@ -19,13 +19,95 @@
...
-from sodasql.scan.warehouse_yml_parser import read_warehouse_yml_file
+from sodasql.scan.warehouse_yml_parser import WarehouseYmlParser, read_warehouse_yml_file
 from sodasql.soda_server_client.soda_server_client import SodaServerClient
 from sodasql.scan.failed_rows_processor import FailedRowsProcessor
 
+
 logger = logging.getLogger(__name__)
 
 
+def build_warehouse_yml_parser(
```
JCZuurmond, line 30: I tried to reuse as much as possible from the Scan class. Took out some pieces and put them in separate functions, so that they can be used in different locations more easily

JCZuurmond (commented):
I'll move around some stuff and add some more testing. The essence is here though. Let me know if you have any remarks!

core/sodasql/cli/cli.py
JCZuurmond: This is changed, but it does not have the expected effect: the title in Soda cloud for monitor is not updated accordingly

core/sodasql/cli/cli.py
JCZuurmond: Added this, but it does **not** show up in the Soda cloud.

core/sodasql/scan/parser.py
```diff
@@ -337,6 +337,7 @@ def parse_test(self,
             self.error(f'Syntax error in test {test_title}:\n' +
                        ('\n'.join(stacktrace_lines[-3:])))
 
+    @classmethod
```
JCZuurmond, line 340: can be removed, not used any more

core/sodasql/cli/ingest.py
```diff
@@ -0,0 +1,189 @@
...
+    if tool == "dbt":
+        if dbt_manifest is None:
+            raise ValueError(f"Dbt manifest is required: {dbt_manifest}")
+        if dbt_run_results is None:
+            raise ValueError(f"Dbt run results is required: {dbt_run_results}")
+        test_results_iterator = create_dbt_test_results_iterator(dbt_manifest, dbt_run_results)
+    else:
+        raise ValueError(f"Unknown tool: {tool}")
+
+    for table_name, test_results in test_results_iterator:
```
JCZuurmond, line 174: move stuff below to separate flush test results function function

packages/dbt/sodasql/dbt.py
```diff
@@ -2,9 +2,15 @@
...
 from __future__ import annotations
 
+from collections import defaultdict
+from functools import reduce
+from operator import or_
 from typing import Any
 
-from dbt.contracts.graph.compiled import ParsedModelNode, CompiledSchemaTestNode
+from dbt.contracts.graph.compiled import (
+    CompiledModelNode, CompiledSchemaTestNode
```
JCZuurmond, line 11: note to self: double check which nodes `compiled` or `parsed` and other test besides `SchemaTest` we should expect.

core/sodasql/cli/ingest.py
```diff
@@ -0,0 +1,189 @@
...
+    dbt_run_results : Optional[Path]
+        The path to the dbt run results.
+
+    Raises
+    ------
+    ValueError :
+       If the tool is unrecognized.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info(SODA_SQL_VERSION)
```
JCZuurmond, line 156: Add more logging for cli user feedback. @vijaykiran : should we add the telemetry stuff here?

core/sodasql/cli/ingest.py
```diff
@@ -0,0 +1,189 @@
...
+    parsed_run_results = soda_dbt.parse_run_results(run_results)
+    tests_with_test_result = create_dbt_run_result_to_test_result_mapping(
+        test_nodes, parsed_run_results
+    )
+    models_with_tests = soda_dbt.create_models_to_tests_mapping(
+        model_nodes, test_nodes, parsed_run_results
+    )
+
+    for model_unique_id, test_unique_ids in models_with_tests.items():
+        table_name = model_nodes[model_unique_id].alias
```
JCZuurmond, line 121: double check if `alias` is correctly used

JCZuurmond (commented):
Structure wise the code is getting in a better shape. Most importantly is testing. Two main things that do not work as expected:

1. The title is in Soda cloud is not the same as the title given to the tests.
2. The number of failures is not shown. Maybe because there are no metrics linked?

core/sodasql/cli/ingest.py
```diff
@@ -0,0 +1,189 @@
...
+from sodasql.scan.test_result import TestResult
+
+
+
+def create_dbt_run_result_to_test_result_mapping(
+    test_nodes: dict[str, CompiledSchemaTestNode],
+    run_results: list[RunResultOutput],
+):
+    """
+    Create a mapping with which test result belong to which run result.
```
bastienboutonnet, line 30: I think this rephrasing might be easier to read:

Map test results to run result

core/sodasql/cli/ingest.py
bastienboutonnet: I believe we will want to indeed.

vijaykiran (approved):

core/sodasql/cli/ingest.py
```diff
@@ -0,0 +1,189 @@
...
+        A mapping from run result to test result.
+    """
+    from dbt.contracts.results import TestStatus
+    from dbt.contracts.graph.compiled import CompiledSchemaTestNode
+
+    dbt_tests_with_soda_test = {
+        test_node.unique_id:
+        Test(
+            id=Parser.create_test_id(
+                test_expression=test_node.compiled_sql if isinstance(test_node, CompiledSchemaTestNode) else None,
```
bastienboutonnet, line 53: Because the compiled sql can change from result to result (especially for user defined macro tests), pushing the expression in here will lead to an entirely new test to be created on soda SQL because of how our model works.

We should find another way to generate the ID. As far as I can see, we could just as well leave it as None, or something else.

bastienboutonnet (commented):
From a chat with @Antoninj we will need to change the logic of how we create the test id (see comments in the code)

We might also want to make sure we wait for the backend API to be able to accept the source of a test. @Antoninj is working on it.

Finally, I'll add another issue to add support for dbt's `store_failures` but that's out of the initial scope of the PR so I won't be the annoying guy that increases the PR scope at the time that it's almost ready.

core/sodasql/cli/ingest.py
JCZuurmond: Then I'll leave it up to you or @vijaykiran to do this, I'm ok with someone making those changes to this PR.

core/sodasql/cli/ingest.py
JCZuurmond: ah, oke, that is good to know. If can confirm that `None` is fine, than that is ok for me. Another option is the raw sql:

```
"raw_sql": "{{ test_accepted_values(**_dbt_schema_test_kwargs) }}{{ config(alias='accepted_values_stg_soda__scan') }}",
```

core/sodasql/cli/ingest.py
JCZuurmond: According to the [dbt docs](https://docs.getdbt.com/docs/building-a-dbt-project/building-models/using-custom-aliases/), the `alias` is what ultimately defines the table in the warehouse, therefore we should use it.

core/sodasql/soda_server_client/soda_server_client.py
```diff
@@ -42,41 +42,75 @@ def __init__(self,
...
+                    'minValue': validity.min,
+                    'maxValue': validity.max
+                }
+            soda_column_cfgs[column_name] = soda_column_cfg
+
         return self.execute_command({
             'type': 'sodaSqlScanStart',
-            'warehouseName': warehouse.name,
-            'warehouseType': warehouse.dialect.type,
-            'warehouseDatabaseName': database_and_schema.get('database_name'),
```
JCZuurmond, line 77: @vijaykiran : I am not really happy with this change, the schema and database require to have a warehouse object, and some warehouses are not lazily loaded (like parsing connection details or creating a connection). This forces that the ingest has to have a working warehouse set-up, where I think ingest only should have a working Soda cloud connection, besides the dbt artifacts.

core/sodasql/soda_server_client/soda_server_client.py
```diff
@@ -42,41 +42,75 @@ def __init__(self,
...
+        """
+        Start the scan.
+
+        Parameters
+        ----------
+        warehouse_name : str
+            The warehouse name.
+        warehouse_type : str
+            The warehouse (and dialect) type.
+        warehouse_database_name : Optional[str]
```
JCZuurmond, line 65: I moved these to the arguments, so that they do not require the `warehouse` object

packages/dbt/sodasql/dbt.py
JCZuurmond: For now, I think we are ok, the dbt v1 might cause troubles, see [release notes](https://github.com/dbt-labs/dbt-core/releases/tag/v1.0.0) on tests:

> The two type of test definitions are now "singular" and "generic" (instead of "data" and "schema", respectively). The test_type: selection method accepts test_type:singular and test_type:generic. (It will also accept test_type:schema and test_type:data for backwards compatibility) (#3234, #3880). Not backwards compatible: The --data and --schema flags to dbt test are no longer supported, and tests no longer have the tags 'data' and 'schema' automatically applied.

JCZuurmond: More general: I asked around in the [dbt slack](https://getdbt.slack.com/archives/C01F91XJ5PY/p1638266195027300) about another possible entrypoint for parsing the artifacts. Conclusion is: that using the node objects is the way to go - for now

core/sodasql/cli/cli.py
JCZuurmond: I do not have a complete answer for this yet, the `if` statement solves the issue though, so I am approaching this as non-urgent. Moreover, we need to decide what is a proper way to create the test id, as mentioned in one of the comments below, maybe we should remove the `compiled_sql` all together.

JCZuurmond:
Ok, lights are green again, PR is in sync with main. Main issue to resolve: what to do with the test id.

Furthermore, some points that might be covered in another PR:
- add logging to the `ingest` cli function
- add stored failures
- improve integration with dbt cloud, like showing the number of failed rows as measure

core/sodasql/cli/ingest.py
JCZuurmond: the current `create_test_id` returns something like:

```
        "id": '{"column":"result","test_name":"test.soda.accepted_values_stg_soda__scan__result__pass_fail.81f"}',
```

Is that to be expected?

bastienboutonnet: For the ID, I think the raw SQL may be ok but then again it may change as well. I would prefer to go by the test name and associated column or so.

I'm not sure I understand the latest comment.

JCZuurmond: The last comment shows what the [`create_test_id` method](https://github.com/sodadata/soda-sql/blob/dc31150e3de38060683d27ed1e5bff5b6ec2802e/core/sodasql/scan/parser.py#L357) generates for the test data. It was a fyi, even though more fields are mentioned in [`create_test_id`](https://github.com/sodadata/soda-sql/blob/dc31150e3de38060683d27ed1e5bff5b6ec2802e/core/sodasql/scan/parser.py#L357) for the test data the `column` and `test_name` are the only two that are included.

JCZuurmond:
After discussion with @Antoninj, we decided to use the `unique_id` as test id. The Soda cloud requires a unique identification id for the test, which is the case for the `unique_id` created by dbt.

core/sodasql/cli/cli.py
JCZuurmond: not relevant anymore.

JCZuurmond:
@vijaykiran : Could you wait with merging, I still have some changes that I would like to make.

vijaykiran:
@JCZuurmond  I marked the PR as draft, when you are ready please switch it  out of draft.

bastienboutonnet (commented):
@JCZuurmond are you planning to pass the source (new property now available on backend) in another PR? Personally, I wouldn't want us to start ingesting tests from dbt until they are labelled as coming from dbt so if you do it in another PR I would maybe hold off merging this.

Although, ultimately that's up to @vijaykiran since we are not tagging yet no one is going to use it in production but might be good to make sure we align those.

vijaykiran:
@bastienboutonnet  I'll merge this since I don't want to wait too long. I plan to tag a new release on Tuesday - as long as we can resolve the open issues then we should be good to go.

PR #592, opened 2021-12-10
Force use of warehouse class

JCZuurmond:
Copy of #497. But not from a fork to make the CI work.


vijaykiran:
Closing this - we will consider redesigns for next major version.

PR #598, opened 2021-12-14
feat: add dbt as source, fix test names, use dbt v1 schemas

bastienboutonnet:
Closes https://github.com/sodadata/soda-sql/issues/597

### DIFF
--- a/core/sodasql/scan/test_result.py
+++ b/core/sodasql/scan/test_result.py
@@ -28,43 +28,49 @@ class TestResult:
 
     def __str__(self):
         if self.passed:
-            status_str = 'passed'
+            status_str = "passed"
         elif self.skipped:
-            status_str = 'skipped'
+            status_str = "skipped"
         else:
-            status_str = 'failed'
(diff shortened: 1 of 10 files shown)


core/sodasql/scan/test_result.py
```diff
@@ -28,43 +28,49 @@ class TestResult:
...
         }
 
         if self.test.column:
-            test_result_json['columnName'] = self.test.column
+            test_result_json["columnName"] = self.test.column
+
+        if self.test.source:
+            test_result_json["source"] = self.test.source
+        else:
+            test_result_json["source"] = "soda-sql"
```
Antoninj, line 59: This is never going to get executed since we added a default value to source property. We either need to change the default value to `soda-sql` or add it to the test creation inside parser.py (L329 - L333)

vijaykiran: @Antoninj I added a default value, and  informed @bastienboutonnet to fix it there - https://github.com/sodadata/soda-sql/blob/main/core/sodasql/scan/test.py#L27

PR #606, opened 2021-12-16
[Spark] Use qualified column name in describe column

JCZuurmond:
Added the qualified column for Spark in #591. It should be used in the `describe_column` method.


vijaykiran (approved):

JCZuurmond:
The license/cla is stuck for some reason. @vijaykiran: can you reset it?

PR #607, opened 2021-12-16
Add ingest artifacts from dbt cloud

### LINKED ISSUE #589: [FEATURE] Soda ingest dbt artifacts from the dbt cloud
We would like to ingest the dbt artifacts from the dbt cloud.

Ability to ingest the dbt artifacts from the dbt cloud. 

We can use the [dbt cloud API](https://docs.getdbt.com/dbt-cloud/api-v2#section/How-to-use-this-API) for this. Rough logic would be:

0. [Authenticate](https://docs.getdbt.com/dbt-cloud/api-v2#section/Authentication)
1. [Get the runs](https://docs.getdbt.com/dbt-cloud/api-v2#operation/listRunsForAccount)
2. Filter runs for which we would like to ingest the artifacts, e.g. based on:
    - time window
    - job name 
    - id 
    The filter info should be provided by the user, we could have a default time window, though.
3. [Get the run artifacts](https://docs.getdbt.com/dbt-cloud/api-v2#operation/getArtifactsByRunId)
4. Use the ingest dbt [logic](https://github.com/sodadata/soda-sql/pull/577) to upload the test results to the Soda cloud.

At minimum the user needs to provide the API token and the account id.

To consider: how to handle the paging of the API.

**Additional context**
I think that the tricky part will be how to decide which runs to upload the artifacts for. If users have an advanced job structure that run sporadic - e.g. based on a trigger for when data is ingested into their warehouse - instead of on a (daily) schedule, it can become difficult to decide what artifacts to ingest. We should definitely check first if the alternative below works.

**Alternative**
Test if a user can install `soda` in their job definition and add the `soda ingest dbt` command there. If so, then we do not need this issue per se.

JCZuurmond:
Ingest the artifacts for a specific job run.

Fixes #589 .

### DIFF
--- a/core/sodasql/cli/cli.py
+++ b/core/sodasql/cli/cli.py
@@ -525,6 +525,18 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
     default=None,
     type=Path,
 )
+@click.option(
+    "--dbt-cloud-account-id",
+    help="The id of your dbt cloud account",
+    default=None,
+    type=Path,
+)
+@click.option(
+    "--dbt-cloud-run-id",
+    help="The id of the dbt job run of which you would like to ingest the test results",
+    default=None,
+    type=Path,
+)
 def ingest(*args, **kwargs):
     """
     Ingest test information from different tools.
--- a/core/sodasql/cli/ingest.py
+++ b/core/sodasql/cli/ingest.py
@@ -9,8 +9,10 @@ import datetime as dt
 import json
 import logging
 import os
+import requests
 from pathlib import Path
-from typing import Iterator, Optional, Tuple
+from typing import Iterator
+from requests.structures import CaseInsensitiveDict
 
 from sodasql.__version__ import SODA_SQL_VERSION
 from sodasql.scan.scan_builder import (
@@ -21,6 +23,7 @@ from sodasql.scan.test import Test
 from sodasql.scan.test_result import TestResult
 from sodasql.soda_server_client.soda_server_client import SodaServerClient
 
+KEY_DBT_CLOUD_API_TOKEN = "dbt_cloud_api_token"
 
 @dataclasses.dataclass(frozen=True)
 class Table:
@@ -78,17 +81,17 @@ def map_dbt_run_result_to_test_result(
 
 
 def map_dbt_test_results_iterator(
-    manifest_file: Path, run_results_file: Path
+    manifest: dict, run_results: dict
 ) -> Iterator[tuple[Table, list[TestResult]]]:
     """
     Create an iterator for the dbt test results.
 
     Parameters
     ----------
-    manifest_file : Path
-        The path to the manifest file.
-    run_results_file : Path
-        The path to the run results file.
+    manifest : dict
+        The manifest.
+    run_results : dict
+        The run results
 
     Returns
     -------
@@ -102,11 +105,6 @@ def map_dbt_test_results_iterator(
         raise RuntimeError(
             "Soda SQL dbt extension is not installed: $ pip install soda-sql-dbt"
         ) from e
-
-    with manifest_file.open("r") as file:
-        manifest = json.load(file)
-    with run_results_file.open("r") as file:
-        run_results = json.load(file)
 
     model_nodes, seed_nodes, test_nodes, source_nodes = soda_dbt.parse_manifest(manifest)
     parsed_run_results = soda_dbt.parse_run_results(run_results)
@@ -153,7 +151,9 @@ def flush_test_results(
     """
     for table, test_results in test_results_iterator:
         test_results_jsons = [
-            test_result.to_dict() for test_result in test_results if not test_result.skipped
+            test_result.to_dict()
+            for test_result in test_results
+            if not test_result.skipped
         ]
         if len(test_results_jsons) == 0:
             continue
@@ -174,25 +174,104 @@ def flush_test_results(
         soda_server_client.scan_ended(start_scan_response["scanReference"])
 
 
-def resolve_artifacts_paths(
-    dbt_artifacts: Optional[Path] = None,
-    dbt_manifest: Optional[Path] = None,
-    dbt_run_results: Optional[Path] = None
-) -> Tuple[Path, Path]:
-    if dbt_artifacts:
-        dbt_manifest = Path(dbt_artifacts) / 'manifest.json'
-        dbt_run_results = Path(dbt_artifacts) / 'run_results.json'
-    elif dbt_manifest is None:
-        raise ValueError(
-            "--dbt-manifest or --dbt-artifacts are required. "
-            f"Currently, dbt_manifest={dbt_manifest} and dbt_artifacts={dbt_artifacts}"
-        )
-    elif dbt_run_results is None:
-        raise ValueError(
-            "--dbt-run-results or --dbt-artifacts are required. "
-            f"Currently, dbt_run_results={dbt_manifest} and dbt_artifacts={dbt_artifacts}"
-        )
-    return dbt_manifest, dbt_run_results
+def load_dbt_artifacts(
+    manifest_file: Path,
+    run_results_file: Path,
+) -> tuple[dict, dict]:
+    """
+    Resolve artifacts.
+
+    Arguments
+    ---------
+    manifest_file : Path
+        The manifest file.
+    run_results_file : Path
+        The run results file.
+
+    Return
+    ------
+    out : tuple[dict, dict]
+        The loaded manifest and run results.
+    """
+    with manifest_file.open("r") as file:
+        manifest = json.load(file)
+    with run_results_file.open("r") as file:
+        run_results = json.load(file)
+    return manifest, run_results
+
+
+def download_dbt_artifact_from_cloud(
+    artifact: str,
+    api_token: str,
+    account_id: str,
+    run_id: str,
+) -> dict:
+    """
+    Download an artifact from the dbt cloud.
+
+    Parameters
+    ----------
+    artifact : str
+        The artifact name.
+    api_token : str
+        The dbt cloud API token.
+    account_id: str :
+        The account id.
+    run_id :  str
+        The run id.
+
+    Returns
+    -------
+    out : dict
+        The artifact.
+
+    Sources
+    -------
+    https://docs.getdbt.com/dbt-cloud/api-v2#operation/getArtifactsByRunId
+    """
+    url = f"https://cloud.getdbt.com/api/v2/accounts/{account_id}/runs/{run_id}/artifacts/{artifact}"
+
+    headers = CaseInsensitiveDict()
+    headers["Authorization"] = f"Token {api_token}"
+    headers["Content-Type"] = "application/json"
+
+    response = requests.get(url, headers=headers)
+
+    if response.status_code != requests.codes.ok:
+        response.raise_for_status()
+
+    return response.json()
+
+
+def download_dbt_artifacts_from_cloud(
+    api_token: str,
+    account_id: str,
+    run_id: str,
+) -> tuple[dict, dict]:
+    """
+    Download the dbt artifacts from the cloud.
+
+    Parameters
+    ----------
+    api_token : str
+        The dbt cloud API token.
+    account_id : str
+        The account id.
+    run_id :  str
+        The run id.
+
+    Returns
+    -------
+    out : tuple[dict, dict]
+        The loaded artifacts.
+    """
+    manifest = download_dbt_artifact_from_cloud(
+        "manifest.json", api_token, account_id, run_id
+    )
+    run_results = download_dbt_artifact_from_cloud(
+        "run_results.json", api_token, account_id, run_id
+    )
+    return manifest, run_results
 
 
 def ingest(
@@ -201,6 +280,8 @@ def ingest(
     dbt_artifacts: Path | None = None,
     dbt_manifest: Path | None = None,
     dbt_run_results: Path | None = None,
+    dbt_cloud_account_id: str | None = None,
+    dbt_cloud_run_id: str | None = None,
 ) -> None:
     """
     Ingest test information from different tools.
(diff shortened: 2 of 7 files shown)


core/sodasql/cli/cli.py
```diff
@@ -525,6 +525,24 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
     default=None,
     type=Path,
 )
+@click.option(
```
JCZuurmond, line 528: it would be nice to move these arguments to a sub command:

```
soda ingest dbt
```

And only show the dbt flag when the `dbt` is given.

packages/dbt/sodasql/dbt.py
```diff
@@ -56,9 +57,11 @@ def parse_manifest(
     ------
     https://docs.getdbt.com/reference/artifacts/manifest-json
     """
+    logger = logging.getLogger(__name__)
+
     dbt_v4_schema = "https://schemas.getdbt.com/dbt/manifest/v4.json"
     if manifest["metadata"]["dbt_schema_version"] != dbt_v4_schema:
-        raise NotImplementedError("Dbt manifest parsing only supported for V4 schema.")
+        logger.warning("Dbt manifest parsing only tested for v4 schema.")
```
JCZuurmond, line 64: parsing worked for `v3` too, did not add a proper test for that, tested it manually

JCZuurmond (commented):
Most is in there. Adding testing is most important.

bastienboutonnet (commented):
Awesome @JCZuurmond !

Few questions and remarks, otherwise seems good to me but I'll defer to @vijaykiran for approval and to answer some of your other design questions.

core/sodasql/cli/cli.py
```diff
@@ -525,6 +525,24 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+    "--dbt-cloud-api-token",
```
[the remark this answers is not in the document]
JCZuurmond: I have it loaded as env var in my shell and reference it like `$MY_SUPER_SECRET_API_KEY`

packages/dbt/sodasql/dbt.py
JCZuurmond: Dunno exactly, I tested it with a v4 schema, not sure if that came from dbt v1.

tests/cli/test_ingest.py
```diff
@@ -117,54 +120,79 @@ def test_dbt_flush_test_results_soda_server_scan_test_result(
...
-            "",
-            Path('my_dbt_project/target/run_results.json'),
-            "",
+            None,
+            Path("my_dbt_project/target/manifest.json"),
+            None,
             id="missing dbt_run_results and artifact",
         ),
-    ]
+    ],
```
bastienboutonnet, line 138: shouldn't we add a test case for the scenario where we provide a `dbt_artifacts` only?

JCZuurmond: done!

core/sodasql/cli/cli.py
JCZuurmond: I will have a go at this in a separate PR. With argparse I would do this by adding a [sub parser](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers). Most likely `click` has a similar option

bastienboutonnet: Sounds good to me yeah, seems to be more of an improvement on design rather than a functional need for this PR. I don't really know click well (nor am I a fan of it so I've stayed away from it)

core/sodasql/cli/cli.py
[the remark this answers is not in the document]
bastienboutonnet: I think I'm missing something then. I don't see this code change parsing any env vars related to the API token of dbt cloud.

I would expect that our env var file parsing now also tries to look for something like `DBT_CLOUD_API_TOKEN`.

I don't think it makes a lot of sense to feed that via the CLI. If anything, because it's inconsistent with how soda-sql gets this kind of thing (cf. soda cloud token etc).

My guess would be that it's better to have it in there and simply remove this option from the CLI.

I think env_vars parsing is happening in this file: https://github.com/sodadata/soda-sql/blob/main/core/sodasql/scan/env_vars.py

But maybe check with @vijaykiran for advice.

packages/dbt/sodasql/dbt.py
bastienboutonnet: Just so I understand a bit more and see if I can help.

Do I understand well that you made a new dbt project with dbt v1 and got your artifacts to be v4 by default?

tests/cli/test_ingest.py
bastienboutonnet: Are you sure you pushed your change? I don't see the test case I'm expecting in the file

packages/dbt/sodasql/dbt.py
JCZuurmond: I don't know what version for dbt was used to create the artifacts that are found in the test folder. 

What is the current scenario: we test with a v4 artifact schema. 

The question is: are we going to test for all versions of the artifact schema? Or do we leave it like this and do a regression test once we find a schema for which the conversion breaks. Note: we can constrain the versions of the artifacts by setting the version of the dbt dependency e.g. `dbt>=1.0` and we use dbt internal data classes so that should help with the conversion - if the user uses the same dbt version

core/sodasql/cli/cli.py
JCZuurmond: See [here](https://github.com/sodadata/soda-sql/pull/608)

core/sodasql/cli/cli.py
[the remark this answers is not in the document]
vijaykiran: I don't think we should encourage adding keys as parameters, it would be consistent with other parameters (passwords/tokens storing in the .env_vars). They can always be loaded from shell environment vars by pointing them in the yml file though.

core/sodasql/cli/ingest.py
```diff
@@ -172,25 +173,125 @@ def flush_test_results(
...
-    dbt_artifacts: Optional[Path] = None,
-    dbt_manifest: Optional[Path] = None,
-    dbt_run_results: Optional[Path] = None
-) -> Tuple[Path, Path]:
-    if dbt_artifacts:
-        dbt_manifest = Path(dbt_artifacts) / 'manifest.json'
-        dbt_run_results = Path(dbt_artifacts) / 'run_results.json'
-    elif dbt_manifest is None:
+def load_dbt_artifacts(
+    artifacts_dir: Path | None = None,
```
vijaykiran, line 177: Isn't this 3.10 only ? type checkers working < 3.10 will get confused with this .. right ?

core/sodasql/cli/ingest.py
```diff
@@ -172,25 +173,125 @@ def flush_test_results(
...
+
+    return manifest, run_results
+
+
+def download_dbt_artifact_from_cloud(
+    artifact: str,
+    api_token: str,
+    account_id: str,
+    run_id: str,
+) -> dict():
```
vijaykiran, line 228: This should be just `dict`

core/sodasql/cli/ingest.py
```diff
@@ -232,13 +342,35 @@ def ingest(
...
+                dbt_run_results,
+            )
+        else:
+            if dbt_cloud_api_token is None:
+                raise ValueError(
+                    f"Expecting an DBT cloud api token. Currently, dbt_cloud_api_token={dbt_cloud_api_token}."
+                )
+            elif dbt_cloud_account_id is None:
+                raise ValueError(
+                    f"Expecting an DBT cloud api token. Currently, dbt_cloud_account_id={dbt_cloud_account_id}."
```
vijaykiran, line 363: `Expecting dbt_cloud_account_id` , I don't think Currently is needed, maybe just rewrite to `dbt_cloud_account_id not provided`

core/sodasql/cli/ingest.py
```diff
@@ -232,13 +342,35 @@ def ingest(
...
-    if tool == 'dbt':
-        dbt_manifest, dbt_run_results = resolve_artifacts_paths(
-            dbt_artifacts=dbt_artifacts,
-            dbt_manifest=dbt_manifest,
-            dbt_run_results=dbt_run_results
-        )
-        test_results_iterator = map_dbt_test_results_iterator(dbt_manifest, dbt_run_results)
+    if tool == "dbt":
+        if (
+            dbt_artifacts is not None
```
vijaykiran, line 347: Shouldn't this be 
```
 dbt_artifacts is not None
 or (dbt_manifest is not None and dbt_run_results is not None)
```
?

core/sodasql/cli/ingest.py
```diff
@@ -232,13 +342,35 @@ def ingest(
...
+            if dbt_cloud_api_token is None:
+                raise ValueError(
+                    f"Expecting an DBT cloud api token. Currently, dbt_cloud_api_token={dbt_cloud_api_token}."
+                )
+            elif dbt_cloud_account_id is None:
+                raise ValueError(
+                    f"Expecting an DBT cloud api token. Currently, dbt_cloud_account_id={dbt_cloud_account_id}."
+                )
+            elif dbt_cloud_run_id is None:
+                raise ValueError(
```
vijaykiran, line 366: Perhaps a better way is to collect the errors and then print them instead of failing early:

(untested code):

```
errors = [f"Invalid value provided for {x=}" for x in
          [dbt_cloud_api_token, dbt_cloud_account_id, dbt_cloud_run_id] if x is None]
if errors:
    raise ValueError("\n".join(errors))
else:
    manifest, run_results = download_dbt_artifacts_from_cloud(
        dbt_cloud_api_token, dbt_cloud_account_id, dbt_cloud_run_id
    )
```

packages/dbt/sodasql/dbt.py
```diff
@@ -107,9 +110,11 @@ def parse_run_results(run_results: dict[str, Any]) -> list[RunResultOutput]:
     ------
     https://docs.getdbt.com/reference/artifacts/run-results-json
     """
+    logger = logging.getLogger(__name__)
+
     dbt_v4_schema = "https://schemas.getdbt.com/dbt/run-results/v4.json"
     if run_results["metadata"]["dbt_schema_version"] != dbt_v4_schema:
-        raise NotImplementedError("Dbt run results parsing only supported for v4 schema.")
+        logger.warning("Dbt run results parsing only tested for v4 schema.")
```
vijaykiran, line 117: Let's use `dbt` (lowercase everywhere)

vijaykiran (changes requested):

packages/dbt/sodasql/dbt.py
bastienboutonnet: I would enforce dbt >= 1 (I think that's already now the case in the `setup.py` file). 

I personally don't want to really do version checks and import checks unless we have to. v4 schemas have now changed:
- `CompiledTestNode` to `CompiledGenericTestNode`
- `ParsedTestNode` to `ParsedGenericTestNode`

if we support < v4 it means we have to import different classes from the  `contracts` module and although we will have to do it one day, the risk of it changing throughout the lifetime of dbt v1 is unlikely.

bastienboutonnet: What we could do is check that dbt v1 always generates v4 schemas and in that case we should be good.

core/sodasql/cli/ingest.py
JCZuurmond: the `from __future__ import annotations` allows to do this in earlier versions

core/sodasql/cli/ingest.py
JCZuurmond: nope, that is handled in the `load_dbt_artifacts` (previously the `resolve_dbt_artifacts`)

core/sodasql/cli/ingest.py
JCZuurmond: done

packages/dbt/sodasql/dbt.py
JCZuurmond: done

core/sodasql/cli/ingest.py
JCZuurmond: hmm, yes, I would actually prefer to push this down to the click level, something [like this](https://stackoverflow.com/questions/55584012/python-click-dependent-options-on-another-option) maybe

vijaykiran: While click dealing with parameter errors is good, it won't help when we expose API for dbt ingestion. So I think it is better to change at the code at this level.

core/sodasql/cli/ingest.py
vijaykiran: Ok - I'd expect the validation to happen in `download_dbt_artifacts_from_cloud` function instead of here though.

core/sodasql/cli/ingest.py
```diff
@@ -232,13 +342,29 @@ def ingest(
...
+            or dbt_run_results is not None
+        ):
+            manifest, run_results = load_dbt_artifacts(
+                dbt_artifacts,
+                dbt_manifest,
+                dbt_run_results,
+            )
+        else:
+            if dbt_cloud_api_token is None:
+                raise ValueError(f"Expecting an dbt cloud api token: {dbt_cloud_api_token}")
```
vijaykiran, line 358: `s/an dbt/a dbt/`

core/sodasql/cli/ingest.py
JCZuurmond: ok, I'll change it

core/sodasql/cli/ingest.py
JCZuurmond: done

packages/dbt/sodasql/dbt.py
JCZuurmond: The dbt cloud we use for testing uses the dbt version 1.0.0. This returns a schema v4. I removed the checks, and we solve this by the `dbt` version constrain.

core/sodasql/cli/cli.py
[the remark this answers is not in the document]
JCZuurmond: In that case: could you specify where exactly the environment variable should go? Are we going to add a new section for this? Or add it to the soda account?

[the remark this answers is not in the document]
bastienboutonnet: Personally, I would indeed expect them to be in the .env_vars file as we do for other passwords and tokens and make the ingest or dbt package read from that env var?

@vijaykiran, is that also in line with your thinking?

vijaykiran:
Note to self: Use env_vars for the dbt token stuff.

vijaykiran (approved):

PR #609, opened 2021-12-17
Add main module

JCZuurmond:
It's the more [pythonic way](https://docs.python.org/3/library/__main__.html) of calling a package. It allows to use Soda like: `python -m sodasql`.


JCZuurmond:
Closing this PR, created #611 that is not pointing to my fork

PR #610, opened 2021-12-17
Let ingest have subcommands

JCZuurmond:
This allows us to have flags only shown when a user wants to ingest for that tool:

```
soda ingest dbt --help      # shows flags for dbt
```

```
soda ingest other_tool --help      # shows flags for other_tool
```


core/sodasql/cli/cli.py
```diff
@@ -493,19 +493,22 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
...
+    """Commands for ingesting test information from other tools."""
+
+
+@ingest.command(
+    name="dbt",
+    short_help="Ingest test information from different tools"
+)
 @click.option(
-    "--dbt-artifacts",
+    "--artifacts-directory",
```
bastienboutonnet, line 511: I'm not sure I buy into the rename here but maybe I need to check the following: is it the case that all if we were to use a different sub-command the set of options would be different and therefore we do not need to make the options bear a tool-specific name?

One thing though, is that since we've already released this with the old names, not offering support for the old name would be a breaking change, which we don't want. I'm pretty sure there's a way to make click accept different versions of an option that map onto the same one right? If so, would we temporarily allow for that? Then What I would do is throw a deprecation warning that `--dbt-` CLI options will stop being supported in version x and make sure that we deprecate it in said version.

core/sodasql/cli/cli.py
```diff
@@ -514,24 +517,24 @@ def scan(scan_yml_file: str, warehouse_yml_file: str, variables: tuple, time: st
     type=Path,
 )
 @click.option(
-    "--dbt-manifest",
+    "--manifest-file",
```
bastienboutonnet, line 520: does click do autocomplete? if so, maybe it's ok if not I would air on the side of shorter option names (which I know is not ideal from a clear internal naming point of view but I wouldn't be surprised if click had a way to internally map something to something else. I think I've done it with argparse before.

core/sodasql/cli/ingest.py
```diff
@@ -193,28 +193,25 @@ def resolve_artifacts_paths(
...
-    dbt_run_results: Path | None = None,
+def ingest_dbt(
+    warehouse_yml_file: Path,
+    artifacts_directory: Path | None = None,
+    manifest_file: Path | None = None,
+    run_results_file: Path | None = None,
 ) -> None:
     """
-    Ingest test information from different tools.
+    Ingest DBT test information.
```
bastienboutonnet, line 203: dbt must always be lowercase

core/sodasql/cli/ingest.py
```diff
@@ -232,15 +229,12 @@ def ingest(
...
-    if tool == 'dbt':
-        dbt_manifest, dbt_run_results = resolve_artifacts_paths(
-            dbt_artifacts=dbt_artifacts,
-            dbt_manifest=dbt_manifest,
-            dbt_run_results=dbt_run_results
-        )
-        test_results_iterator = map_dbt_test_results_iterator(dbt_manifest, dbt_run_results)
-    else:
-        raise NotImplementedError(f"Unknown tool: {tool}")
+    dbt_manifest, dbt_run_results = resolve_artifacts_paths(
```
bastienboutonnet, line 232: I like that refactoring to a subcommand made this a lot less full of if/elses! 👌

bastienboutonnet (commented):
I have a few almost nitpicky comments that I would personally want to see addressed, but I'll leave it to @vijaykiran for final words

core/sodasql/cli/cli.py
JCZuurmond: > is it the case that all if we were to use a different sub-command the set of options would be different and therefore we do not need to make the options bear a tool-specific name?

Yes, with the set-up in this PR you can have `dbt` specific flags, with the "old" set-up this was not possible. Therefore I decided to remove the `dbt` prefix, since it is redundant now.

>One thing though, is that since we've already released this with the old names, not offering support for the old name would be a breaking change, which we don't want. I'm pretty sure there's a way to make click accept different versions of an option that map onto the same one right? If so, would we temporarily allow for that? Then What I would do is throw a deprecation warning that --dbt- CLI options will stop being supported in version x and make sure that we deprecate it in said version.

Ok, true, but wasn't it a "pre-release" for dbt ingestion. Therefore it is ok to change the flags, but this is ultimately a decision up to you - or Vijay.

core/sodasql/cli/cli.py
JCZuurmond: with shorter you mean `--dbt-manifest` to `--manifest` without the `-file` suffix?

core/sodasql/cli/cli.py
bastienboutonnet: Don't get me wrong, I like and think that the renaming makes sense given that we don't need to specify `dbt` anymore.

The issue I have is that this dbt functionality is in a non pre-release tag. We may say it's pre-release but not according to semver. So, at the very least I would just offer the two as possibilities via:

```python
@click.option("--dbt-artifacts", "artifacts")
```

And then later, we can decide to just deprecate the one that we think is silly.

That being said, I would like @vijaykiran to make a call on this.

core/sodasql/cli/cli.py
bastienboutonnet: Yeah maybe? What do you think? I like the specificity but I also am lazy. I leave it up to you! we can always add short versions of those if users complain.

PR #611, opened 2021-12-17
Add main module

JCZuurmond:
It's the more [pythonic way](https://docs.python.org/3/library/__main__.html) of calling a package. It allows to use Soda like: `python -m sodasql`.


bastienboutonnet (commented):
As far as I'm concerned this makes sense to me. I leave it up @vijaykiran for final approval.

vijaykiran (approved):

PR #616, opened 2021-12-23
Make cryptography dependency only for snowflake package

sebastianvillarroel:
`cryptography` is only used in the `snowflake` package. It would be best to take this dependency away from the core installation.

I came up with this as I wanted to install soda in Google Cloud Composer v2 latest version. Unfortunately, it was not possible due to the `cryptography` dependency. I hope this helps!

Under the hood: correct a typo in the contribution guidelines.


vijaykiran:
testing via #619

PR #618, opened 2021-12-28
[bigquery] Introduce context_auth

vijaykiran:
The authentication can now be specified as context auth


vijaykiran:
@all-contributors please add @IIBenII for code

PR #646, opened 2022-01-10
fix: resolve key access error when sources are not present in manifest

### LINKED ISSUE #641: Maybe revise the hard key access for `source` nodes in dbt integration
**Describe the bug**
If a `manifest.json` does not have nodes of type `sources` the parsing can throw a key error:
```
11:40:11.599 INFO  SodaSqlCon | STDERR:   File "/usr/local/lib/python3.8/site-packages/sodasql/dbt.py", line 83, in parse_manifest
11:40:11.600 INFO  SodaSqlCon | STDERR:     for source_name, source in manifest["sources"].items()
11:40:11.600 INFO  SodaSqlCon | STDERR: KeyError: 'sources'
```

We might want to do a "soft" access via a `.get()` if possible.

I didn't do further research into how feasible this is but if no one picks this up I'll have a closer look as soon as I have a minute

The issue was highlighted by @Antoninj who was writing some integration tests and had an outdated test file which did not contain any `source` nodes.

bastienboutonnet:
Resolves #641

### DIFF
--- a/core/sodasql/cli/ingest.py
+++ b/core/sodasql/cli/ingest.py
@@ -60,7 +60,7 @@ def map_dbt_run_result_to_test_result(
         A mapping from run result to test result.
     """
     from dbt.contracts.results import TestStatus
-    assert test_nodes, "No test nodes were retrieved from the manifest.json. This could be because no tests have been implemented in dbt yet or you never ran `dbt test`."
+    assert test_nodes is not None, "No test nodes were retrieved from the manifest.json. This could be because no tests have been implemented in dbt yet or you never ran `dbt test`."
     dbt_tests_with_soda_test = {
         test_node.unique_id: Test(
             id=test_node.unique_id,
--- a/core/sodasql/cli/ingest.py
+++ b/core/sodasql/cli/ingest.py
@@ -41,7 +41,7 @@ class Table:
 
 
 def map_dbt_run_result_to_test_result(
-    test_nodes: Optional[dict[str, "DbtTestNode"]],
+    test_nodes: dict[str, "DbtTestNode"] | None,
     run_results: list["RunResultOutput"],
 ) -> dict[str, set["DbtModelNode"]]:
     """
--- a/core/sodasql/cli/ingest.py
+++ b/core/sodasql/cli/ingest.py
@@ -41,7 +41,7 @@ class Table:
 
 
 def map_dbt_run_result_to_test_result(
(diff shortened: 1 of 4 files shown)


core/sodasql/cli/ingest.py
```diff
@@ -35,7 +35,7 @@ class Table:
 
 
 def map_dbt_run_result_to_test_result(
-    test_nodes: dict[str, "DbtTestNode"],
+    test_nodes: Optional[dict[str, "DbtTestNode"]],
```
vijaykiran, line 38: Looks like `Optional` isn't imported, can you add it?

Alternatively, you can also use shortcut syntax here `dict[str, "DbtTestNode"] | None`

packages/dbt/sodasql/dbt.py
```diff
@@ -25,15 +25,15 @@
 def parse_manifest(
     manifest: dict[str, Any]
 ) -> tuple[
-    dict[str, ParsedModelNode | CompiledModelNode],
-    dict[str, ParsedSeedNode | CompiledSeedNode],
-    dict[str, ParsedGenericTestNode | CompiledGenericTestNode],
-    dict[str, ParsedSourceDefinition],
+        Optional[dict[str, ParsedModelNode | CompiledModelNode]],
```
vijaykiran, line 28: Same comment as above for `Optional`s here

packages/dbt/sodasql/dbt.py
```diff
@@ -114,10 +118,9 @@ def parse_run_results(run_results: dict[str, Any]) -> list[RunResultOutput]:
     parsed_run_results = [RunResultOutput(**result) for result in run_results["results"]]
     return parsed_run_results
 
-
 def create_nodes_to_tests_mapping(
     model_nodes: dict[str, ParsedModelNode],
-    test_nodes: dict[str, CompiledGenericTestNode],
+    test_nodes: Optional[dict[str, CompiledGenericTestNode | ParsedGenericTestNode]],
```
vijaykiran, line 123: Same as above :)

vijaykiran (changes requested):

core/sodasql/cli/ingest.py
bastienboutonnet: That's strange that I didn't get an error here...
I'll change it to the more modern (i.e. without `Optional` though as I think it's better).

PR #680, opened 2022-03-03
Snowflake: quote columns in queries, fixes #679.

### LINKED ISSUE #679: Deal with reserved keywords in column names
**Describe the bug**
When a reserved keyword is used as a column name soda-sql does not quote the name when building queries which causes the queries to fail.

**To Reproduce**
```
table_name: CUSTOMERS
metrics:
  - missing_percentage
```
and use a table with column which is a reserved keyword, e.g. on Snowflake `VALUES`

Proposed solution:
quote column names in all queries. Case sensitivity should not be an issue, as column names are retrieved from the database so should represent case sensitivity accurately.

m1n0:


m1n0:
Skipping Athena for now as it requires more work - it needs different quoting strategy for different query types.
Keeping dialects that already quote all column untouched (postgresql) to avoid introducing unnecessary regression.

vijaykiran (approved):

PR #1170, opened 2022-03-25
Include threshold in check name

### LINKED ISSUE #1154: Metrics and thresholds wrong check name when warnings used
Given the `checks.yml` configuration:
```
checks for bitcoin:
  - min(usd):
      warn: when < 20000
      fail: when < 10000
```
expected check name:
```
min(usd) fails when < 10000
```
received check name:
```
min(usd)
```

m1n0:
Fix #1154


m1n0:
Takes warn and fail config if available and includes them in check name. This means that if these are changed it will create separate checks in the cloud, which seems desirable:
Screen Shot 2022-03-25 at 11 14 44 AM

vijaykiran (approved):

PR #1227, opened 2022-03-29
feat: anomaly detection check

bastienboutonnet:
# Anomaly Detection
- brings anomaly detection logic into the scientific package
- brings bare-bones version of anomaly check which will currently only work on `row_count` if such a metric exists on Soda Cloud or is derived by a separate check such as `row_count > 0`

Closes https://github.com/sodadata/soda-sql-v3/issues/254. Closes https://github.com/sodadata/soda-core/issues/1285

# Notes for Reviewers
@vijaykiran @m1n0 
- At this point, the only issue I have, and hoping we can solve when reviewing this PR is around logging.

- Currently, `debug` level logs make it to the stdout even when `soda scan` does not require elevated verbosity levels from the CLI.

- I am also wondering, whether we want to have some sort of logging pre-fixes for different checks. Anomaly Detection would benefit from throwing a few "progress" related log messages to stdout but currently, there is no standard way to do that other than deciding to prefix those messages with some string. Have we already thought about this at all? If not, we might want to. I don't think this is the point of this PR but we should have a follow-up soon after.

- Finally, unit tests are currently failing because of the change on the historical descriptor. @vijaykiran I think you were planning to open a PR to be merged in this one so that this can be solved?

@baturayo 
- Can you double-check the `AnomalyDetector` sections of this PR? This PR mainly copied and pasted code that you actually refactored from the private repo. However, I did have to make some changes here and there to make things work based on the new contract, so it would be good to have a look still.


soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
```diff
@@ -0,0 +1,263 @@
...
+    identity: str
+    value: float
+    dataTime: datetime.datetime
+
+
+class AnomalyHistoricalMeasurements(BaseModel):
+    results: List[AnomalyHistoricalMeasurement]
+
+
+MOCK_ANOMALY_DIAGNOSTICS = [
```
baturayo, line 88: maybe we can remove this part and move it to tests

soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
```diff
@@ -0,0 +1,263 @@
...
+                "warn": None,
+                "fail": None,
+                "anomalyProbability": None,
+                "anomalyPredictedValue": None,
+            }
+
+        diagnostics_dict: Dict[str, Any] = AnomalyDiagnostics.parse_obj(diagnostics).dict()
+        return level, diagnostics_dict
+
+    def evaluate(self) -> Tuple[str, Dict[str, Any]]:
```
baturayo, line 247: Very minor thing, but to increase the readability, we can move this method just next to __init method (i.e line 115)

soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
```diff
@@ -0,0 +1,263 @@
...
+
+        column_maps = self.params["request_params"]["columns_mapping"]
+        df_flattened = df_flattened[df_flattened.columns[df_flattened.columns.isin(list(column_maps.keys()))]]
+        df_flattened = df_flattened.rename(columns=column_maps)  # type: ignore
+        df_flattened["ds"] = pd.to_datetime(df_flattened["ds"])  # type: ignore
+        df_flattened["ds"] = df_flattened["ds"].dt.tz_localize(None)
+        return df_flattened
+
+    @staticmethod
+    def flatten_df(df: pd.DataFrame, target_col_name: str) -> pd.DataFrame:
```
baturayo, line 163: as far as I remember, this method is just used in this class. So, to make it consistent we can change the name from flatten_df to _flatten_df

soda/scientific/soda/scientific/anomaly_detection/feedback_processor.py
```diff
@@ -0,0 +1,169 @@
...
+    "Tuesday": "TUE",
+    "Wednesday": "WED",
+    "Thursday": "THU",
+    "Friday": "FRI",
+    "Saturday": "SAT",
+    "Sunday": "SUN",
+}
+
+
+class FeedbackProcessor:
```
baturayo, line 48: Again for this class, we can move `run` method to the top, and prepend private methods with `_`.

baturayo (approved):

soda/core/soda/execution/anomaly_metric_check.py
```diff
@@ -23,59 +28,62 @@ def __init__(
             column=column,
         )
 
+        self.skip_anomaly_check = False
         metric_check_cfg: MetricCheckCfg = self.check_cfg
+        if not metric_check_cfg.fail_threshold_cfg or not metric_check_cfg.warn_threshold_cfg:
```
vijaykiran, line 33: Is this intentional? Because it will shortcut if the `fail_threshold_cfg` is `None` and will skip even if `warn_threshold_cfg` is set.

vijaykiran: Maybe you meant `if not metric_check_cfg.fail_threshold_cfg and not metric_check_cfg.warn_threshold_cfg:` ?

bastienboutonnet: It was intentional, because when `default` is passed this actually gets stored as the `fail_threshold_cfg`, while the `warn_threshold_cfg` is optional so at this point and given the fact that I'm requiring users to use `default` this would work.

Unless you have another suggestion. I do feel, however, that an `and` approach would not be suitable since `warn` can legally be `None`

soda/scientific/soda/scientific/anomaly_detection/feedback_processor.py
bastienboutonnet: I'll move the run indeed. I might skip pretending the private methods for now. I think it's not that crucial and I would only do it if we decide we want to 100% enforce this convention. What do you think?

soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
bastienboutonnet: I have something in the tests so I think I'll just remove that, good catch.

soda/core/soda/execution/anomaly_metric_check.py
vijaykiran: okay - I wasn't sure what default was interpreted to be, in that case, It is fine.

PR #1264, opened 2022-04-08
Unify datasource metadata query.

m1n0:
#1175

### DIFF
--- a/soda/core/soda/execution/data_source.py
+++ b/soda/core/soda/execution/data_source.py
@@ -242,7 +242,7 @@ class DataSource:
     # Query Execution
     ######################
 
-    def get_row_counts_for_all_tables(
+    def get_row_counts_all_tables(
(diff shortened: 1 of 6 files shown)


soda/core/soda/execution/data_source.py
```diff
@@ -177,51 +183,104 @@ def sql_get_column(self, include_tables: list[str] | None = None, exclude_tables
...
+            return {row[0]: row[1] for row in query.rows}
+
+        # Single query to get the metadata not available, get the counts one by one.
+        all_tables = self.find_table_names()
+        result = {}
+
+        for table in all_tables:
+            query = Query(
+                data_source_scan=self,
+                unqualified_query_name=f"get_counts_by_tables_for_row_count_anomalies_{table}",
```
bastienboutonnet, line 268: Why is this called row count anomalies? It seems that this method should be agnostic no?

m1n0: thanks for catching that, changed.

PR #1266, opened 2022-04-08
Support overriding project id in BQ

### LINKED ISSUE #1265: Support overriding project id from SA in BQ
Project id from service account is used. When compute is separated from storage it required creating separate SAs for each project which is impractical. Support overriding project it that comes from SA explicitly for each data source.

m1n0:
Fix #1265


soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -64,8 +64,15 @@ def connect(self, connection_properties):
...
             # else:
             self.account_info_dict = self.__parse_json_credential()
-            if self.account_info_dict:
+
+            # Use explicitly set project id if available, or the one from SA account.
+            if connection_properties.get("project_id"):
+                self.project_id = connection_properties.get("project_id")
+            elif self.account_info_dict:
                 self.project_id = self.account_info_dict.get("project_id")
+            else:
```
tombaeyens, line 73: This error message should be in a separate if.

```
if not isinstance(self.project_id, str):
    log error message
```

tombaeyens (approved):

PR #1268, opened 2022-04-08
1175 refactor data source methods

m1n0:


soda/core/soda/execution/data_source.py
```diff
@@ -238,49 +238,71 @@ def sql_find_table_names(
...
+        all_tables = self.get_table_names(include_tables=include_tables, exclude_tables=exclude_tables)
         result = {}
 
         for table in all_tables:
             query = Query(
-                data_source_scan=self,
-                unqualified_query_name=f"get_row_count_{table}",
-                sql=self.sql_get_table_count(table),
+                data_source_scan=self.data_source_scan,
+                unqualified_query_name=f"{query_name}_{table}" or f"get_row_count_{table}",
```
tombaeyens, line 274: Didn't see this constructor string `or` string before
Wouldn't this always take the first?  Even if `query_name` or `table` are None ?

m1n0: ah this slipped, you are totally right, originally I used `{query_name} or f"get_row_count_{table}",` which would of course work, but then I added the _table part which broke it, thanks, fixed!

vijaykiran (approved):

PR #1291, opened 2022-04-22
Handle missing data source.

### LINKED ISSUE #1169: Bad error message for non existing data source
```
soda scan -d does_not_exist -c configuration.yml checks.yml
```

results in 

```
Soda Core 3.0.0b4
Unknown error while executing scan.
  | dictionary changed size during iteration
  | Stacktrace:
  | Traceback (most recent call last):
  |   File "/Users/m1n0/dev/soda/soda-sql/soda-core/soda/core/soda/scan.py", line 267, in execute
  |     for data_source_scan_cfg in self._sodacl_cfg.data_source_scan_cfgs.values():
  | RuntimeError: dictionary changed size during iteration
```

m1n0:
Introduces a global `verbose` variable in the scan file so that we can use verbosity level anywhere in the codebase.

Also changes the global behaviour of logging errors with exceptions, prints the stacktrace only in verbose mode (globally).

Fix #1169

### DIFF
--- a/soda/core/soda/common/log.py
+++ b/soda/core/soda/common/log.py
@@ -63,10 +63,13 @@ class Log:
             exception_str = indent(text=exception_str, prefix="  | ")
             logger.log(python_log_level, exception_str)
             # Logging the stack trace
-            stacktrace = get_exception_stacktrace(self.exception)
-            indented_stacktrace = indent(text=stacktrace, prefix="  | ")
-            logger.log(python_log_level, f"  | Stacktrace:")
-            logger.log(python_log_level, indented_stacktrace)
+            from soda.scan import verbose
+
+            if verbose:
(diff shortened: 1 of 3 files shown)


vijaykiran:
@m1n0  is the verbose option linked to cli as well?

m1n0:
> @m1n0 is the verbose option linked to cli as well?

yes the change is that the verbose option is not only set as an attribute of scan, but also exposed as global variable. Now we can use both to get the verbosity level, and they are set in the same place so unless we change either of them somewhere else they are in sync

vijaykiran (approved):

m1n0:
@vijaykiran One more note - unfortunately I had to use exception to kill the scan execution if given data source does not exists - this is because there is no way to jump out from scan execution at the point of parsing the configs.

Issue #1296, opened 2022-04-22
Review check identity


tombaeyens:
Check identity is a short hash string produced by Soda Core, which is used by Soda Cloud to determine if a new check result should be linked to an already existing check on Soda Cloud.

The identity should hash all the details of a check.  Initially, any change to a check will change the identity and therefor result in a new check being created on Soda Cloud.  Later check linking may improve and link new checks to older check versions.

In order to see if a check is the same, Soda Cloud will use
* scan reference (unique within the organisation)
* file path of the check
* check identity (hash that includes metric, metric configuration, threshold and all check configurations)

Initially all three of these have to be equal for a check to match.  That implies the following statements:

In the following situations, a new check is created, the old one is (soft) deleted and the history is lost:
* When a user changes the threshold of a metric check
* When a user cuts the check from 1 file and pasts it into another file
* When the user adds or removes a check configuration like a validity format
* When the user changes a single space in the custom SQL expression or query of a check

In the following situations, the check is considered the same:
* When the whole check file is unchanged
* When a new check is added or deleted before the check and hence the line number of the check changes
* When a check is moved around in the file, eg by cutting it and pasting it in another location in the same file

PR #1302, opened 2022-04-26
Switched check identity to hash of definition

tombaeyens:
PR for issue #1296

In `ConsistentHashBuilder` inside soda/core/soda/execution/identity.py, 
```
        elif (
            isinstance(value, Location)
            or isinstance(value, MissingAndValidCfg)
            or isinstance(value, ChangeOverTimeCfg)
            or isinstance(value, ThresholdCfg)
            or isinstance(value, SchemaValidations)
        ):
            self.add_all(value.get_identity_parts())
```
still should be removed / cleaned up.  I think from a first look, it is not being needed by the metric identity creation, but it should be double checked.


m1n0:
I like the change, it simplifies things and removes flaky logic of what is included in identity for different check types. We also have a [test](https://github.com/sodadata/soda-core/blob/47a81f7baaa5ddf119c2de7d107e4cfb2bdda204/soda/core/tests/data_source/test_model_identity.py) that tests whether similarly looking checks will pass (which previously caused issues), but may I suggest that we should add one test for this change? 

The tests should tests that line changes are not affecting the check identity by creating a check file, adding one check into it, then calculate identity (e.g. identity is X), then add another check into the file on a line before the original (so that the original check with identity X is now on a different line) and calculating the identities again, making sure that the original check still has identity X.

This scenario was reported as a bug earlier when moving a check to a different line changed its identity and confused Cloud.

PR #1303, opened 2022-04-27
Inform user if no checks found.

### LINKED ISSUE #1233: Scan without checks
```yaml
soda/demos/soda-core-test via 🐍 v3.9.4 (soda-core)
➜ cat empty.yml
soda/demos/soda-core-test via 🐍 v3.9.4 (soda-core)
➜ soda scan -d adventureworks -c configuration.yml empty.yml
Soda Core 0.0.1
Scan summary:
All is good. No failures. No warnings. No errors.
```

Seems to be okay with empty file with no checks. 

here's the output for no checks but  with a table in the yaml:
```
➜ cat empty.yml
checks for dim_account:

soda/demos/soda-core-test via 🐍 v3.9.4 (soda-core)
➜ soda scan -d adventureworks -c configuration.yml empty.yml
Soda Core 0.0.1
Skipping section "checks for dim_account" because content is not a list
  +-> line=1,col=1 in empty.yml
Scan summary:
1 errors.
Oops! 1 error. 0 failures. 0 warnings. 0 pass.
ERRORS:
Skipping section "checks for dim_account" because content is not a list
  +-> line=1,col=1 in empty.yml
```

m1n0:
Also removes what seems to be forgotten line, result of the method is not used and there is no side effect in place.

Fix #1233


soda/core/soda/scan.py
```diff
@@ -370,14 +370,15 @@ def execute(self):
             fail_text = "failure" if checks_warn_count == 1 else "failures"
             error_count = len(self.get_error_logs())
             error_text = "error" if error_count == 1 else "errors"
-            self.__log_checks(None)
```
tombaeyens, line 373: This line logs all the checks that are `NOT EVALUATED`. This may happen when there are problems computing the metrics it depends on.  

I think it's important that the summary explicitely logs which checks have not been evaluated.  Is that done in another way?  Or in other words, why do you think this line can be removed?

soda/core/soda/scan.py
```diff
@@ -370,14 +370,15 @@ def execute(self):
...
             checks_not_evaluated = len(self._checks) - checks_pass_count - checks_warn_count - checks_fail_count
 
+            if len(self._checks) == 0:
+                self._logs.warning("No checks found, 0 checks evaluated.")
```
tombaeyens, line 376: The remainder check logging should be put in an else branch of this `if len(self._checks) == 0:` because no other summary needs to be printed if there are no checks.

tombaeyens (changes requested):

soda/core/soda/scan.py
m1n0: I misunderstood what the method did and did not realize it actually produced output to logs in such case, I thought it was a forgotten line. Thanks.

soda/core/soda/scan.py
m1n0: indeed, we can save a bit of compute here. Changed, I left the cloud communication intact, so that we send empty scans as well, this might be useful info on user activity etc, would you agree?

PR #1304, opened 2022-04-27
Fail freshness if variable is missing.

### LINKED ISSUE #1200: Missing variable in freshness check should throw an error or at least warning
running this without scan_execution_date being available as a variable

```
checks for fact_internet_sales:
  - freshness using ship_date with scan_execution_date < 12h45
```

it uses current scan run time instead of complaining about missing variable

m1n0:
Fix #1200


tombaeyens (changes requested):
Can you add a test for this case?  I think that is a justified use case to protect.

m1n0:
@tombaeyens good point, test added, I also added a way to allow warnings and failures to happen in test scans with a flag, this way we dont have to invoke that logic as originally intended, but also have the flexibility to accept errors/warnings.

vijaykiran (approved):

Issue #1258, opened 2022-04-05
Implement profiling/EDA metrics


bastienboutonnet:
soda-sql offered the following profiling metrics to be pushed into the EDA views of Soda Cloud:
- [ ] count
  - [x] numeric columns:
    - [x] min
    - [x] max
    - [x] avg
    - [x] stddev
    - [x] variance
    - [x] frequent values
    - [x] histograms
    - [x] sum
    - [x] min 5 values
    - [x] max 5 values
    - [x] distinct values
    - [x] missing values
  - [ ] text columns:
    - [x] frequent_values
    - [ ] min length
    - [ ] max length
    - [ ] avg length 

These metrics all need to be re-implemented in soda-core this means:
- fit into the new metrics objects
- be able to generate the right queries for each supported dialect

bastienboutonnet:
More info and links to cloud DTOs can also be found in this closed issue: https://github.com/sodadata/soda-sql-v3/issues/250

bastienboutonnet:
In soda cloud, I see that text columns have the "Extreme Values" tab enabled. However, for text columns, I don't really see how that makes sense on text columns.

@tombaeyens @dirkgroenen do you have an idea why that is?

tombaeyens:
@bastienboutonnet I think this is a duplicate of https://github.com/sodadata/soda-core/issues/1256

Can you close this one if you agree or make a comment if you think otherwise?

bastienboutonnet:
Implemented

Issue #1256, opened 2022-04-05
SodaCL column profiling


bastienboutonnet:
Generating and pushing profiling/EDA metrics to cloud will be done in soda-core via a profiling run.

The original proposition looks like this:

```yaml
profiling advanced:
  tables:
    - SODATEST_%
    - include SODATEST_%
    - exclude SODATEST_%
  columns:
    - TABLE%.COLNAME%
    - include TABLE%.COLNAME%
    - exclude TABLE%.COLNAME%
  max sample size: 100
  statistics: enabled  
```

We are however thinking about another top-level name along the lines of (based on a more explicit proposal from @janet-can:
```yaml
profile columns:
```

I assume the rest of the controls are going to stay the same meaning that users should configure column profiling via this canonical sodaCL spec:

```yaml
profile columns:
  tables: (required?)
    - prod_%
    - include prod_%
    - exclude dev_%
  columns: # not required
    - prod_%.some_column
    - prod_%.some_column_prefix%
    - include table.column
    - exclude table.column
  max sample size: 100 # not required and will default to a sensible number
  statistics: enabled
```

@tombaeyens can you confirm this makes sense language-wise? Also, can you explain the intention behind the `statistics` key? What does it control?

tombaeyens:
Agreed.  Blessing given for changing the name to of the top level element to `profile column:`

tombaeyens:
@bastienboutonnet It would be good to do a review on the proposal of the nested configuration parameters.  What are the options that all can be configured?  And how will they look like in SodaCL.  Can you do a proposal?

bastienboutonnet:
Thanks for the questions @tombaeyens 

I can give further thoughts to what you ask. Before I do that I need some clarifications:
- what does "sample activation" mean?
- by "which metrics" I assume we may want to let users specifiy/override the metrics to be derived?
- "probably in groups or tshirt sizes or just on off switch" I don't really know what you mean. Can you rephrase a bit the way it fits in the sentence has me confused.

> should sample destination (Soda Cloud or S3) be configured under this column profiling section as well?

Am I correct in assuming you refer the sample data dataset tab? (screenshot below
<img width="1095" alt="Screenshot 2022-04-06 at 17 27 10" src="https://user-images.githubusercontent.com/4304794/162011172-85aafaad-cdf0-469e-8fb4-dd21a13f675a.png">
)

tombaeyens:
> what does "sample activation" mean?
[...]

PR #1307, opened 2022-04-28
feat: column profiling (postgres, redshift)

### LINKED ISSUE #1260: Profiling executor
Soda CL `discover tables` and `profile columns` will both be handled by the profiling executor which, based on a `profiling_cfg` will call upon the correct code-path and metrics.

The executor will also collect the results of either profiling run and take care of uploading them to cloud by leveraging soda-core cloud interaction technology

bastienboutonnet:
Partially addresses: #1258, #1256.
Closes #1260

**NOTE**: we're intentionally merging this in a feature branch rather than main to keep review of the feature incremental. Once all separate PRs are approved and merged into the base branch here that branch itself will be merged to main.

**Big Query and Snowflake** are tackled in #1314 and #1321, respectively


m1n0:
This looks great! I wont pretend I absorbed the whole profiling logic yet though 😅 

One general remark, but more of an open question - I am a bit concerned that Scan will grow too much and know too much of the domain logic and not be just an orchestrator/runner of everything, I like how the actual run is separated in its own class, but the run methods for profiling and for automated monitoring in scan also contain a bit of logic which IMHO is not ideal. Saying that, I understand that it needs things from scan and data_source_scan so with current state it seems there is no clever way around it.

bastienboutonnet:
Thanks for having a look @m1n0 

Regarding the profiling logic, do you think there is anything I can do to help you understand? Would you prefer someone else to review that part or do we trust it?

For `Scan` becoming too domain specific. I assume you refer to this kind of stuff here: https://github.com/sodadata/soda-core/pull/1307/files#diff-f8f837bc68ee2608ac96d36a4aed01276f8cb946d8cf810ef30247743a5d89c4R404-R436

If so, I also indeed feel like there is a little bit of domain logic in there. I think there should be a way to build a more factory-like thing but I don't really have a perfect and clear idea in mind right now. So I was thinking maybe to get it in as it is now, and make sure it works so we can start allowing some end to end testing and we can take some time to look into whether we can restructure things a bit. 

Do you think involving someone else in the core team would help us in achieving this maybe or should we say let's just try to go back to it a bit later?

bastienboutonnet:
CI should run again once https://github.com/sodadata/soda-core/pull/1316 is merged and this branch is rebased

vijaykiran:
@bastienboutonnet  can you check the tests? The other PR you mentioned is merged.

soda/core/tests/helpers/common_test_tables.py
```diff
@@ -108,3 +108,35 @@
     columns=customers_test_table.columns,
     values=customers_test_table.values,
 )
+
+customers_profiling = TestTable(
```
tombaeyens, line 112: Why this variation?  Why can't the existing test table not be used for this?

tombaeyens:
@vijaykiran Can you review the (dev-)requirements.txt updates?   Are those in line with the package structure we have in place?

soda/core/tests/helpers/common_test_tables.py
bastienboutonnet: it presented some annoying ties between some records and lead to differences in how ordering would be resolved between local and remote dbs although they are the same as well as some painful arbitrary conversions of float into ints. 

So this variation prevent those from happening but is just valid a test as the other ones.

PR #1317, opened 2022-05-04
Test for 'for each' check identity.

### LINKED ISSUE #1160: Identity changes unexpectedly
Given the `checks.yml` configuration
```
checks for bitcoin:
  - row_count between 0 and 1000000
  - min(usd):
      warn: when < 20000
      fail: when < 10000
  - max(usd):
      warn: when > 70000
      fail: when > 100000
```
execution of `soda scan` will create 3 new checks.

After changing the threshold of the first check and leaving the other 2 checks unchanged in `checks.yml`:
```
checks for bitcoin:
  - row_count:
      warn: when between 20 and 100
  - min(usd):
      warn: when < 20000
      fail: when < 10000
  - max(usd):
      warn: when > 70000
      fail: when > 100000
```
all of a sudden `min(usd)` and `max(usd)` get new identity resulting in creation of new checks in `soda-cloud` and you end up with 6 checks instead of 4.

m1n0:
Fix #1160


m1n0:
CI will pass once #1316 is merged and this is rebased.

vijaykiran (approved):

soda/core/tests/data_source/test_check_identity.py
```diff
@@ -141,3 +144,31 @@ def test_explicitely_specified_check_identity(scanner: Scanner):
...
+            checks:
+            - row_count > 0
+          checks for {customers_table_name}:
+            - row_count > 0
+        """,
+    )
+    assert scan_result["checks"][0]["identity"] == "4bb40424"
+    assert f"checks for {customers_table_name}" in scan_result["checks"][0]["definition"]
+
+    assert scan_result["checks"][1]["identity"] == "e8dd2881"
```
vijaykiran, line 170: Are you sure this is going to be consistent across invocations? Perhaps it is just better to check that the identities are different and we don't really need to check the identity value itself.

PR #1320, opened 2022-05-04
Automated monitoring feature skip in soda-core scan SODA-202

baturayo:
resolves [SODA-202](https://sodadata.atlassian.net/browse/SODA-202)


soda/core/tests/data_source/test_automated_monitoring.py
```diff
@@ -17,9 +15,8 @@ def test_automated_monitoring(scanner: Scanner):
               tables:
                 - include SODATEST_%
                 - exclude PROD%
-              find anomalies:
-                row count: true
-                schema: true
         """
     )
+    scan._is_experimental_auto_monitoring = False
```
bastienboutonnet, line 20: did you mean to leave this to `False`? I would assume this would skip the test. Unless that is what you want?

bastienboutonnet (commented):
looks good to me but check my question about the test

soda/core/tests/data_source/test_automated_monitoring.py
bastienboutonnet: And why not use `@pytest.mark.skip(reason="feature not implemented")` instead?

bastienboutonnet: Ah wait sorry, you want to run the test but make sure that it successfully doesn't run the automated monitoring right?

PR #1322, opened 2022-05-05
feat: column profiling

bastienboutonnet:
Resolves: https://sodadata.atlassian.net/browse/SODA-183 and https://sodadata.atlassian.net/browse/SODA-186

## Important Note
This PR merges a version of column profiling that only supports **table inclusion** I would like to have this reviewed and merged while I work on the complication of dealing with exclusion of columns because there it is not something we have in soda-core out of the box and isn't totally straightforward.

That being said, I'm working on it and should have a PR before end of Sprint 9 to merge with support for this. This risk has been recorded here: https://sodadata.atlassian.net/browse/SODA-359

This means the SodaCL construct looks like:

```yaml
profile columns:
  columns:
    - %.%
    - %.price
    - table_a.price
    - table_a.%
    - include table_a.%
```


soda/core/soda/execution/data_source.py
```diff
@@ -242,6 +247,154 @@ def sql_information_schema_identifier(self) -> str:
     def sql_analyze_table(self, table: str) -> str | None:
         return None
 
+    def profiling_sql_values_frequencies_query(self, table_name: str, column_name: str) -> str:
+        return dedent(
+            f"""
+                with freq_values AS (
```
vijaykiran, line 253: Can you change the SQL keywords to uppercase (like in other places in the code)?

soda/core/soda/execution/data_source.py
```diff
@@ -242,6 +247,154 @@ def sql_information_schema_identifier(self) -> str:
...
+        )
+
+    def histogram_sql_and_boundaries(
+        self, table_name: str, column_name: str, min: int | float, max: int | float
+    ) -> tuple[str, list[int | float]]:
+        # TODO: make configurable or derive dynamically based on data quantiles etc.
+        number_of_bins: int = 20
+
+        assert (
+            min < max
```
vijaykiran, line 370: What happens if column contains only one value or min and max are the same?

vijaykiran (commented):
profile stuff could have been in a sub package

soda/core/soda/execution/data_source.py
bastienboutonnet: To be honest, I didn't really question the code as I took it from soda-sql but I see that it could indeed be an issue and I'll add some handling for this.

soda/core/soda/execution/data_source.py
bastienboutonnet: Will do. I imagine that's a convention we want to be consistent about? Because I am not a fan of uppercase in SQL however if it is a convention I'll comply.

bastienboutonnet:
@vijaykiran regarding subpackage comment, would you mind explaining a bit more why that should be? I imagine in this sprint we won't want to change it to a subpackage so that we have some hope to be able to ship this right?

If we want to make it a subpackage now, we'll need to make sure we're raising this as a risk

soda/core/soda/execution/data_source.py
vijaykiran: Since in other places we use capitals already - I think it would be good to keep the same - because these queries end up in the log and can be a bit annoying for the users if the formatting is inconsistent.

soda/core/soda/execution/data_source.py
vijaykiran: Since in other places we use capitals already - I think it would be good to keep the same - because these queries end up in the log and can be a bit annoying for the users if the formatting is inconsistent.

vijaykiran:
> @vijaykiran regarding subpackage comment, would you mind explaining a bit more why that should be? I imagine in this sprint we won't want to change it to a subpackage so that we have some hope to be able to ship this right?
> 
> If we want to make it a subpackage now, we'll need to make sure we're raising this as a risk

@bastienboutonnet  Sorry it wasn't clear from my comment - I typed it in a hurry. 

I meant just moving it into namespace  `soda.scan.profiling` not a new package.

soda/core/soda/execution/data_source.py
vijaykiran: Fair enough :) I haven't checked soda-sql code - since this is much more focused PR it was easy to spot.

bastienboutonnet:
@vijaykiran all of the files? or just the executor? Because right now, the SQL lives in data source so that it can easily be overridden etc. So I'm not sure what your proposal really entails. 

If you think it's not critical now, I would say, let's open a few follow up issues and discuss the approach. If you think it's critical, then if you can explain here a bit more so that I can get it where you want it to be at quicker that would be awesome.

vijaykiran:
Just the files that are `profiling_*` e.g. `soda/execution/profile_columns_result.py` -> `soda/execution/profiling/columns_result.py`

soda/core/soda/execution/data_source.py
bastienboutonnet: @vijaykiran on this one, we're actually passing the min and the max that we will have obtained from aggregates query before so there will always be a value or sorts.

If this assertion fails (and it must be true) then I guess histograms will not run.

I think it may be better to raise a log error rather than throw an assertion correct?

bastienboutonnet:
@vijaykiran I was able to resolve conflicts. I have one more question regarding the min max question you asked about. Let me know what you think and then I can either do nothing or change that accordingly.

PR #1323, opened 2022-05-05
feat: implement automated monitoring executor/runner

baturayo:
Added by @bastienboutonnet: this PR will also resolve: https://sodadata.atlassian.net/browse/SODA-178


baturayo:
Fixes: SODA-370

bastienboutonnet (approved):
As far as I'm concerned the logic and approach seems legit with me. Let's make sure it works out for @vijaykiran as well especially with regards to the changes we've made to patch the identities of metrics.

Btw, did you check whether the issue with identities that we're fixing here also apply to check result identities? Because I think the same theoretical issue would occur. But I'm also quite happy to tackle this in a follow up PR

 Here's a "shipit cat (it's a cat most of the time)"

soda/core/soda/execution/schema_check.py
```diff
@@ -61,15 +59,10 @@ def evaluate(self, metrics: Dict[str, Metric], historic_values: Dict[str, object
...
         self.measured_schema: List[Dict[str, str]] = metrics.get(KEY_SCHEMA_MEASURED).value
 
-        if KEY_SCHEMA_PREVIOUS in historic_values and historic_values.get(KEY_SCHEMA_PREVIOUS).get("measurements").get(
-            "results"
-        ):
-            logging.info("No previous schema checks available, skipping schema check")
-            self.skipped = True
-
         schema_previous_measurement = (
             historic_values.get(KEY_SCHEMA_PREVIOUS).get("measurements").get("results")[0].get("value")
```
baturayo, line 63: @vijaykiran should we point to the first-ever created schema reference or the latest schema reference in order to run schema check. In other words, should I keep 0 or change it to -1 :D

vijaykiran: Good point - I think `-1` is better too.

PR #1340, opened 2022-05-16
1237 samples2

tombaeyens:


soda/core/soda/common/json_helper.py
```diff
@@ -63,6 +65,8 @@ def to_jsonnable(o):
             return str(o)
         if isinstance(o, Enum):
             return o.value
+        if isinstance(o, Undefined):
```
vijaykiran, line 68: What is this checking for? I'm a bit confused about having a class that is named `Undefined`

soda/core/soda/execution/derived_formula.py
```diff
@@ -5,3 +5,12 @@ class DerivedFormula:
     def __init__(self, function: Callable, metric_dependencies: Dict[str, "Metric"]):
         self.function: Callable = function
         self.metric_dependencies: Dict[str, "Metric"] = metric_dependencies
+
+    def collect_failed_rows_sample_refs(self):
+        if isinstance(self.metric_dependencies, dict):
```
vijaykiran, line 10: When would this be _not_ an instance of `dict` ? I mean - why check for object type instead of value?

vijaykiran:
@tombaeyens  added a few comments, can you also check the snowflake failures?

soda/core/soda/common/json_helper.py
tombaeyens: When querying and getting a metric value from the query, it may return `None`.  So afterwards it would not be possible to distinct between metrics for which the query did not run ok vs queries that returned a None value for the metric

Therefor the metric values are initialized with undefined

It's used in the scan execution later
```
                    if metric.value is not undefined:
                        check_metrics[check_metric_name] = metric
                    else:
                        missing_value_metrics.append(metric)
```

soda/core/soda/execution/derived_formula.py
tombaeyens: No particular reason.  I start to use isinstance more since just checking for value has bitten me couple of times.

vijaykiran: ok, thanks!

PR #1349, opened 2022-05-17
Test anomaly detection for numeric metrics

baturayo:


soda/core/soda/scan.py
```diff
@@ -379,7 +379,6 @@ def execute(self) -> int:
                         f"Metrics {missing_metrics_str} were not computed for check {check.check_cfg.source_line}"
                     )
 
-            self.run_automated_monitoring()
```
baturayo, line 382: This is a forgotten line

PR #1351, opened 2022-05-18
Fix: Automated monitoring revert issues SODA-489

baturayo:
Resolves: https://sodadata.atlassian.net/browse/SODA-462 as well


bastienboutonnet (approved):
LGTM!

PR #1361, opened 2022-05-19
refactor: remove code duplication in sodacl_parser

baturayo:


soda/core/soda/sodacl/sodacl_parser.py
```diff
@@ -56,6 +56,19 @@
 ]
 
 
+def handle_header_content_is_not_an_object_error(func):
```
baturayo, line 59: in the future, we can maybe create an error.py file to store the error handlers

bastienboutonnet: What do you think about a name such that's more along the lines of `assert_header_content_is_dict`. It seems to be more close to what the decorator does right?

Another thing with decorators, and again my knowledge is limited and a bit rusty. But I know that when you use the more vanilla approach, what ends up happening is that sometimes the name of the wrapped function (which is the one you actually care about to have in, for example, tracebacks) gets hidden.

However, you can use `functools.wraps` to put all this in order for you. Check this article out maybe: https://realpython.com/primer-on-python-decorators/#who-are-you-really

bastienboutonnet (commented):
To me this refactor seems very much welcome! I added a comment about naming and the building of the decorator for you to check out but other than that I think it looks ok to me.

I would however want to make sure @vijaykiran has a second pair of eyes once you have taken care of my comment.

soda/core/soda/sodacl/sodacl_parser.py
baturayo: done and thanks for the very useful functools.wraps info :)

bastienboutonnet: Awesome, thanks @baturayo. I'll go ahead and approve but make sure @vijaykiran also gives an ok.

bastienboutonnet (approved):
Looks good to me, I would like to make sure it's all good with @vijaykiran as well.

soda/core/soda/sodacl/sodacl_parser.py
```diff
@@ -1184,99 +1199,59 @@ def __parse_table_filter_section(self, antlr_table_filter_header, header_str, he
...
+                    cfg.include_tables.append(include_table_expression)
         else:
             self.logs.error(
-                f'Skipping section "{header_str}" because content is not an object/dict',
-                location=self.location,
+                'Content of "tables" must be a list of include and/or exclude expressions', location=self.location
             )
 
+    @assert_header_content_is_dict
+    def __parse_automated_monitoring_section(self, header_str, header_content):
```
vijaykiran, line 1221: `header_str` is unused - Same in other other places. `__parse_discover_tables_section`, `__parse_profile_columns_section`

soda/core/soda/sodacl/sodacl_parser.py
```diff
@@ -1450,6 +1425,9 @@ def antlr_parse_column_configuration(self, text: str) -> AntlrParser:
     def antlr_parse_threshold(self, text: str) -> AntlrParser:
         return AntlrParser(text, lambda p: p.threshold())
 
+    def get_data_source_scan_cfgs(self):
+        return self.sodacl_cfg._get_or_create_data_source_scan_cfgs(self.data_source_name)
```
vijaykiran, line 1428: Can you also change the function name in SodaCLCfg and remove `_` ? We use it outside the class, so it doesn't make sense there.

vijaykiran (changes requested):

soda/core/soda/sodacl/sodacl_parser.py
baturayo: yes but it's used in `assert_header_content_is_dict`. Do you have a recommendation how I can remove it?

soda/core/soda/sodacl/sodacl_parser.py
baturayo: done

soda/core/soda/sodacl/sodacl_parser.py
vijaykiran: Ah - you mean it is used in decorator - sorry

vijaykiran (approved):

baturayo:
Resolves: [SODA-523](https://sodadata.atlassian.net/browse/SODA-523)

PR #1370, opened 2022-05-23
Refactored to actual table and actual column names

tombaeyens:
1) Renamed the `format...` methods that handles data source specific case rules for table names, column names and type names to `actual...`

This renaming is an improvement imo because "format" has name clash with the soda validity formats and "actual" has a hint towards testing so that fits better imo.  So anything referring actual means the data source specific name taking the case defaults into account.

2) Introduced variables for the actual names in the TestTable data structure that are initialized in the `ensure_test_table` method.  
 * `TestTable.actual_name` the actual table name
 * `TestTable.test_column[x].actual_name` are the actual column names

3) added convenience methods to scanner to access the data_source.actual_*** methods

### DIFF
--- a/soda/athena/soda/data_sources/athena_data_source.py
+++ b/soda/athena/soda/data_sources/athena_data_source.py
@@ -99,7 +99,7 @@ class DataSourceImpl(DataSource):
         return "table_schema"
 
     @staticmethod
-    def format_column_default(identifier: str) -> str:
+    def actual_column_name(identifier: str) -> str:
         return identifier.lower()
 
     def get_metric_sql_aggregation_expression(self, metric_name: str, metric_args: list[object] | None, expr: str):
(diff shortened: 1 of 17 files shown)


m1n0:
I have one objection for the renaming - these formatting changes are not actual for these data sources, they are just the default ones. Meaning:
if you do 
`create table TEST ...` in postgres, it will save it as `test`, thats why we need these methods.
But if you do `create table "TEST" ...` (quoted table name) it will save, but also return the table name as `TEST` - so formatting method does not come into play. Thats why I named them `format_X_default`.

I understand the naming clash and agree that a better name would make sense, but I am not sure whether `actual_X` describes the nature of these better.

tombaeyens:
@m1n0 I removed the `actual_name` memberfields from the TestTable and TestColumn.  Turned out that was only rarely used and it did not justify a separate mechanism.  So it all boils down to delegating to the datasource.  I guess like you started it.  So net change is pretty minimal :)

m1n0 (approved):

PR #1372, opened 2022-05-23
feat: add ability to send dataset samples to soda cloud (SODA-284) 

baturayo:
Resolves: https://sodadata.atlassian.net/browse/SODA-491


soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -153,6 +153,12 @@ def sql_get_table_names_with_count(
         where_clause = f"\nWHERE {table_filter_expression} \n" if table_filter_expression else ""
         return f"SELECT table_id, row_count \n" f"FROM {self.dataset_name}.__TABLES__" f"{where_clause}"
 
+    def sql_get_table(self, table_name: str, limit: int = None) -> str:
```
tombaeyens, line 156: @baturayo I think it's more appropriate to name it `sql_get_table_sample`

soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -153,6 +153,12 @@ def sql_get_table_names_with_count(
...
+        sql = f"SELECT * \n" f"FROM {table_name}"
```
tombaeyens, line 157: Either concatenate the 2 strings or do something like
```
        sql = (f"SELECT * \n" 
               f"FROM {table_name.lower()}")
```

soda/core/soda/execution/data_source.py
```diff
@@ -165,6 +165,12 @@ def column_metadata_catalog_column() -> str:
     # SQL Queries
     ######################
 
+    def sql_get_table(self, table_name: str, limit: int = None) -> str:
+        sql = f"SELECT * \n" f"FROM {table_name.lower()}"
```
tombaeyens, line 169: same comments as in the bigquery data source

soda/core/soda/execution/sample_tables_run.py
```diff
@@ -0,0 +1,56 @@
...
+from soda.sodacl.discover_tables_cfg import DiscoverTablesCfg
+
+if TYPE_CHECKING:
+    from soda.execution.data_source_scan import DataSourceScan
+
+
+QUERY_PREFIX = "sample_tables: "
+
+
+class SampleTablesRun:
```
tombaeyens, line 16: nice!

soda/core/soda/execution/sample_tables_run.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, data_source_scan: DataSourceScan, discover_tables_cfg: DiscoverTablesCfg):
+
+        self.data_source_scan = data_source_scan
+        self.soda_cloud = data_source_scan.scan._configuration.soda_cloud
+        self.data_source = data_source_scan.data_source
+        self.data_source_name = self.data_source.data_source_name
+        self.discover_tables_cfg: DiscoverTablesCfg = discover_tables_cfg
```
tombaeyens, line 23: cool to see this discover tables cfg abstraction 👍

soda/core/soda/execution/sample_tables_run.py
```diff
@@ -0,0 +1,56 @@
...
+            sample_table_query = Query(
+                data_source_scan=self.data_source_scan,
+                unqualified_query_name=f"{QUERY_PREFIX} get samples from table: {table_name}",
+                sql=sample_table_sql,
+                sample_name="table_sample",
+            )
+            sample_table_query.store()
+            sample_ref = sample_table_query.sample_ref
+
+            _ = sample_tables_result.create_table(table_name, self.data_source_name, sample_ref)
```
tombaeyens, line 50: What's this underscore thingy?  Seems like a cool python trick that i don't know about

tombaeyens (changes requested):

tombaeyens:
Great work, @baturayo !  Just 2 minor change requests.

soda/bigquery/soda/data_sources/bigquery_data_source.py
baturayo: Indeed makes sense👍

soda/core/soda/execution/sample_tables_run.py
baturayo: Actually, I created `SampleTablesCfg` but forget to change it :D SampleTablesCfg is exactly same with `DiscoverTablesCfg`. Maybe we can rename `DiscoverTablesCfg` to `TablesCfg` and use it in both discover table and sample table. Wdyt @bastienboutonnet @tombaeyens ?

bastienboutonnet: No objection with this idea, so go ahead and ping me to review again on this.

soda/core/soda/execution/sample_tables_run.py
bastienboutonnet: `_` is conventionally used in the community when you use a function that has a return that you do not care about. It makes it clear in the code that you do not intent to use it and makes the namespace less busy. But no magic hehe.

soda/bigquery/soda/data_sources/bigquery_data_source.py
bastienboutonnet: This might sound wild (🤯 ) but why not go for `sql_select_star_with_limit`?

bastienboutonnet, line 156: I believe if you use a type checker in your IDE (which you should btw even if we don't enforce it on this repo) that the hint should have warned you. `None` cannot be assigned to an `int`. So if we expect `limit` to sometimes be `None` then I would type this as
`int | None` or Optional[None]

soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -153,6 +153,12 @@ def sql_get_table_names_with_count(
...
+        if limit is not None:
+            sql += f" \n LIMIT {limit}"
```
bastienboutonnet, line 159: I have an allergy for string concatenation like this as it makes the code harder to maintain.

I would advise to have the template at the end, and have your ifs either generate an empty str or not before with explicit names for each part. This means, someone who would later add anything to the code doesn't need to make sure they place their code in exactly the right order.

In this instance, I would:
- put the SQL string at the end as an `f` string with all placeholders in the right order
- store each part of the template before with `''` when they don't get executed.

Hope this is clear if not, I can just show you a concrete example of what I mean.

soda/core/soda/execution/data_source.py
bastienboutonnet: and also here the same comment about templating as I made earlier would apply.

soda/core/soda/soda_cloud/soda_cloud.py
```diff
@@ -49,6 +49,10 @@ def build_scan_results(scan) -> dict:
             if check.outcome is not None and check.archetype is not None
         ]
 
+        profiling = [
```
bastienboutonnet, line 52: Is that really the root object in which the backend expects us to push samples into?

bastienboutonnet (commented):
Nice one! I have one question and a few stylistic/maintainability recommendations.

Let me know when you want another round of reviews! Thanks a lot for your work on this!

soda/bigquery/soda/data_sources/bigquery_data_source.py
baturayo: interesting but more explanatory. Sample includes multiple different techniques (e.g random sampling, limit, stratified etc.) but limit is very specific. I can change it to `sql_select_star_with_limit`. Wdyt @tombaeyens ?

soda/core/soda/soda_cloud/soda_cloud.py
baturayo: Yes, unfortunately for now we have to put sampling inside profiling for backend expectations.

soda/core/soda/execution/sample_tables_run.py
baturayo: removed `_` and changed `create_table` to `append_table` to make it clear

soda/bigquery/soda/data_sources/bigquery_data_source.py
tombaeyens: `sql_select_star_with_limit` is an option but only a level lower.  At the top level, when entering the data_source, it should be per use-case.  So there should be 1 method that can be overridden for customizing only the samples fetching.

If there are multiple use cases that are based on the same query, those use case methods can delegate to a common `sql_select_star_with_limit` underneith.  But we should ensure that we don't have to refactor the data source class if a single data source impl wants to customize the query for a single use case like getting the samples.  It should be possible in the data source impl only by overriding the appropriate method.

soda/core/soda/execution/sample_tables_run.py
tombaeyens: good idea indeed to merge them.  consider `TableSelectionCfg` or `TableSetCfg` as a more descriptive name.

soda/bigquery/soda/data_sources/bigquery_data_source.py
tombaeyens: good point i ll try to keep that in mind as well

soda/core/soda/soda_cloud/soda_cloud.py
bastienboutonnet: Ok that's actually strange as it was discussed that this would be part of `discover tables` originally. However, this is also not ideal and I think we should see with backend if we can get them to plan something that is more dedicated to receive those files. Here's what I think may happen

In this run, profiling info will not be collected, which means that we will submit nothing about profiling. This needs to be confirmed with backend (I would reach out to @dirkgroenen ), but I would suspect that an empty profiling results set may cause previous metrics to be deleted. If this is not the case, then let's go with this approach as it will work and let's work with backend to clean this up later.

If it causes previously pushed profiling metrics to be deleted, we can't ship this just now IMO until some handling is done on the backend.

soda/core/soda/sodacl/tables_cfg.py
```diff
@@ -3,7 +3,7 @@
 from soda.sodacl.location import Location
 
 
-class DiscoverTablesCfg:
+class TablesCfg:
```
bastienboutonnet, line 6: We need to find a better name for this.

The generic version of this, provided we intend both discover tables and send samples to leverage this class should be something like:

`DatasourceTablesCfg` or something like that

bastienboutonnet (commented):
Let's make sure we confirm the backend behaviour + I have a naming suggestion for you to consider

bastienboutonnet (approved):
This PR is now ready to merge from my point of view.

tombaeyens (approved):
Hey @baturayo I see you didn't rename the `sql_select_star_with_limit` yet.  I have a slightly different preference, but we should not block it longer for that discussion.  Progress is more important atm.  So if you want to keep the name just to ahead and merge it.

bastienboutonnet:
@tombaeyens what are your thoughts so we can at least consider it?

bastienboutonnet:
As per agreement with @vijaykiran I won't be merging this until we delineate code-ownership as Vijay would like to control what gets into each release so as far as I'm concerned PR is good to go with me.

PR #1375, opened 2022-05-24
refactor: inform when columns are skipped in profiling via logs

bastienboutonnet:
Log messages for `profile columns` were a bit too insensitive and potentially alarming.

This code change now distinguishes between columns that are skipped and throws an appropriate log message that will be more informative to the user and less misleading.

Previously users would get something like:
```
No profiling information derived for column employee_key in dim_employee
No profiling information derived for column parent_employee_key in dim_employee
No profiling information derived for column sales_territory_key in dim_employee
No profiling information derived for column pay_frequency in dim_employee
...
```

Now they will get:

```
Column: employee_key in table: dim_employee is skipped from profiling.
Column: parent_employee_key in table: dim_employee is skipped from profiling.
Column: sales_territory_key in table: dim_employee is skipped from profiling.
Column: pay_frequency in table: dim_employee is skipped from profiling.
...
```

### DIFF
--- a/soda/core/soda/execution/profile_columns_run.py
+++ b/soda/core/soda/execution/profile_columns_run.py
@@ -22,8 +22,12 @@ class ProfileColumnsRun:
         self.logs = self.data_source_scan.scan._logs
 
     def run(self) -> ProfileColumnsResult:
-        profile_columns_result: ProfileColumnsResult = ProfileColumnsResult(self.profile_columns_cfg)
-        self.logs.info(f"Running column profiling for data source: {self.data_source.data_source_name}")
+        profile_columns_result: ProfileColumnsResult = ProfileColumnsResult(
+            self.profile_columns_cfg
+        )
+        self.logs.info(
+            f"Running column profiling for data source: {self.data_source.data_source_name}"
+        )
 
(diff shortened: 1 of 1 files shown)


m1n0:
looks good, is there some more info for the user if they want to dig into why a column is skipped? That could maybe be shown as a debug level wdyt?

tombaeyens:
Let's change those info messages to debug.  

It's only when you execute with verbose enabled that you would want to see this i think.  There is no UI to see the profiling logs i think.  And probably no way to activate the verbose on those scans.  So i don't think it's used.  And hence it's most likely better if we limit the number of logs we generate until that UX/UI flow is fixed.

tombaeyens (changes requested):
as mentioned please update info -> debug

bastienboutonnet:
Good point on moving this to debug. Done

tombaeyens (approved):
Super!  Thanks!

A note for the future: logs will also be sent to Soda Cloud.   So it might be weird to say `Soda Core could not ...`.  We don't have any guidelines on this yet.  It will become part of our Cloud UX as well as our Soda Core usage UX.  But we'll deal with that later.

m1n0 (approved):

PR #1377, opened 2022-05-25
fix: handle `%.%` in profile columns properly and other bugs

bastienboutonnet:
Fixes improper inclusion handling for `columns: "%.%"` and fixes additional bugs uncovered by adding additional test case.

Resolves: https://sodadata.atlassian.net/browse/SODA-629


soda/core/soda/execution/profile_columns_run.py
```diff
@@ -141,41 +144,63 @@ def profile_numeric_column(
...
             aggregates_query.execute()
             if aggregates_query.rows is not None:
-                profile_columns_result_column.average = float(aggregates_query.rows[0][0])
-                profile_columns_result_column.sum = float(aggregates_query.rows[0][1])
-                profile_columns_result_column.variance = float(aggregates_query.rows[0][2])
-                profile_columns_result_column.standard_deviation = float(aggregates_query.rows[0][3])
-                profile_columns_result_column.distinct_values = int(aggregates_query.rows[0][4])
-                profile_columns_result_column.missing_values = int(aggregates_query.rows[0][5])
+                profile_columns_result_column.average = (
+                    float(aggregates_query.rows[0][0]) if aggregates_query.rows[0][0] is not None else None
```
baturayo, line 148: can't we create a helper method like;

```
def cast_float_with_none(float_number):
     if float_number is not None:
          return float(float_number)
     return None

```

bastienboutonnet: good thinking @baturayo I'll take care of that indeed.

soda/core/soda/execution/profile_columns_run.py
```diff
@@ -224,8 +249,12 @@ def profile_text_column(
             )
             text_aggregates_query.execute()
             if text_aggregates_query.rows:
-                profile_columns_result_column.distinct_values = int(text_aggregates_query.rows[0][0])
-                profile_columns_result_column.missing_values = int(text_aggregates_query.rows[0][1])
+                profile_columns_result_column.distinct_values = (
```
baturayo, line 252: same comment for here instead of float maybe we can use cast_int... helper

soda/core/tests/helpers/scanner.py
```diff
@@ -77,11 +77,16 @@ def enable_mock_sampler(self) -> MockSampler:
             self._configuration.sampler = MockSampler()
         return self._configuration.sampler
 
-    def execute(self, allow_error_warning: bool = False):
+    def execute(self, allow_error_warning: bool = False, allow_warnings_only: bool = False):
         super().execute()
 
         if not allow_error_warning:
-            self.assert_no_error_nor_warning_logs()
+            if not allow_warnings_only:
```
baturayo, line 84: Love it <3

soda/core/soda/execution/profile_columns_run.py
bastienboutonnet: yeah I'll make an even more generic one!

soda/snowflake/soda/data_sources/snowflake_data_source.py
```diff
@@ -119,8 +119,14 @@ def sql_get_table_names_with_count(
         table_filter_expression = self.sql_table_include_exclude_filter(
             "table_name", "row_count", include_tables, exclude_tables
         )
-        where_clause = f"\nWHERE {table_filter_expression} \n" if table_filter_expression else ""
-        return f"SELECT table_name, row_count \n" f"FROM information_schema.tables" f"{where_clause}"
+        where_clause = f"AND {table_filter_expression}" if table_filter_expression else ""
+        sql = f"""
+            SELECT table_name, row_count
+            FROM information_schema.tables
+            WHERE table_schema != 'INFORMATION_SCHEMA'
```
m1n0, line 126: cannot this be the other way around, get metadata from only the desired db and schema? if so, then doing
```
self.schema = connection_properties.get("schema")
self.database = connection_properties.get("database")
```
in the `connect` method should do the trick, schema is picked up automatically in the `sql_table_include_exclude_filter` method and where clause for the schema is added.

If you are ok with this I can take care of it.

m1n0:
pls rebase after #1374 is merged, it should fix some of the athena issues. I will also take a look into the athena tests after that myself, the new test fails because of the same issue as you fixed here for snowflake

bastienboutonnet:
@m1n0 I rebased and it looks like the athena tests still fail. Could you take a look? I have no idea what the assumptions with this db are.

Regarding your comment about the `schema` and `database` variables. I indeed thought this was going to be the case that it was handled. I checked the `sql_table_include_exclude_filter` which seems to indeed be leveraging the `schema` attribute, so my feeling is that it does not work properly.

I'd welcome you trying to make it work. If so, please open a branch from this one as I have a tendency to force push when there are upstream changes that I'm not aware of.

baturayo (approved):

PR #1379, opened 2022-05-25
feat: cap distribution check to 1M rows by default

tituskx:
Adds default limit of 1M rows per sample used in the distribution check. This prevents the use of too much local memory and also puts a limit to the sensitivity of the tests. Look [here](https://github.com/sodadata/docs/commit/e2db6b48a18a1579458ff590e10166efd4a421d7) and [here](https://sodadata.atlassian.net/browse/SODA-578) for more information. 

Resolves: Resolves: https://sodadata.atlassian.net/browse/SODA-578


bastienboutonnet:
Resolves: https://sodadata.atlassian.net/browse/SODA-578

bastienboutonnet:
@m1n0 athena fails here but not sure what we should be expecting, could you take a look at let @tituskx know how it could be fixed?

bastienboutonnet (approved):
As far as the logic of the change is concerned this PR gets my approval.

However it suffers from some issues with Athena tests for which I think we need to have @m1n0 guide us on as I have not got sufficient knowledge of what to expect here.

m1n0 (approved):

bastienboutonnet:
I take the liberty of merging this as it touches solely files that the IA team is responsible for.

Issue #1373, opened 2022-05-23
Issue with failed rows for checks like duplicates


albinkjellin:
This is causing the sync with Soda Cloud.

albinkjellin:
I was able to replicate this issue using the reference data check:
``` yaml
checks for retail_customers:
  - values in country_code must exist in ref_countries iso:
    name: Ensure valid country codes
```
Gives this output:
```
(soda-cl-v3.0.0b15) $soda scan -d aws_postgres_retail uploaderror.yml
Soda Core 3.0.0b15
Empty file upload detected, not sending Content-Length header
No fileId received in response: {'code': 'invalid_empty_upload', 'message': 'File uploads may not be empty'}
Soda cloud error: Could not upload sample failed_rows
  | 'fileId'
Scan summary:
1/1 check PASSED: 
    retail_customers in aws_postgres_retail
      values in country_code must exist in ref_countries iso [PASSED]
1 errors.
Oops! 1 error. 0 failures. 0 warnings. 1 pass.
ERRORS:
Soda cloud error: Could not upload sample failed_rows
  | 'fileId'
Sending results to Soda Cloud
```

albinkjellin:
Another example:
``` yaml
checks for dim_gift:
   - row_count = 551573989
   - duplicate_count(transaction_id) = 0
```

Results in this:
```
soda scan -V -d dwh_2020 -c C:\Users\dasher\.soda\configuration.yml C:\Users\dasher\soda_bigquery\checks.yml
Soda Core 3.0.0b15
Reading configuration file "C:\Users\dasher\.soda\configuration.yml"
Reading SodaCL file "C:\Users\dasher\soda_bigquery\checks.yml"
Scan execution starts
C:\Users\dasher\.venv\lib\site-packages\google\cloud\bigquery\client.py:535: UserWarning: Cannot create BigQuery Storage client, the dependency google-cloud-bigquery-storage is not installed.
  warnings.warn(
Query dwh_2020.dim_gift.aggregation[0]:
SELECT
  COUNT(*)
FROM dim_gift
Invalid type NoneType for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
Invalid type NoneType for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
Invalid type NoneType for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
[...]

PR #1381, opened 2022-05-25
reduce log level

tombaeyens:
Fixing the open trace logging #1373

It should remove the messages

```
Invalid type NoneType for attribute value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

but I don't know how to test it.


m1n0 (changes requested):
I dont think we should silence these messages, it means that invalid values are being set to telemetry attributes. We should rather investigate that.

tombaeyens:
Actually I think we should do both: 

1) Now silence the error message (so merge this PR)
2) Keep an issue open to remind us that we need to resolve the root cause.


@m1n0 if ok, please merge

PR #1387, opened 2022-05-27
fix: typo in data source package import

bastienboutonnet:


soda/core/soda/execution/data_source.py
```diff
@@ -85,7 +85,7 @@ def create(
                 logs.error(f'Data source type "{connection_type}" not found. Did you mean postgres?')
             else:
                 raise DataSourceError(
-                    f'Data source type "{connection_type}" not found. Did you spell {connection_type} correct? Did you install module soda-core-{connection_type}?'
+                    f'Data source type "{connection_type}" not found. Did you spell {connection_type} correctly? Did you install module soda-core-{connection_type}?'
```
vijaykiran, line 88: I might be completely off here, I'm clearly not a native English speaker :P 

In `did you spell x correct?` - correct is an adjective, so it looks okay to me. If we use `correctly` (adverb) then for me `Did you correctly spell x?` 

Paging @janet-can  for authoritative answer :)

bastienboutonnet: Will definitely defer to our native speaker.

janet-can: This. is the best conversation. I have ever had in GitHub.

Vijay is right. "correctly" is the right way to go as it is indeed an adverb. 
You can use "Did you correctly spell x?"  or "Did you spell x correctly?" , though the latter sounds a bit more natural.

PR #1389, opened 2022-05-30
refactor: clean up logging for anomaly detection

bastienboutonnet:


m1n0 (approved):
the SuppressStdoutStderr is interesting, but if all works as expected then LGTM

bastienboutonnet:
@m1n0 yeah it's very shit I took it from somewhere where they were also having issues with prophet logging and making it shut up.

PR #1395, opened 2022-06-03
Feat/add new distr checks

tituskx:
@bastienboutonnet Hey this became a lot bigger than initially expected. Besides adding the new DistributionAlgorithms this PR implements [the following](https://sodadata.slack.com/archives/C033ZCE96TY/p1654169953739229). @baturayo  and I had a chat and thought it made sense that I move on and implement this. If you disagree with this approach I would obviously be happy to take a different route. 

This PR makes the following changes/additions:

- Change `soda update` so that it parses a `dtype` key from the DRO. This `dtype` key is necessary to resample from the DRO.
- Update the RefDataCfg so that it contains a `dtype` key and not a `method` key
- Changes the `__parse_metric_check` method so that it parses a `method` key if present and passes this to the `DistributionCheckCfg`
- Changes `DistributionCheckCfg` so that it contains a `method` attribute
- Changed the return value of `DistributionAlgorithm` so that it makes sense for non-statistical tests like wasserstein and psi
- Added wasserstein and psi algorithms
- Refactored the unit tests and added new ones that check whether wasserstein and psi work for both continuous and categoric data, and whether exceptions are thrown if the specified `method` is incompatible with the `dtype`

Resolves: [add wasserstein metric](https://sodadata.atlassian.net/browse/SODA-195) and [add psi](https://sodadata.atlassian.net/browse/SODA-196)

### DIFF
--- a/soda/core/soda/cli/cli.py
+++ b/soda/core/soda/cli/cli.py
@@ -225,16 +225,16 @@ def update(
     if not column_name:
         logging.error(f"Missing key 'column' in distribution reference file {distribution_reference_file}")
 
-    method = distribution_dict.get("method")
-    if not method:
-        logging.error(f"Missing key 'method' in distribution reference file {distribution_reference_file}")
+    dtype = distribution_dict.get("dtype")
+    if not dtype:
+        logging.error(f"Missing key 'dtype' in distribution reference file {distribution_reference_file}")
 
     filter = distribution_dict.get("filter")
     filter_clause = ""
     if filter is not None:
         filter_clause = f"WHERE {filter}"
 
-    if table_name and column_name and method:
+    if table_name and column_name and dtype:
         query = f"SELECT {column_name} FROM {table_name} {filter_clause}"
         logging.info(f"Querying column values to build distribution reference:\n{query}")
 
@@ -249,7 +249,7 @@ def update(
         from soda.scientific.distribution.comparison import RefDataCfg
         from soda.scientific.distribution.generate_dro import DROGenerator
 
-        dro = DROGenerator(RefDataCfg(method=method), column_values).generate()
+        dro = DROGenerator(RefDataCfg(dtype = dtype), column_values).generate()
         distribution_dict["distribution reference"] = dro.dict()
 
         new_file_content = to_yaml_str(distribution_dict)
--- a/soda/core/soda/execution/distribution_check.py
+++ b/soda/core/soda/execution/distribution_check.py
@@ -43,8 +43,8 @@ class DistributionCheck(Check):
         if self.query.exception is None and self.query.rows is not None:
             test_data = [row[0] for row in self.query.rows]
 
-            _, p_value = DistributionChecker(distribution_check_cfg, test_data).run()
-            self.check_value = p_value
+            distr_check_results = DistributionChecker(distribution_check_cfg, test_data).run()
+            self.check_value = distr_check_results['check_value'] 
 
             self.set_outcome_based_on_check_value()
 
--- a/soda/scientific/soda/scientific/distribution/comparison.py
+++ b/soda/scientific/soda/scientific/distribution/comparison.py
@@ -1,12 +1,13 @@
 import abc
 import logging
+
 from typing import Any, List, Tuple
 
 import numpy as np
 import pandas as pd
 import yaml
 from pydantic import FilePath
-from scipy.stats import chisquare, ks_2samp
+from scipy.stats import chisquare, ks_2samp, wasserstein_distance
 from soda.sodacl.distribution_check_cfg import DistributionCheckCfg
 
 from soda.scientific.distribution.generate_dro import DROGenerator
@@ -16,6 +17,7 @@ from soda.scientific.distribution.utils import (
     assert_bidirectional_categorial_values,
     assert_categorical_min_sample_size,
     distribution_is_all_null,
+    generate_ref_data
 )
 
 
@@ -34,6 +36,8 @@ class DistributionRefParsingException(Exception):
 class MissingCategories(Exception):
     """Thrown when a category in the test data is missing from the ref data."""
 
+class DistributionRefIncompatibleException(Exception):
+    """Thrown when the DRO dtype is incompatible with the test that is used."""
 
 class DistributionChecker:
     def __init__(self, distribution_check_cfg: DistributionCheckCfg, data: List[Any]):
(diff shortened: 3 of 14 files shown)


soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -35,50 +36,60 @@ class MissingCategories(Exception):
...
-        algo_mapping = {
-            "categorical": ChiSqAlgorithm,
-            "continuous": KSAlgorithm,
-            "chi_square": ChiSqAlgorithm,
-            "ks": KSAlgorithm,
-        }
-        self.choosen_algo = algo_mapping[self.ref_cfg.method]
+        self.ref_cfg = self._parse_reference_cfg(cfg.reference_file_path)
 
     def run(self) -> Tuple[float, float]:
```
bastienboutonnet, line 60: the return has changed here so we should update the type hints.

bastienboutonnet, line 60: the return has changed here so we should update the type hints.

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -35,50 +36,60 @@ class MissingCategories(Exception):
...
 
     def _parse_reference_cfg(self, ref_file_path: FilePath) -> RefDataCfg:
         with open(str(ref_file_path)) as stream:
             try:
                 parsed_file: dict = yaml.safe_load(stream)
                 ref_data_cfg = {}
-                if "method" in parsed_file:
-                    ref_data_cfg["method"] = parsed_file["method"]
+
+                correct_configs = {"continuous": ["ks", "psi", "swd"], "categorical": ["chi_square", "psi", "swd"]}
```
bastienboutonnet: The Wasserstein metric is also know as Earth Mover's distance https://en.wikipedia.org/wiki/Earth_mover%27s_distance

We might want to allow for `emd` as an option here and when we write docs refer to both as one and the same and say they can be used interchangeably

bastienboutonnet: The Wasserstein metric is also know as Earth Mover's distance https://en.wikipedia.org/wiki/Earth_mover%27s_distance

We might want to allow for `emd` as an option here and when we write docs refer to both as one and the same and say they can be used interchangeably

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -35,50 +36,60 @@ class MissingCategories(Exception):
...
+                if "dtype" not in parsed_file:
+                    raise DistributionRefKeyException(f"Your {ref_file_path} reference yaml file must have `dtype` key")
```
bastienboutonnet: When you rename `dtype` to whatever you chose from my earlier comment. I think we could have a little more helpful of a message here.

Like, why do users need that key or what does that key do and perhaps even where to find help in the documentation.

bastienboutonnet: When you rename `dtype` to whatever you chose from my earlier comment. I think we could have a little more helpful of a message here.

Like, why do users need that key or what does that key do and perhaps even where to find help in the documentation.

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -35,50 +36,60 @@ class MissingCategories(Exception):
...
+                elif self.method in correct_configs[parsed_file["dtype"]]:
+                    ref_data_cfg["dtype"] = parsed_file["dtype"]
                 else:
-                    raise DistributionRefKeyException(
-                        f"Your {ref_file_path} reference yaml file must have `method` key"
+                    raise DistributionRefIncompatibleException(
+                        f"Your DRO dtype '{parsed_file['dtype']}' is incompatible with the method '{self.method}'"
```
bastienboutonnet, line 102: Could we add something that says what it should be for that method since we know it?

bastienboutonnet, line 102: Could we add something that says what it should be for that method since we know it?

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -99,27 +110,22 @@ def _parse_reference_cfg(self, ref_file_path: FilePath) -> RefDataCfg:
...
     @abc.abstractmethod
-    def generate_ref_data(self) -> pd.Series:
+    def evaluate(self) -> dict:
         ...
 
 
-class ChiSqAlgorithm(DistributionAlgorihm):
+class ChiSqAlgorithm(DistributionAlgorithm):
     def __init__(self, cfg: RefDataCfg, test_data: pd.Series, seed: int = 61) -> None:
+        cfg.dtype = "categorical"
```
bastienboutonnet: I don't get why we go and modify the `dtype` of the cfg in here? Isn't it supposed to be something that the user has to provide?

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -99,27 +110,22 @@ def _parse_reference_cfg(self, ref_file_path: FilePath) -> RefDataCfg:
...
         super().__init__(cfg, test_data, seed)
 
-    def evaluate(self) -> Tuple[float, float]:
+    def evaluate(self) -> dict:
```
bastienboutonnet, line 133: dict of?

bastienboutonnet, line 133: dict of?

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -165,28 +171,45 @@ def evaluate(self) -> Tuple[float, float]:
...
 
 
-class KSAlgorithm(DistributionAlgorihm):
+class KSAlgorithm(DistributionAlgorithm):
     def __init__(self, cfg: RefDataCfg, test_data: pd.Series, seed: int = 61) -> None:
+        cfg.dtype = "continuous"
         super().__init__(cfg, test_data, seed)
 
-    def evaluate(self) -> Tuple[float, float]:
+    def evaluate(self) -> dict:
```
bastienboutonnet, line 183: better to have fully specified types. There are a few similar issues in the rest of the code. I won't point repeat them to not make the review too overwhelming but please go and change them so our types are a bit more useful

bastienboutonnet (commented):
Nice work and really nice refactor/cleanup + proposal for better user contract. 

I've already made a few suggestions via zoom that I know you'll implement but here are a few from an initial read of the PR I did before/shortly after we talked because I had some time.

soda/core/soda/cli/cli.py
```diff
@@ -225,16 +225,16 @@ def update(
     if not column_name:
         logging.error(f"Missing key 'column' in distribution reference file {distribution_reference_file}")
 
-    method = distribution_dict.get("method")
-    if not method:
-        logging.error(f"Missing key 'method' in distribution reference file {distribution_reference_file}")
+    dtype = distribution_dict.get("dtype")
```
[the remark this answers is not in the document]
tituskx: Sure that makes sense!

soda/core/soda/execution/distribution_check.py
```diff
@@ -43,8 +43,8 @@ def evaluate(self, metrics: Dict[str, Metric], historic_values: Dict[str, object
         if self.query.exception is None and self.query.rows is not None:
             test_data = [row[0] for row in self.query.rows]
 
-            _, p_value = DistributionChecker(distribution_check_cfg, test_data).run()
-            self.check_value = p_value
+            distr_check_results = DistributionChecker(distribution_check_cfg, test_data).run()
```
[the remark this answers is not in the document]
tituskx: Got it, my bad!

soda/scientific/soda/scientific/distribution/comparison.py
tituskx: Got it. We use the standardized wasserstein distance metric, so for now I have added `semd` as an option (standardized earth mover's distance). Let me know if you would also like to add the unstandardized versions (I think they are a lot less applicable in our case).

soda/scientific/soda/scientific/distribution/comparison.py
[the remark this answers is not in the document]
tituskx: Sorry my bad this is something from a previous approach that shouldn't be there. My initial approach was to create a `dtype` that would be set using the `generate_ref_data` method, depending on whether the data was of categorical or continuous nature. This method would decide whether the data was categorical or continuous and would not require user input and would always be set to categorical for Chi-square and always be set to continuous for KS. However, Baturay told me to drop that approach since it's not easy to determine whether data is continuous or categorical.

soda/core/soda/execution/distribution_check.py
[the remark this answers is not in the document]
bastienboutonnet: Don't forget to change it in all other places

bastienboutonnet (commented):
Just one little naming change and we're good to go.

PR #1396, opened 2022-06-08
feat: add column exclusion to profile columns

bastienboutonnet:


soda/core/soda/execution/data_source.py
```diff
@@ -205,33 +205,73 @@ def sql_select_all(self, table_name: str, limit: int | None = None) -> str:
...
+            database_filter = f" \n  AND lower({self.column_metadata_catalog_column()}) = '{self.database.lower()}'"
+        else:
+            database_filter = ""
+
+        if self.schema:
+            schema_filter = f" \n  AND lower(table_schema) = '{self.schema.lower()}'"
+        else:
+            schema_filter = ""
+
+        order_clause = f"\nORDER BY ORDINAL_POSITION"
```
tombaeyens, line 248: 1) Why no do this inline in the `# compose  query template` below ?

2) What does it do and why is it added to the query?

soda/core/soda/execution/data_source.py
```diff
@@ -205,33 +205,73 @@ def sql_select_all(self, table_name: str, limit: int | None = None) -> str:
...
     def create_table_columns_query(self, partition: Partition, schema_metric: SchemaMetric) -> TableColumnsQuery:
         return TableColumnsQuery(partition, schema_metric)
 
-    def sql_get_table_columns(self, table_name: str) -> str:
+    def sql_get_table_columns(
+        self, table_name: str, included_columns: list[str] | None = None, excluded_columns: list[str] | None = None
+    ) -> str:
+        # build optional filter clauses
+        if self.database:
+            database_filter = f" \n  AND lower({self.column_metadata_catalog_column()}) = '{self.database.lower()}'"
```
tombaeyens, line 239: Elsewhere I take a different approach that I like better:

filter_clauses = []

...append to the filterclauses like eg...
filter_clauses.append(f"lower({self.column_metadata_catalog_column()}) = '{self.database.lower()}'")

then at the end
filter_clauses_sql = "\n    AND ".join(filter_clauses)

then in the query template something like:
```
return (
    f"SELECT ... \n"
    f"FROM ... \n"
    f"WHERE {filter_clauses_sql} \n"
    ...
)
```

That way you also prevent if-then-else in case you only have 1 clause.  And the approach can be consistent.

soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -17,11 +18,11 @@
 class DataSourceImpl(DataSource):
     TYPE = "bigquery"
 
-    SCHEMA_CHECK_TYPES_MAPPING: Dict = {
+    SCHEMA_CHECK_TYPES_MAPPING: dict = {
```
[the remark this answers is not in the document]
m1n0: yeah `dict` and `type1 | type2` are more recent. We have a task for updating the contributing.md, I added a comment about this there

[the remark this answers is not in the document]
bastienboutonnet: Adding `from __future__ import annotations` because I was using the new versions of annotations in some code introduced in that file, lead to one of our formatters (maybe pyupgrade) to then change them all.

soda/core/soda/execution/data_source.py
```diff
@@ -205,33 +205,73 @@ def sql_select_all(self, table_name: str, limit: int | None = None) -> str:
...
+            schema_filter = f" \n  AND lower(table_schema) = '{self.schema.lower()}'"
+        else:
+            schema_filter = ""
+
+        order_clause = f"\nORDER BY ORDINAL_POSITION"
+
+        if included_columns:
+            included_columns_filter = ""
+            for col in included_columns:
+                included_columns_filter += f"\n AND lower(column_name) LIKE '{col}'"
```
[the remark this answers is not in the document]
bastienboutonnet: I'll throw a lowercase for good measure. They are in PG as far as I know but I also think based on the other queries we often would rather do it twice than rely that it happens somewhere else.

soda/core/soda/execution/data_source.py
bastienboutonnet: What I find potentially annoying with the other approaches or the concatenation approaches is that they are not explicit about how things land in the template.

In this approach, I see the template at then end. I know which elements I potentially have and what they all do via explicit names.

I don't really have an issue with having a few if here and there as they actually make things quite explicit. I feel like maintaining a piece of code that must concatenate or inject in the string such as

`sql +=` which was done before much more damaging to maintain since the layout of the code itself has to happen in the right order whereas here the template at the end takes care of that.

Now if you think you don't want several placeholders in the filter clauses (not sure why that would be because I think they document nicely what kind of things we filter on) and want instead a bunch of strings to be appended into a list, we'll actually still need a fair bit of if and else's.

Also, note that the if-else logic for `self.database` and `self.schema` were already there as if else's so it doesn't look like we have strong consistency.

To me this comes down to stylistic preference, and I am not convinced that the alternative is fundamentally better. If there is a reason other than that that could make maintenance harder (which I don't think it does) or potentially create bugs, I'll be happy to change this but at this point, I would say that the unifying query building style should be in an effort that is focused on that and across the board rather than the scope of this PR.

soda/core/soda/execution/data_source.py
bastienboutonnet: What does "this" refer to? You mean why not always add the `order by` statement in the template?

I don't have a problem with that. It just that previously in the old code, it was separate and I tend to try to leave original decisions mostly untouched unless they make things totally complicated or I see a strong problem with it. I'll take care of that now though

soda/core/soda/execution/data_source.py
tombaeyens: agreed it's style preference.  mostly i wanted to share the join-approach in case you would like it better.

soda/core/soda/execution/data_source.py
tombaeyens: Let me rephrase the questions:

1) Why is the `ORDER BY ORDINAL_POSITION` first assigned as a variable and later added in the template as a variable.  The variable only is assigned once and not changed.  So you could just add the text `ORDER BY ORDINAL_POSITION` straight  into the template below without using a variable

2) Explain what `ORDER BY ORDINAL_POSITION` does and why it is included in the query (if you know) as I don't know what it does and want to learn

tombaeyens (approved):

soda/core/soda/execution/data_source.py
bastienboutonnet: In the most recent code I've now made directly part of the string.

As for 2. I have no idea, it was there before, and I didn't question it. However, I propose to create a follow up ticket to investigate its meaning and whether it's required at all so that if it's not we can remove it. Would that be a good idea?

PR #1398, opened 2022-06-09
Another approach for the Docker image for Soda Core

jmarien:
Proposal for another approach for the Docker image for Soda Core.

This image builds without errors and scores significantly better on Snyk and other security scanners (no critical/high findings).

Image uses:

- Ubuntu 22.04 LTS
- deadsnakes PPA for Python 3.9 packages
- number of libs and dependencies
- Prophet requirements from tagged requirements.txt + prerequisite libs as well
- removal of a small number of no longer required packages after pip install

It is just a proposal, so we may choose not to follow this path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,38 @@
-FROM python:3.9-bullseye
+FROM ubuntu:22.04
 
-# Remove after direct git reference to Prophet is removed.
-RUN apt-get update && apt-get -y install git
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=Etc/UTC
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+# install dependencies
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get remove -y python3 && \
+    apt-get install -y --no-install-recommends software-properties-common gnupg2 && \
+    add-apt-repository -y ppa:deadsnakes/ppa && \
+    apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y --no-install-recommends \
+    build-essential curl \
+    libpq-dev \
+    libsasl2-dev \
+    libssl-dev libffi-dev \
+    python3.9 python3.9-dev libpython3.9-dev libpython3.9 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,38 @@
...
+    unixodbc-dev git && \
+    apt-get clean -qq -y && \
+    apt-get autoclean -qq -y && \
+    apt-get autoremove -qq -y && \
+    rm -rf /var/lib/apt/lists/*
+RUN ln -s /usr/bin/python3.9 /usr/bin/python && \
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+    python3.9 get-pip.py
+
+# extra dependencies for soda-scientific
```
m1n0, line 32: shouldnt these be in setup.py of scientific package?

m1n0 (changes requested):

Dockerfile
bastienboutonnet: They may be but for some weird unknown reason sometimes installing them independently leads to better success. Maybe @jonas has a more scientific answer since he did it recently (see what I did there? Scientific? Ok...)

Dockerfile
```diff
@@ -1,7 +1,38 @@
...
+    libsasl2-dev \
+    libssl-dev libffi-dev \
+    python3.9 python3.9-dev python3.9-venv libpython3.9-dev libpython3.9 \
+    python3.9-distutils \
+    unixodbc-dev git && \
+    apt-get clean -qq -y && \
+    apt-get autoclean -qq -y && \
+    apt-get autoremove -qq -y && \
+    rm -rf /var/lib/apt/lists/*
+RUN ln -s /usr/bin/python3.9 /usr/bin/python && \
```
vijaykiran, line 28: Why is this needed? can't we just use `past install python3-pip`

jmarien: There is no python3-pip in the deadsnakes PPA

Dockerfile
jmarien: When they were added through setup.py, it failed during testing, so I had to install them upfront through the Docker image instead.

PR #1401, opened 2022-06-09
feat: log no threshold error during parsing and provide more informative error during check summary 

tituskx:
Hi! This PR adds the following:

- Logs error during parsing if the check is a `DistributionCheck` and no threshold has been provided. If no threshold is provided the check cannot be evaluated.
- Overrides the `Check.get_summary()` method in `DistributionCheck` so that a more informative message can be shown in the check summary (`NO THRESHOLD PROVIDED` is added to the summary). Previously, a distribution check without a threshold would show a `NOT EVALUATED` message, but did not explain why the check was not evaluated. 

Result without this PR
Screenshot 2022-06-09 at 21 13 34

Result with changes from this PR 
Screenshot 2022-06-09 at 21 12 21


bastienboutonnet (approved):
All good!

soda/core/soda/execution/distribution_check.py
```diff
@@ -105,3 +105,11 @@ def sql_column_values_query(self, distribution_check_cfg, limit=1000000) -> str:
...
         )
         return sql
+
+    def get_summary(self) -> str:
+        error_summary = (
+            " NO THRESHOLD PROVIDED ->"
+            if not self.check_cfg.fail_threshold_cfg and not self.check_cfg.warn_threshold_cfg
+            else ""
+        )
+        return super().get_summary() + error_summary
```
bastienboutonnet, line 115:
```suggestion
        summary_message = f"{super().get_summary()}{error_summary}"
        return summary_message
```

PR #1402, opened 2022-06-10
feat: send distribution check results to cloud so that they can be plotted

tituskx:
This PR makes the following updates:

- Changes `DistributionCheck` so that it uses a `Metric`

This PR is resolves the following issue: [show the p-value or distance metric in a threshold plot in soda cloud](https://sodadata.atlassian.net/browse/SODA-714)


bastienboutonnet (approved):
Looks good to me. @vijaykiran I'll let you merge it before you tag

PR #1406, opened 2022-06-13
add build scan results method

data-fool:


soda/core/soda/scan.py
```diff
@@ -59,6 +60,50 @@ def __init__(self):
...
+            if check.outcome is not None and check.archetype is not None
+        ]
+
+        # TODO: [SODA-608] separate profile columns and sample tables by aligning with the backend team
+        profiling = [
+            profile_table.get_cloud_dict()
+            for profile_table in scan._profile_columns_result_tables + scan._sample_tables_result_tables
+        ]
+
+        return JsonHelper.to_jsonnable(  # type: ignore
```
tombaeyens, line 83: Exposing this data is a good idea.  But it should be done in a dedicated data structure.  I see you have recycled the soda cloud data structure, which is not suitable for this use case.

Can you propose a new Python class containing the data structures that you need?  I guess it will mostly be checks with their outcomes, and logs (debug, warn, error)

What other information do you need/want from the scan?

(If you need a faster, temporary workaround, we can see if there is a quick way for you to get this dict out without code changes)

soda/core/soda/scan.py
```diff
@@ -481,7 +526,7 @@ def execute(self) -> int:
             exit_value = 3
             self._logs.error(f"Error occurred while executing scan.", exception=e)
         finally:
-
+            self.scan_results = self.build_scan_results(scan=self)
```
tombaeyens, line 529: set_data_source_name is a method to configure the scan called before the execute.  i don't think the build scan results belongs here.

Let us know what you want to accomplish and we can help you find the most appropriate and fastest way to do it.

tombaeyens (changes requested):

soda/core/soda/scan.py
data-fool: Hey @tombaeyens! Thank you for reviewing it and for providing your feedback. Yes, my goal is to expose and send this data to a PubSub topic through an Airflow PythonOperator and then from the PubSub topic using an additional step, land the data in BigQuery and use it to create dashboards, operational reports for data quality ( to start with, I'm thinking to report on: number of quality checks failed for a given pipeline, trend over time, and ability to drill down into a specific scan/check looking at runtime metadata, actual scan results, logs and any other info that can give useful insights). 

Regarding the data structure, I tried to implement it based on the suggestion received here https://soda-community.slack.com/archives/C038FFU79J5/p1654809292744639?thread_ts=1654798898.271269&cid=C038FFU79J5 (hope my understanding was correct :)

If there is a faster, better workaround for this, I would love to try to implement it and test it out.

data-fool: @tombaeyens Any updates? Any guidance on how can I get this dict out using a faster, temporary workaround? Thanks!

data-fool: @tombaeyens ↗️ :)

tombaeyens: We'll help you with a temporary fix now and schedule the proper solution on the roadmap.  

The short term workaround is for you to use the Soda Cloud dict.  It's not intended for external usage.  We cannot give any stability guarantees on it.  I think it probably is possible to get the cloud dict out without codechanges.  @vijay, can you find out and explain how a user can get the cloud dict from a scan, potentially also using protected or private members that doesn't require any code changes?

The better long term solution is introducing stable scan results data classes.  That's something that we ll put on our roadmap.

PR #1408, opened 2022-06-14
SODA-44 Added Z to timestamps in soda cloud json

tombaeyens:


soda/core/soda/common/json_helper.py
```diff
@@ -56,7 +56,7 @@ def to_jsonnable(o):
         if isinstance(o, Decimal):
             return float(o)
         if isinstance(o, datetime.datetime):
-            return o.isoformat(timespec="seconds")
+            return o.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
vijaykiran, line 59: Are you sure we need the micro seconds? I thought cloud was only accepting until seconds

vijaykiran: I think the better option instead of adding Z is to use :

`o.astimezone(timezone.utc).isoformat(timespec="seconds")`

vijaykiran: The above would return:  `2022-06-14T09:46:17+00:00`

tombaeyens: 👍 i ll change it.

PR #1411, opened 2022-06-14
CORE-57 improved exception handling when creating data source

tombaeyens:


vijaykiran:
@bastienboutonnet  we prefer to log all the exceptions since it provides better UX for the agent based execution, otherwise the process will die and nothing will be reported to agent. We will keep this approach at least until we refactor the cloud interaction to be async and running separately

PR #1416, opened 2022-06-15
feat: profile and discover view tables

baturayo:
Resolves: https://sodadata.atlassian.net/browse/SODA-887


soda/core/soda/execution/discover_tables_run.py
```diff
@@ -23,14 +23,14 @@ def run(self) -> DiscoverTablesResult:
         self.logs.info(f"Running discover datasets for data source: {self.data_source.data_source_name}")
 
         # row_counts is a dict that maps table names to row counts.
-        row_counts_by_table_name: dict[str, int] = self.data_source.get_row_counts_all_tables(
```
bastienboutonnet, line 26: I would check with the core team if it doesn't make sense to update the `get_row_counts_all_tables` method to also handle views or deprecate it if we prefer to use the `get_table_names` option instead as it could cause confusions where people think they can use this but it actually has an issue with views.

Did you have a chat about this with anyone on this or now? If not, I think I'll create a ticket so that we can address it at a later time.

bastienboutonnet (approved):
Looks good to me.

soda/core/soda/execution/discover_tables_run.py
baturayo: Not really actually, wdyt @vijaykiran about the current logic? The current `get_row_counts_all_tables` cannot find view tables and therefore I made some changes in this PR. Do you think it makes more sense to update `get_row_counts_all_tables` method?

PR #1418, opened 2022-06-16
bug: bugfix for historic measurements having none values

baturayo:


bastienboutonnet (approved):
Ok as a hotfix until https://sodadata.atlassian.net/browse/CORE-74 gets tackled as we probably should fix this much higher upstream.

PR #1421, opened 2022-06-16
feat: Raise a MissingBinsAndWeights exception if `soda scan` runs without `distribution_reference` present

tituskx:
This PR does the following:

- Create a new exception called `MissingBinsAndWeights` that is thrown when no bins and weights are present in the DRO at the time that `soda scan` is run.
- Raise exception `MissingBinsAndWeights` exception if the DRO does not have a `distribution_reference` key or `distribution reference` key when `soda scan` is run.

Resolves: [Throw user warning when we generate the DRO from the test data](https://sodadata.atlassian.net/browse/CLOUD-231). For clarity: after a brief discussion we decided that it would be better to raise an exception instead of just warning the user.


bastienboutonnet (commented):
In principle this looks good. But since I saw some refactoring proposals that seem to point to preferring logging errors rather than raising them, let's wait until an answer to the following thread: https://sodadata.slack.com/archives/C02J6Q493PY/p1655455563276299 gets posted.

soda/scientific/tests/distribution_comparison_test.py
```diff
@@ -120,53 +104,6 @@ def test_distribution_checker(method, reference_file_path, test_data, expected_s
     assert check_results["check_value"] == pytest.approx(expected_p, abs=1e-3)
 
 
-@pytest.mark.parametrize(
```
bastienboutonnet, line 123: As mentioned in DMs. We should:
- use the execute with allowing for errors in a separate test that tests for the errors specifically
- and then peek into the cloud result logs for a specific log message using `pop_scan_result()

bastienboutonnet (commented):
As mentioned in DM I think we can get pretty close to what we were doing before with pytest raise with some custom logic.

vijaykiran:
@tituskx can you please resolve conflicts?

bastienboutonnet (approved):
Looks good. @vijaykiran you can go ahead and merge

PR #1422, opened 2022-06-17
feat: refer to DROs by name

tituskx:
This PR changes the following:

- In `cli.py` add the optional argument `-n` that can be used to specify a `distribution_name` in your distribution reference file. 
- In `cli.py` add functionality of using `-n` argument to select a specific `distribution_name`. Also log errors if 
    - The distribution reference file structure appears to be named and no name is given
    - The distribution reference file appears to contain both named and unnamed DROs
    - The `distribution_name` is not found in the DRO file
- In `cli.py` use `round_trip_dump()` that can deal with multiple DRO dictionary structure to correctly update the distribution reference file.
- In `comparison.py` parse the correct DRO from the distribution reference file if the `distribution_name` is present. Throw exceptions if
    - The `distribution_name` is not present in the distribution reference file
    - The distribution reference file appears to contain named DROs, but no `distribution_name` is provided
- In `sodacl_parser.py` check whether `distribution_name` is provided and log error if no `distribution reference file` key was privded
- In `test_distribution_check.py`:
    - Remove the `my_happy_ml_model_distribution` argument from unit test with unnamed DRO in distribution reference file
    - Add a unit test that checks whether the DRO name is parsed correctly when a named DRO is present in the distribution reference file

This PR resolves: [Refer to DROs by name rather than files](https://sodadata.atlassian.net/browse/CLOUD-206)


soda/core/tests/data_source/test_distribution_check.py
```diff
@@ -79,7 +79,7 @@ def test_distribution_sql(scanner: Scanner, mock_file_system, table, expectation
         f"{user_home_dir}/the_distribution_check_file.yml": dedent(
             f"""
                 checks for {table_name}:
-                  - distribution_difference(size, my_happy_ml_model_distribution) >= 0.05:
+                  - distribution_difference(size) >= 0.05:
```
bastienboutonnet, line 82: shouldn't we be testing for a name to be parsed? Since this PR is about allowing DROs to be referred to by name, then I think it would make sense.

bastienboutonnet (commented):
I have a question regarding some of the changes in the test files

soda/core/tests/data_source/test_distribution_check.py
tituskx: Passing the DRO name is still optional (the changes that I made in this PR are backwards-compatible). The reason I removed the DRO name in this test is because the distribution reference file that is used does not have a DRO name (so passing `my_happy_ml_model_distribution` as an argument does not make sense). 

That being said, I think you are right that we should test that the DRO name is parsed correctly. I will create a new test that uses a distribution reference file with a named DRO and add it to this PR.

bastienboutonnet (approved):
Looks good to me

vijaykiran:
@tituskx  this is not targeting main branch - so I'm not merging it - let me if/when it is ready.

PR #1435, opened 2022-06-21
CORE-63 Added relative percentage change over time

tombaeyens:


soda/core/soda/execution/change_over_time_metric_check.py
```diff
@@ -45,6 +45,17 @@ def evaluate(self, metrics: Dict[str, Metric], historic_values: Dict[str, object
             else:
                 self.check_value = float(metric_value) - float(historic_value)
 
+            historic_descriptor = self.historic_descriptors[KEY_HISTORIC_METRIC_AGGREGATE]
+            if isinstance(historic_descriptor, HistoricChangeOverTimeDescriptor):
+                if historic_descriptor.change_over_time_cfg.percent:
+                    if historic_value == 0:
+                        if self.check_value == 0:
```
vijaykiran, line 52: I am not sure what this if/else is doing .. why set it to 0 if it is 0?

vijaykiran (changes requested):

soda/core/soda/execution/change_over_time_metric_check.py
vijaykiran: @tombaeyens  can you check this ?

tombaeyens: It's if the metric value evolves from 0 to 0, the change should be 0 %.  If the metric value changes from 0 to any other number there is an infinite change %, which was decided to make it not evaluated.  (still think there should be a better solution for this.)

tombaeyens:
This PR is related to docs PR https://github.com/sodadata/docs/pull/308

PR #1437, opened 2022-06-22
feat: Support profile columns inclusion/exclusion behaviour for Spark

baturayo:


bastienboutonnet (approved):
LGTM!

PR #1440, opened 2022-06-22
fix: Suppress `prophet`'s `pandas: frame.append` deprecation warning

tituskx:
The change in this PR ensures that the pandas `frame.append` method deprecation warning is suppressed. This warning is caused by the `prophet` package that is used for anomaly detection.

Resolves the following issue: [Check pandas warnings in automated monitoring](https://sodadata.atlassian.net/browse/CLOUD-278)


bastienboutonnet (approved):
looks good

PR #1444, opened 2022-06-23
Change: rename the `update` command to `update-dro` as it better describes what the command is used for

tituskx:
This PR implements the following

* Use `soda update-dro` instead of `soda update` to update a DRO in the distribution reference file

This PR resolves the following: [Rename `soda update` to `soda update-dro`](https://sodadata.atlassian.net/browse/CLOUD-372)


bastienboutonnet (approved):
Good stuff, just in time before we move to stable-er version numbers :p

PR #1445, opened 2022-06-23
test: Improve profiling test for all tables and all columns

baturayo:


soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -114,24 +119,46 @@ def test_profile_columns_text(data_source_fixture: DataSourceFixture):
...
+    assert parsed_excluded_tables_and_columns == {}
+
+    for table_name in table_names:
+        columns_metadata_result = data_source.get_table_columns(
+            table_name=table_name,
+            query_name=f"profile-columns-get-column-metadata-for-{table_name}",
+            included_columns=parsed_included_tables_and_columns,
+            excluded_columns=parsed_excluded_tables_and_columns,
+        )
+        assert columns_metadata_result
```
bastienboutonnet, line 161: Does this just assert that there is at least one column (i.e. that the object is not None, but would we want to/can we have a more expected content to assert maybe?

soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -114,24 +119,46 @@ def test_profile_columns_text(data_source_fixture: DataSourceFixture):
...
+        include_tables=profile_columns_run._get_table_expression(include_columns),
+        exclude_tables=profile_columns_run._get_table_expression(exclude_columns, is_for_exclusion=True),
+        query_name="profile-columns-get-table-names",
+    )
+    assert len(table_names) > 2
+
+    parsed_included_tables_and_columns = profile_columns_run._build_column_expression_list(include_columns)
+    assert parsed_included_tables_and_columns == {"%": ["%"]}
+
+    parsed_excluded_tables_and_columns = profile_columns_run._build_column_expression_list(exclude_columns)
```
bastienboutonnet, line 151: Nice find!

bastienboutonnet (commented):
Nice find, just curious about whether we could have an explanation that's a little bit more specific if we can afford to.

tombaeyens:
For me it's important that 3 things in this test are split up and tested separately and that we have coverage for these 3 separatey:

1) querying the metadata : does the metadata query fetch the right tables.  

2) table name matching: test that include/exclude filters pass properly the various table/column combinations that are given.  This should be ideally be tested without any queries.  Given that 1 is tested separately, this might be done with without any db querying.  It might require internal methods to be called but that could be a tradeoff.

3) column profiling.  this is *very* heavy in querying and here the minimal profiling should be executed to get sufficient coverage.  1 profile for a text column and 1 profile for a numeric column is the max that should be executed.

Those are the inputs I can give at this time.  If that's not sufficient, then just assign to one of us so we will have a look deeper at how to balance those tradeoffs

baturayo:
> 

Indeed column profiling is heavy; therefore, we use internal methods to skip most of the heavy parts. At least we target to test the table and column names when we pass `%.%`. The functionalities of profiling are already tested in separate unit tests. 

The issue is when we test `%.%`. We always get different tables, since we generate the tables in the other unit tests. If those tests that generate tables don't run, then we have no tables to test at DB. Our proposal is that we can create the tables at the beginning of unit tests. We are already ensuring the existence of tables in unit tests. So, the logic of other unit tests won't change, assuming that we don't alter tables. Wdyt @tombaeyens about generating all tables at the beginning of unit tests?

tombaeyens:
@baturayo given that profiling is tested in other tests, you have to ensure that no more profiling is executed in this test.  

You have to write your test logic in such a way that it only assumes on certain tables being present (the max 2 tables that you specify in the test with `ensure_table`).  Your test logic should not assume anything of tables not being present.  It could be that a lot of other tables are present or that no other tables are present.  Your test code should be written such that this does not matter. That is a part of the basic test infrastructure which is very hard to change.

Your test code can still rely that for example a table like `osidufos8s76df` or `non_existing_table` do not exist.

bastienboutonnet:
@baturayo I guess what we can do is expect that the tables we have ensured are in the result and not care too much about the rest. That's kind of the only way out of this that I can think of. It, unfortunately, isn't the best test for `%.%` in my opinion, but if we are not going to be able to make deterministic expectations about what tables may be on the test db at any given time (which btw, I think is a problem --but maybe that's an issue for another day) then I can't see any other option.

@tombaeyens what does that sound like? It is hard from your comments to know what you have in mind concretely so this is my creative license here.

I have to say, though, I wonder what makes it so difficult for us to consider ensuring that all tables used in tests be present at the start of the test run, other than "That is a part of the basic test infrastructure which is very hard to change" which I fail to see as a terribly hard thing. To me, it would seem a series of `ensure_table()` calls that iterate over a list of tables maintained in a central place should actually do the trick.

bastienboutonnet:
pre-commit.ci autofix

PR #1447, opened 2022-06-24
feat/fix: ensure empty bins for integer columns are not created and fix bin width derivation

baturayo:


soda/core/soda/execution/data_source.py
```diff
@@ -488,26 +488,35 @@ def profiling_sql_aggregates_text(self, table_name: str, column_name: str) -> st
...
             self.logs.warning(
-                f"Min of {column_name} on table: {table_name} must be smaller than max value. Min is {min}, and max is {max}"
+                f"Min of {column_name} on table: {table_name} must be smaller than max value. Min is {minimum}, and max is {maximum}"
             )
             return None, []
 
-        min_value = floor(min * 1000) / 1000
-        max_value = ceil(max * 1000) / 1000
-        bin_width = (max_value - min_value) / number_of_bins
+        min_value = floor(minimum * 1000) / 1000
```
baturayo, line 510: @bastienboutonnet why do we need floor and ceil in these lines?

soda/core/soda/execution/data_source.py
```diff
@@ -488,26 +488,41 @@ def profiling_sql_aggregates_text(self, table_name: str, column_name: str) -> st
...
+        else:
+            bin_width = (max_value - min_value) / number_of_bins
 
         boundary_start = min_value
         bins_list = [min_value]
-        for _ in range(0, number_of_bins):
+        for _ in range(0, number_of_intervals):
             boundary_start += bin_width
-            bins_list.append(round(boundary_start, 3))
+            bins_list.append(round(boundary_start, 2))
```
tituskx: Would this be a suitable alternative to the for loop?

```
bins_list = [min_value]
bins_list += \
    [(round(min_value + bin_width * i) for i in range(1, number_of_intervals + 1)]
```

soda/core/soda/execution/data_source.py
```diff
@@ -488,26 +488,41 @@ def profiling_sql_aggregates_text(self, table_name: str, column_name: str) -> st
...
+        min_value = floor(minimum * 1000) / 1000
+        max_value = ceil(maximum * 1000) / 1000
+        if column_type == "integer":
+            min_value = int(min_value)
+            max_value = int(max_value)
+            bin_width = (max_value - min_value) / number_of_intervals
+            if bin_width.is_integer():
+                bin_width = int(bin_width)
+        else:
+            bin_width = (max_value - min_value) / number_of_bins
```
tituskx: Hey Baturay, not sure whether this else statement should be here. Setting the `bin_width = (max_value - min_value) / number_of_bins` seems inconsistent with `for _ in range(0, number_of_intervals)` that is used later. 

If you have float data with `min_value` and `max_value`, `5` and `30`, respectively, then `number_of_bins` is `20`. In that case your `bin_width` would be set to `1.25`. In the for loop your bins would be set to `[5, 6.25, ..., 28.75]` whereas you would want `[5, 6.32, ..., 30]`.

soda/core/soda/execution/data_source.py
bastienboutonnet: I literally have no idea, I stole the code from soda-sql to maintain feature parity and consistency with derivation and didn't take the time to properly evaluate the soundness.

Do you have an issue with it?

bastienboutonnet: I would say if it's unsound, maybe let's open up a separate issue for this and tackle it then, because if we change the way things are derived it may cause unexpected differences and I would prefer that we have an explanation and potentially an annoucement or release note for it.

baturayo: nope just wondering whether we should keep it or not :D I think deleting shouldn't cause any issue since we already round the bins

bastienboutonnet: yeah... seems like so. Please open a ticket so we can prio it and check whether it leads to differences or not and if not, we can remove it because unnecessary code isn't nice to have.

tituskx (commented):
Nice work!

soda/core/soda/execution/data_source.py
baturayo: yep fixed it with a slightly diff approach

soda/core/soda/execution/data_source.py
baturayo: fixed

soda/core/soda/execution/data_source.py
```diff
@@ -488,26 +487,33 @@ def profiling_sql_aggregates_text(self, table_name: str, column_name: str) -> st
...
+            bin_width = int(bin_width)
+            min_value = int(min_value)
+            max_value = int(max_value)
 
-        boundary_start = min_value
-        bins_list = [min_value]
-        for _ in range(0, number_of_bins):
-            boundary_start += bin_width
-            bins_list.append(round(boundary_start, 3))
+        bins_list = [round(min_value + i * bin_width, 2) for i in range(0, number_of_bins)]
```
tituskx, line 516: Tiny difference with previous code regarding functionality because `min_value` did not use to be rounded and it is now

tituskx, line 516: Tiny difference with previous code regarding functionality because `min_value` did not use to be rounded and it is now

tituskx (commented):
Tiny difference now with the rounding of `min_value`

soda/core/soda/execution/data_source.py
tituskx: That does mean that you round `min_value` whereas previously we did not right?

soda/core/soda/execution/data_source.py
baturayo: isn't it same in both cases we sum it first and then round it?

baturayo: hmm indeed good catch, then the first min_value was rounded in the upper lines and now it's rounded in the list.

tituskx: Btw, wouldn't it make more sense to deal with the formatting in the front-end? Is it correct that we're rounding to two decimals because of the way the bins will displayed?

tituskx: Not sure if any of this matters though haha, probably not important for now right?

baturayo: yes, that's true. Personally, to have more control I would prefer to handle it on soda-core. Wdyt @bastienboutonnet ?

bastienboutonnet: Front end most definitely won't apply any domain logic. Cloud only receives data and doesn't perform any other kind of logic on it. So indeed this should be something soda-core is responsible for.

bastienboutonnet (approved):
Nice rework @baturayo Good to go from my side!

tituskx (approved):

PR #1451, opened 2022-06-25
added table data to generation script

tombaeyens:
See `tests/integration/compatibility_script.py`


vijaykiran:
This is no longer required - we go with providing documentation on how to add new  datasource support

PR #1460, opened 2022-06-28
Comment for pinned prophet version

m1n0:


soda/scientific/setup.py
```diff
@@ -21,7 +21,7 @@
     "PyYAML>=5.4.1,<6.0.0",
     "cython>=0.22",
     "pystan==2.19.1.1",
-    "prophet==1.0.0"
+    "prophet==1.0.0"  # Pinned 1.0.0 until 1.1 (and above) installation issues are resolved.
```
bastienboutonnet, line 24: You can even refer to this: https://sodadata.atlassian.net/browse/CLOUD-446

bastienboutonnet (approved):
Thanks for that indeeed! Feel free to add reference to the issue even

PR #1461, opened 2022-06-28
fix the e parameter

jmarien:
Fix for the Docker image for non-dev


Dockerfile-dev
```diff
@@ -0,0 +1,57 @@
+FROM ubuntu:22.04
```
m1n0, line 1: wouldnt it make sense to extend the non-dev one and just reinstall the packages with -e?

PR #1463, opened 2022-06-29
fix: handle exceptions from scientific library and log them instead or letting them raise

tituskx:
This PR does the following: 

- Wraps the initialisation of the `AnomalyMetricCheck` in a try - except block
- Wraps the execution of data source scans in a try - except block
- Sets default `metric.value` and `self.check_value` attribute values in `DistributionCheck`
- Wraps the whole part from computation of `check_result_dict` until and including `self.set_outcome_based_on_check_value()` in a try - except block
- Adds line to `to_jsonnable` that checks whether value `o` instance is `Exception`. If yes returns `str(o)`, so that `Exceptions` raised in `DistributionCheck` can be send to cloud and do not prevent the results from other successful checks being sent to cloud.

This resolves: [Make sure Exceptions everywhere in Sci lib or checks that we own send logging rather than raise to crash](https://sodadata.atlassian.net/browse/CLOUD-357)

@bastienboutonnet As I mentioned during the call we just had I initially thought that the scope was a bit bigger (i.e. ensure that no check owned by the IA team can stop soda-core from executing all its other checks and sending the results to cloud). This is why I had changed the `to_jsonnable` function; to prevent the `Exception` being sent to the API (which will result in a 400 I believe). If you would like me to remove that part and address this in a separate issue I would be happy to.


soda/core/soda/common/json_helper.py
```diff
@@ -68,6 +69,8 @@ def to_jsonnable(o):
             return o.value
         if isinstance(o, Undefined):
             return None
+        if isinstance(o, Exception):
```
bastienboutonnet, line 72: What caused having to implement this?

bastienboutonnet: I'm surprised that this comes up now since I believe soda-core already has a bunch of exceptions as part of their logs no?

bastienboutonnet: For posterity as discussed in DMs: this didn't come up before because all usages of logs either stringify the exception or perform string completion.

I would still be in favour of this handling though since it's a more generic/safe option in case developers forget to stringify things before.

bastienboutonnet (approved):
Looks good to me!

soda/core/soda/common/json_helper.py
vijaykiran: LGTM

vijaykiran (approved):

PR #1469, opened 2022-06-29
fix: use default distribution comparison method when user has not provided one

tituskx:
This PR does the following:

* Ensure that distribution checks run successfully when no `method`  key is provided to the check in the `checks.yml` file. If no `method` key is provided the distribution check defaults to the Kolmogov-Smirnov test or the Chi-square test depending on whether the data is continuous or categorical respectively.

This resolves: [When method is not filled in the distribution check unhandled error is not handled](https://sodadata.atlassian.net/browse/CLOUD-402)


bastienboutonnet (approved):
Good to go!

PR #1479, opened 2022-07-05
Async Docker image building through Actions and dispatch

jmarien:
Docker image building through Github Actions, with a final step in main flow to dispatch an event in case of a version tag.

The triggered workflow will then sleep for 900sec, then build and push the Docker image.

Closes [CLOUD-423]


jmarien:
Tested on a private repo.

Issue #1483, opened 2022-07-06
get_all_check_text uses '/n' instead of '\n' for newlines


bbakersmith:

jmarien:
SODA-953

PR #1484, opened 2022-07-06
Join check text with newlines instead of /n

bbakersmith:
#1483

### DIFF
--- a/soda/core/soda/scan.py
+++ b/soda/core/soda/scan.py
@@ -495,7 +495,7 @@ class Scan:
                     )
 
     def __checks_to_text(self, checks: list[Check]):
-        return "/n".join([str(check) for check in checks])
+        return "\n".join([str(check) for check in checks])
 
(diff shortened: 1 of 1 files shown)


soda/core/soda/scan.py
```diff
@@ -495,7 +495,7 @@ def run_data_source_scan(self):
                     )
 
     def __checks_to_text(self, checks: list[Check]):
-        return "/n".join([str(check) for check in checks])
```
Adamage, line 498: How about `os.linesep` ?

bbakersmith: I don't think so, that note is in reference to its use as a line terminator, not a file terminator. But I'll try to write some tests to see how it behaves.

vijaykiran: I think we can live with this for now :)

vijaykiran (approved):

Issue #1311, opened 2022-05-02
ZeroDivisionError for `*_percentage` checks should result in zero


jasonicarter:
For percentage calculations, if you have zero rows you should just return a zero result. My thinking is...

1. The `ZeroDivisionError` shouldn't error out causing the entire scan to fail
2. If there are zero rows then there is no missing_percentage and therefore the answer should be zero `missing_percentage = 0`

```
Error occurred while executing scan.
  | division by zero
  | Stacktrace:
  | Traceback (most recent call last):
  |   File "/usr/local/lib/python3.9/site-packages/soda/scan.py", line 314, in execute
  |     metric.compute_derived_metric_values()
  |   File "/usr/local/lib/python3.9/site-packages/soda/execution/derived_metric.py", line 123, in compute_derived_metric_values
  |     self.value = self.derived_formula.function(formula_values)
  |   File "/usr/local/lib/python3.9/site-packages/soda/execution/derived_metric.py", line 62, in derive_missing_percentage
  |     missing_percentage = values["missing_count"] * 100 / values["row_count"]
  | ZeroDivisionError: division by zero
  ```

jmarien:
SODA-212

PR #1487, opened 2022-07-06
Treat zero missing or invalid rows as zero percent

bbakersmith:
#1311

### DIFF
--- a/soda/core/soda/execution/derived_metric.py
+++ b/soda/core/soda/execution/derived_metric.py
@@ -11,19 +11,25 @@ DERIVED_METRIC_NAMES = [
 
 
 def derive_missing_percentage(values: Dict[str, int]):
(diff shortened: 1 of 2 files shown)


soda/core/soda/execution/derived_metric.py
```diff
@@ -10,6 +10,24 @@
 ]
 
 
+def derive_missing_percentage(values: Dict[str, int]):
+    if values["missing_count"] == 0:
```
Adamage, line 14: A more Pythonic approach would be:
```python
missing_count = values.get("missing_count")
row_count = values.get("row_count")

if missing_count and row_count:
    missing_percentage = missing_count * 100 / row_count
else:
    missing_percentage = 0
```

vijaykiran:
Thank you @bbakersmith  @Adamage !

PR #1490, opened 2022-07-08
fix: need to send error messages to cloud for non default anomaly detection checks

baturayo:


bastienboutonnet (approved):
Looks good!

PR #1494, opened 2022-07-11
fix: prevent empty table list from running all tables

baturayo:


bastienboutonnet (approved):
Looks good to me!

PR #1498, opened 2022-07-12
fix: better logging messages for profiling and discover datasets

baturayo:


bastienboutonnet (commented):
I think we should be able to indeed go with errors there. Let's see if it messes up experience. 

Just a few formatting suggestions to apply and we're good.

bastienboutonnet (approved):
Let's go!

PR #1500, opened 2022-07-13
Soda 914 extend bq authentication

tooobsias:
With GCP/BigQuery one has the option to use different methods to authenticate.
The standard authentication methods are:

1. Use of Application Default Credentials
2. Use of Application Default Credentials with Service Account impersonation
3. Use of Service Account Key (already implemented)
4. Use of Service Account Key with Service Account Impersonation  

###  Usage in configuration.yml file:
1. Use of Application Default Credentials
```
data_source <SOURCE-NAME>:
  type: bigquery
  connection:
    use_context_auth: True
```
2. Use of Application Default Credentials with Service Account impersonation
```
data_source <SOURCE-NAME>:
  type: bigquery
  connection:
    use_context_auth: True
    impersonation_account: <SA_EMAIL>
```
3. Use of Service Account Key
...
4. Use of Service Account Key with Service Account Impersonation
```
data_source my_database_name:
  type: bigquery
  connection:
    account_info_json: '{
        "type": "service_account",
        "project_id": "...",
        "private_key_id": "...",
      ...}'
    impersonation_account: <SA_EMAIL>
```


soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -80,6 +81,10 @@ def __init__(self, logs: Logs, data_source_name: str, data_source_properties: di
         if account_info_json_str:
             self.account_info_dict = json.loads(account_info_json_str)
 
+        if self.account_info_dict is None:
+            self.logs.info("Using application default credentials.")
+            self.credentials, _ = default()
```
m1n0, line 86: `self.credentials` is set to None a few lines below so this is not really taken into account anyway. I would suggest removing this `if` and doing 
```
if self.data_source_properties.get("use_context_auth") or self.account_info_dict is None:
```
below

soda/bigquery/soda/data_sources/bigquery_data_source.py
```diff
@@ -94,9 +99,24 @@ def __init__(self, logs: Logs, data_source_name: str, data_source_properties: di
...
+                target_principal=str(impersonation_account),
+                target_scopes=self.data_source_properties.get("scopes", default_auth_scopes),
+            )
+
         # All remaining parameters
-        # Usually the project_id comes from the self.account_info_dict
-        self.project_id = self.account_info_dict.get("project_id") if self.account_info_dict else None
+        # Usually the project_id comes from the self.account_info_dict or the credentials itself
+        self.project_id = (
+            self.account_info_dict.get("project_id") if self.account_info_dict else self.credentials.quota_project_id
```
m1n0, line 118: I dont think `self.credentials.quota_project_id` is safe to do without checking if its there, if default credentials is used then project id might be `None`:
(from google lib doc):
```
the current environment's credentials and project ID. Project ID
may be None, which indicates that the Project ID could not be
ascertained from the environment.
```

m1n0 (changes requested):
Thank you for this! I think we just need some minor adjustments but otherwise this should be good to go!

soda/bigquery/soda/data_sources/bigquery_data_source.py
tooobsias: Agree, this is redundant. Will adjust it accordingly

soda/bigquery/soda/data_sources/bigquery_data_source.py
tooobsias: From how I read the Docs, the `self.credentials.quota_project_id` obtained from the [Credentials Module](https://google-auth.readthedocs.io/en/master/reference/google.auth.credentials.html#google.auth.credentials.Credentials.quota_project_id) is always set.   

Its however different from the `project_id` obtained from `google.auth.default`. As you said, this 
```
Returns the current environment's credentials and project ID. Project ID
may be None, which indicates that the Project ID could not be
ascertained from the environment.
```

By default and when using context auth, the quota_project_id parameter will be used.
The user however can optionally overwrite in the connection properties.
See:
`self.project_id = data_source_properties.get("project_id", self.project_id)`

tooobsias: @m1n0 Following suggestions:

1. We could try receive the `project_id` from the `google.auth.default()`
2. If its not there, fall back to using `quota_project_id`
Optionally: User can set `project_id` via connection properties

m1n0: I see, that makes sense, lets make sure that the option to set the project_id via connection properties is there as well!

m1n0 (changes requested):
I am not sure what happened here with the force push but this PR is now huge, could you please rebase or create a merge commit if thats easy to resolve this? (we can squash-merge afterwards so extra commit is not an issue)
Also, I think I commented about a duplicate line of code but I cant seem to find that comment now, hopefully it will re-appear once the branch is updated

tooobsias:
Hi @m1n0,

yes, the force push was actually my fault. Dont know what happend here either but will try to resolve it.

Regarding the duplicate lines:
This has been fixed with commit [SODA-914: remove redundant lines](https://github.com/sodadata/soda-core/pull/1500/commits/fa86795e29b644e5f413440fe5a8626bbd2c9cd0).
Dont know why it does not show up anymore either

m1n0:
hi @tooobsias, ah I see, I overlooked that one. Thank you!

tooobsias:
@m1n0 
Seems like I made it worse (and honestly dont know what I did to close the PR)
I will reopen another Merge Request which addresses your suggestion regarding:
1. Duplicate lines
2. Selection of project_id/quota_project_id

PR #1502, opened 2022-07-13
Check name refactor

m1n0:
This does two things:
1. remove the original `check.name`:
- we have user specified name, we generate another name for cloud, we have metric name, this was not used anywhere anyway and its hard to see a use case where it will be needed. Just for the reference, it was either set to metric name in case of metric checks (e.g. `row_count`) or to a hardcoded string like `"Freshness"` 
- the new `check.name` property uses the same logic as what we sent to cloud previously, i.e. uses the user defined name if provided, generates one if not.
- I am invoking you @bastienboutonnet just for one specific thing - previously we used this `check.name` in anomaly metric checks as a key to get previous values, I just switched it to use the metric name directly since metric checks used that in the name property anyway. Just check that one please.
- @tombaeyens pls give your thoughts on whether you think removing the name is an issue, we had a discussion already on this and it seems non-crucial. We can always re-add a property like `machine_name` or something like that for some sort of more controlled check name if needed internally.

2. Refactor query/check/metric files into submodules for convenience


bastienboutonnet (commented):
@m1n0 Thanks for the heads up. I think it should be ok. Shouldn't impact still being able to get the same metrics right?

vijaykiran:
@m1n0  Can you please check the conflicts? I merged another PR and didn't check for the order

PR #1534, opened 2022-08-11
feat(profiling): add support for extra numeric and text datatypes

bastienboutonnet:
I recently realised that we were only considering the most common data types in column profiling for most dbs.

I've now added support to pretty much all numeric and text types for the following databases:
- postgres (base) & therefore redshift
- snowflake
- bigquery
- db2

I've left out:
- athena because it seems like types for profiling were commented out (not sure why), see: https://github.com/sodadata/soda-core/blob/main/soda/athena/soda/data_sources/athena_data_source.py#L84-L85
- sqlserver (all types seem to be already there from a previous implementation)
- MySQL (all types seem to be already there from a previous implementation)

### DIFF
--- a/soda/bigquery/soda/data_sources/bigquery_data_source.py
+++ b/soda/bigquery/soda/data_sources/bigquery_data_source.py
@@ -46,7 +46,19 @@ class BigQueryDataSource(DataSource):
         DataType.BOOLEAN: "BOOL",
     }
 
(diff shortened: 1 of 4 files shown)


bastienboutonnet:
There seems to be some failures for some data sources, although these seem to be due to issues with the docker containers. Not sure what I can do there other than re-triggering CI? @vijaykiran let me know if there's anything else I can do to help here.

bastienboutonnet:
Looks like triggering the failed jobs again resolved the failing ones.

PR #1554, opened 2022-08-23
Change spark installation to decouple deps

vijaykiran:
When installing soda-core-spark-df, it is currently pulling in all the
dependencies including PyHive and pyodbc. These are not required for spark-df to
work. This commit moves pyhive and odbc into extras. So from now on,
`soda-core-spark` will only install the base spark support and hive and odbc
support can be installed using
`pip install soda-core-spark[hive]` or `pip install soda-core-spark[odbc]`

### DIFF
--- a/soda/spark/setup.py
+++ b/soda/spark/setup.py
@@ -12,10 +12,7 @@ package_name = "soda-core-spark"
 package_version = "3.0.4"
 description = "Soda Core Spark Package"
 
-requires = [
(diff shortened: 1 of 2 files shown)


vijaykiran:
@janet-can  This PR changes the installation procedure for soda-core-spark from version 3.0.5

janet-can:
Got it-- I'll prep some changes in a docs branch to push live when 3.0.5 merges.

PR #1572, opened 2022-09-06
SAS-322 Handle numerical column/table names

m1n0:


soda/core/soda/execution/identity.py
```diff
@@ -44,7 +44,7 @@ def create_identity(
             if hash:
                 parts.append(hash)
 
-        return "-".join(parts)
+        return "-".join([str(p) for p in parts])
```
m1n0, line 47: This is unsafe code and could result in unexpected things appearing in the identity without anyone noticing, we should at least log a warning. I wanted to add it but logger is not present super easily, should we create a backlog ticket to resolve this later @vijaykiran ?

vijaykiran (approved):

PR #1586, opened 2022-09-14
Basic Sqlserver regex support

m1n0:


m1n0:
@vijaykiran I carried over most of what was done in soda-sql and was applicable, but this is still quite far from making the CI green for all the tests that we have that use regexes.
Furthermore, we have more formats support in soda-core and we dont really test nearly enough of them to actually verify that the regexes we use actually work well. Since sqlserver regex support is so limited it would make sense to cover all of these with tests, but that would be very time consuming.

PR #1588, opened 2022-09-14
SODA-914: add context authentication and impersonation for BigQuery

tooobsias:
With GCP/BigQuery one has the option to use different methods to authenticate.
The standard authentication methods are:

1. Use of Application Default Credentials
2. Use of Application Default Credentials with Service Account impersonation
3. Use of Service Account Key (already implemented)
4. Use of Service Account Key with Service Account Impersonation  

###  Usage in configuration.yml file:
1. Use of Application Default Credentials
```
data_source <SOURCE-NAME>:
  type: bigquery
  connection:
    use_context_auth: True
```
2. Use of Application Default Credentials with Service Account impersonation
```
data_source <SOURCE-NAME>:
  type: bigquery
  connection:
    use_context_auth: True
    impersonation_account: <SA_EMAIL>
```
3. Use of Service Account Key
...
4. Use of Service Account Key with Service Account Impersonation
```
data_source my_database_name:
  type: bigquery
  connection:
    account_info_json: '{
        "type": "service_account",
        "project_id": "...",
        "private_key_id": "...",
      ...}'
    impersonation_account: <SA_EMAIL>
```

@m1n0 
I have incorporated you requested changes regarding some duplicate lines. Thanks for the remark.

### DIFF
--- a/soda/bigquery/soda/data_sources/bigquery_data_source.py
+++ b/soda/bigquery/soda/data_sources/bigquery_data_source.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import logging
 
+from google.auth import default, impersonated_credentials
 from google.cloud import bigquery
 from google.cloud.bigquery import dbapi
 from google.oauth2.service_account import Credentials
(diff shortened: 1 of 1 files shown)


tooobsias:
@m1n0 There is currently only one consideration regarding the project_id:

If using the "Application Default Credentials" or the "Impersonation" it could be possible that the `project_id` defaults to None., if the `project_id` is not manually set in the configurations. This could lead to some issues downstream. 

Should we raise an exception if the `project_id` is not set in the "Application Default Credentials" or via the "impersonation"
or leave this to the users control ?

m1n0:
@tooobsias I dont think we need to handle such situation, google auth library gives reasonable error:
```
Encountered a problem while trying to connect to bigquery: Project was not passed and could not be determined from the environment.
```
so we dont need to replicate that. 

Thank you for the contribution!

PR #1606, opened 2022-09-27
Use filename in check identity

m1n0:


soda/core/soda/execution/check/check.py
```diff
@@ -190,6 +190,10 @@ def create_identity(self, with_datasource: bool = False) -> str:
         if with_datasource:
             hash_builder.add(self.data_source_scan.data_source.data_source_name)
 
+        if with_filename:
+            _, _, filename = self.check_cfg.location.file_path.rpartition("/")
```
vijaykiran, line 194: I think this will fail on windows, maybe use path separator use https://docs.python.org/3/library/os.path.html#os.path.basename

vijaykiran (changes requested):

PR #1611, opened 2022-09-28
fix: ensure PSI and SWD can deal with decimal.Decimal type

tituskx:
This PR:

* Fixes a bug that was encountered while running the PSI test on data containing `decimal.Decimal` data: see [here](https://soda-community.slack.com/archives/C038FFU79J5/p1664247672138869). 
* Fixes a similar bug for the SWD test that was found while fixing the above bug
* Adds to unit tests to ensure that the PSI and SWD now work for values of type `decimal.Decimal`

For both the PSI test and the SWD test this PR ensures that we first check whether `self.test_data` contains any values of type `decimal.Decimal`. If yes, the data in `self.test_data` is converted to type `float`.

**Note 1**

For the PSI test an alternative approach could be to convert just `min_val` and `max_val` to floats if they are not already (instead of `self.test_data` entirely).  However, it would be important that `float(min_val) <= min_val` and `float(max_val) >= max_val` to ensure that all the values in the data are contained in the bins that are created using `np.linspace(float(min_val), float(max_val), 11)`.

Since `float(min_val) <= min_val` and `float(max_val) >= max_val` are not guaranteed as `float(arg)` returns the _closest_ float to `arg`, we would have to round down `float(min_val)` to the nearest float smaller than `float(min_val)` and round up `float(max_val)` to the nearest float larger than `float(max_val)`. 

However, that solution is 

1. More verbose
2. Does not lead to faster running time as using the `decimal.Decimal` type for following computation is slower than using floats

**Note 2**

Probably not super important but wanted to mention it anyways: 

Since PSI uses histograms, converting `self.test_data` to floats when it contains values of type `decimal.Decimal` does mean that values could be theoretically allocated to the incorrect bin if they lie within machine precision of a bin edge. The probability of this happening is generally extremely small though.

If this is something that we want to tackle in the (far) future then I can create a ticket - I don't see any harm in doing this the correct way and I might miss an obvious case where this would be important.

### DIFF
--- a/soda/scientific/soda/scientific/distribution/comparison.py
+++ b/soda/scientific/soda/scientific/distribution/comparison.py
@@ -71,13 +71,15 @@ class DistributionChecker:
             "semd": SWDAlgorithm,
             "psi": PSIAlgorithm,
         }
-        
+
         self.choosen_algo = algo_mapping.get(self.dist_method)
 
     def run(self) -> Dict[str, float]:
         test_data = pd.Series(self.test_data)
-        if (self.dist_method in ["semd", "swd", "psi"]) and pd.core.dtypes.common.is_dtype_equal(test_data, decimal.Decimal):
-            test_data = test_data.astype('float')
+        if (self.dist_method in ["semd", "swd", "psi"]) and pd.core.dtypes.common.is_dtype_equal(
+            test_data, decimal.Decimal
+        ):
+            test_data = test_data.astype("float")
 
         bootstrap_size = 10
         check_values = []
--- a/soda/scientific/soda/scientific/distribution/comparison.py
+++ b/soda/scientific/soda/scientific/distribution/comparison.py
@@ -1,7 +1,7 @@
 import abc
 import logging
 from typing import Any, Dict, List, Tuple, Union
-
+import decimal
 import numpy as np
 import pandas as pd
 from ruamel.yaml import YAML, YAMLError
@@ -240,6 +240,9 @@ class KSAlgorithm(DistributionAlgorithm):
 
 class SWDAlgorithm(DistributionAlgorithm):
     def evaluate(self) -> Dict[str, float]:
+        if any(isinstance(i, decimal.Decimal) for i in self.test_data):
+            self.test_data = self.test_data.astype('float')
+
         wd = wasserstein_distance(self.ref_data, self.test_data)
         swd = wd / np.std(np.concatenate([self.ref_data, self.test_data]))
         return dict(check_value=swd)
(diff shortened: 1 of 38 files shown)


soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -240,15 +241,22 @@ def evaluate(self) -> Dict[str, float]:
...
+            self.test_data = self.test_data.astype("float")
+
         wd = wasserstein_distance(self.ref_data, self.test_data)
         swd = wd / np.std(np.concatenate([self.ref_data, self.test_data]))
         return dict(check_value=swd)
 
 
 class PSIAlgorithm(DistributionAlgorithm):
     def evaluate(self) -> Dict[str, float]:
+        if any(isinstance(i, decimal.Decimal) for i in self.test_data):
```
bastienboutonnet, line 254: what does this loop iterate over? is it rows?

bastienboutonnet (commented):
I have a question about the for loop. I could dig into the code but I think it'll be quicker for you to answer that.

soda/scientific/soda/scientific/distribution/comparison.py
tituskx: Yeah it's rows..

You would probably prefer something like

```if pd.core.dtypes.common.is_dtype_equal(self.test_data, Decimal)```

?

Also,  I'll move this type check higher up in the code to `DistributionChecker.run()`, as we run each test x times and currently convert `test_data` to float x times.

Comment (bastienboutonnet) [reply soda/scientific/soda/scientific/distribution/comparison.py:254]: Yeah I would like to see if we could find a method that doesn't force us to check each value. I would imagine that if any value in that column ends up being decimal then most likely the type of the column will not be `float` right? in that case then we apply. So maybe the function you refer to would work and I think it would be better provided it works as you intend it.

As to your second comment, I think I agree. It doesn't seem like the best idea to do this over and over again especially if it's costly.

Comment (tituskx) [reply soda/scientific/soda/scientific/distribution/comparison.py]: Yes good point!
I have tested `pd.core.dtypes.common.is_dtype_equal(self.test_data, Decimal)` and it returns `True` if there is at least one `decimal.Decimal`. If there are only values of type `float` then it returns `False`

File: soda/scientific/soda/scientific/distribution/comparison.py [hunk 2]
```diff
@@ -249,6 +255,7 @@ class PSIAlgorithm(DistributionAlgorithm):
     def evaluate(self) -> Dict[str, float]:
         max_val = max(np.max(self.test_data), np.max(self.ref_data))
```
mrafayaleem, line 256: Is `ref_data` always of `float` type?

PR #1616, opened 2022-09-29
fix: handle very large bin sizes by filtering out outliers for dro generation

baturayo:
Resolves: https://soda-community.slack.com/archives/C038FFU79J5/p1664395865334819

### DIFF
--- a/soda/scientific/soda/scientific/distribution/generate_dro.py
+++ b/soda/scientific/soda/scientific/distribution/generate_dro.py
@@ -43,22 +43,13 @@ class DROGenerator:
         lower_range = q1 - (1.5 * IQR)
         upper_range = q3 + (1.5 * IQR)
         filtered_data = data[np.where((data >= lower_range) & (data <= upper_range))]
+        # TODO: insert doc link that explains IQR
         logging.warning(
-            f"""While generating the distribution reference object, the automatic bin size detection
-exceeded the memory. It is generally caused by outliers in the data. You can reproduce the issue
-with the following example
+            f"""Generating the distribution reference object using automatic bin size detection would cause a
+memory error. This is generally caused by the presence of outliers in the dataset which leads to very high number
+of bins.
 
-```
-training_input = np.random.rand(7000)
-training_input[1000] = 1000000000000000
-np.histogram(training_input, bins="auto")
-```
-
-To fix the issue, we filtered the outliers using the Interquanlite Range (https://en.wikipedia.org/wiki/Interquartile_range)
-The values  which are less than 1st Ouartile - 1.5 IQR and more than 3rd Quartile + 1.5 IQR has been removed from the data.
-
-lower limit (1st Ouartile - 1.5 IQR) = {lower_range}
-upper limit (3rd Ouartile + 1.5 IQR) = {upper_range}
+We filtered values above {lower_range} and below {upper_range} using IQR
 """
         )
         return filtered_data
@@ -97,18 +88,15 @@ has been ignored!
                 n_sqrt_bins = int(np.ceil(math.sqrt(outlier_filtered_data.size)))
                 if n_sqrt_bins < self.maximum_allowed_bin_size:
                     logging.warning(
-                        f"""Filtering out outliers does not solved the memory error, so we will
-take the square root of the data size to set the number of bins where;
-
-n_bins = {n_bins}
-sqrt(n_bins) = {n_sqrt_bins}
+                        f"""Filtering out outliers did not solve the memory error. As a last resort, we will
+take the square root of the data size to set the number of bins.
 """
                     )
                     weights, bins = np.histogram(outlier_filtered_data, bins=n_sqrt_bins, density=False)
                 else:
                     logging.warning(
-                        f"""We set n_bins={self.maximum_allowed_bin_size} as maximum since your
-n_bins={n_bins} which is higher then {self.maximum_allowed_bin_size}"""
+                        f"""We set n_bins={self.maximum_allowed_bin_size} as maximum since
+automatically computed {n_bins} is higher than maximum allowed bin size: {self.maximum_allowed_bin_size}"""
                     )
                     weights, bins = np.histogram(
                         outlier_filtered_data, bins=self.maximum_allowed_bin_size, density=False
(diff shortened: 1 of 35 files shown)


soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -16,14 +17,51 @@ class Config:
...
+    @staticmethod
+    def _remove_outliers_with_iqr(data: np.ndarray) -> np.ndarray:
+        # Remove outliers
+        q1, q3 = np.percentile(data, [25, 75])
+        IQR = q3 - q1
+        lower_range = q1 - (1.5 * IQR)
+        upper_range = q3 + (1.5 * IQR)
+        filtered_data = data[np.where((data >= lower_range) & (data <= upper_range))]
+        logging.warning(
+            f"""While generating the distribution reference object, the automatic bin size detection
```
bastienboutonnet, line 47: Here I would rephrase a bit:

"Generating the distribution reference object using automatic bin size detection would cause a memory error. This is generally cause by the presence of outliers in the dataset which leads to <INSERT_REASON>."

I think we might want to cut out the part where we give the code, and instead make a PR to the docs repo where we could give some details about this. And instead, provide a link to said docs page.

Then continue with "We filtered values above X and below Y using IQR and provide the link."

Then I would also make sure we have a more verbose explanation of this in the docs PR so that we can also add a link here.

soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -33,11 +71,48 @@ def generate_continuous_dro(self) -> DRO:
...
+            # If n_bins is lower than data size then run auto mode again
+            if n_bins < min(self.maximum_allowed_bin_size, data.size):
+                weights, bins = np.histogram(outlier_filtered_data, bins="auto", density=False)
+
+            # If not then take the sqrt of data size and make it as our new n_bins
+            else:
+                n_sqrt_bins = int(np.ceil(math.sqrt(outlier_filtered_data.size)))
+                if n_sqrt_bins < self.maximum_allowed_bin_size:
+                    logging.warning(
+                        f"""Filtering out outliers does not solved the memory error, so we will
```
bastienboutonnet, line 100:
```suggestion
                        f"""Filtering out outliers dit not solve the memory error. As a last resort we will
```

soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -33,11 +71,48 @@ def generate_continuous_dro(self) -> DRO:
...
+take the square root of the data size to set the number of bins where;
+
+n_bins = {n_bins}
+sqrt(n_bins) = {n_sqrt_bins}
```
bastienboutonnet, line 104: I'm not entirely sure whether the square root is helpful to users here what do you think?

soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -33,11 +71,48 @@ def generate_continuous_dro(self) -> DRO:
...
+"""
+                    )
+                    weights, bins = np.histogram(outlier_filtered_data, bins=n_sqrt_bins, density=False)
+                else:
+                    logging.warning(
+                        f"""We set n_bins={self.maximum_allowed_bin_size} as maximum since your
+n_bins={n_bins} which is higher then {self.maximum_allowed_bin_size}"""
```
bastienboutonnet, line 111:
```suggestion
n_bins={n_bins} which is higher than {self.maximum_allowed_bin_size}"""
```

bastienboutonnet (commented):
Since we reviewed the approach IRL and I'm ok I would say this is good to go from my side. But we could improve the wording and amount of information that we give to users in the warnings.

soda/scientific/soda/scientific/distribution/generate_dro.py
baturayo: true I'll delete it

bastienboutonnet (approved):
All good! Let's make sure we post in the community thread and ping @vijaykiran so he can chime in on when he might tag the next release.

soda/scientific/soda/scientific/distribution/generate_dro.py
mrafayaleem: Also, just for my own understanding and may be this can be added as a comment if true, the rational behind sqrt is this:

1. data size >= `1e12`: sqrt potentially brings down the bins to <= `1e6`
2. data size <= `1e6` for example `999999`: This data size cannot exceed `1e6` bins even with auto so that's safe 
3. data size = `2e6`: `auto` can give a value > `1e6` and sqrt brings down the number of bins to `1414` which is quite drastic compared to the default `1e6` limit

In a case like 3, wouldn't a strategy like exponential backoff be better to get close to the max limit instead of getting to 0.14% of the potential bin size?

baturayo: Hi @mrafayaleem,

Thanks for your feedback and follow-up @mrafayaleem :) 

So, we first check whether the number of bins is greater than the data size. It might be the case like the following array [0.5, 0.43, 0.54, 0.2, 0.1, 0.3 and 1000000000000] generates a super large number of bins as you find out. 

If this is the case, we apply IQR, filter outlier values, and recompute the number of bins. If it still exceeds the data size, then we take the sqrt of data size to use as the number of bins (default strategy of excel to find the number of bins).

If sqrt(data size) still exceeds 1e6, we just wanted to limit an upper boundary as 1e6 to prevent memory issues. (While writing this, I discover a small corner case in the PR so will fix it 👍 )

For your suggestion, exponential backoff seems to be an easy and effective solution as well. We will definitly consider it with the team.

soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -33,11 +62,45 @@ def generate_continuous_dro(self) -> DRO:
             data = data[~np.isnan(data)]
             logging.warning(
                 f"""{none_count} out of {data_len} rows ({none_count/data_len}%)
```
vijaykiran, line 64: Will this ever fail with divide by zero? or `data_len` is guaranteed to be non-zero?

PR #1624, opened 2022-10-05
fix: limit the bin size and handle zero division for continious DRO

baturayo:


soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -30,10 +30,17 @@ def __init__(self, cfg: RefDataCfg, data: list) -> None:
...
+            n_bins = 1
+            return n_bins
+
         first_edge, last_edge = np.lib.histograms._get_outer_edges(data, _range)
-        n_bins = int(np.ceil(np.lib.histograms._unsigned_subtract(last_edge, first_edge) / bin_width))
-        return n_bins
+
+        bin_width = np.lib.histograms._hist_bin_auto(data, _range)
+
+        n_bins = min(np.ceil(np.lib.histograms._unsigned_subtract(last_edge, first_edge) / bin_width), data.size)
```
baturayo, line 42: I set the maximum `n_bins` to the size of data because the bin size larger than the data doesn't add any value. We can prevent infinite bin sizes (causing overflow error) thanks to this limitation.

bastienboutonnet (approved):

Issue #1610, opened 2022-09-28
Schema Check is Unable to Work when Variable is passed


essayyzed:
I am facing an issue when defining variables for my check. The issue is that the variable works fine in all the checks other than `Schema Check`  where is raise the following error:
`"value": [{"columnName": c["name"], "sourceDataType": c["type"]} for c in self.value],
TypeError: 'Undefined' object is not iterable.`

My checks.yml looks like this:
`variables:`
    `table: 'customers'`
`checks for ${table}:`
 ` - schema:`
      `fail:`
        `when required column missing:`
          `- id`
          `- name`
What could be the possible reason for this error.? Or is it a bug related to `variable` usage in `Schema Check`

jmarien:
SODA-1197

vijaykiran:
Closed via #1628

ceyhunkerti:
this issue appears again after the last commit that modifies `schema_query.py`

PR #1628, opened 2022-10-05
fix: variable substitution in schema check query

ceyhunkerti:
fixes variable substitution issue with schema checks. 
https://github.com/sodadata/soda-core/issues/1610


vijaykiran:
@ceyhunkerti  Thanks a lot for PR! Can you please also add a test for this?

PR #1629, opened 2022-10-07
chore: pin prophet to >=1.1

bastienboutonnet:
When tested locally I am able to install the scientific library without issues but I used to never have any issues.

@jmarien and I were able to build the docker image locally as well

This is an attempt to not having to install `pystan` anymore which causes known issues with some of our customers.


bastienboutonnet:
I tested using the docker image to run soda and anomaly detection worked correctly inside the docker so I assume when running in the agent things will be peachy.

jmarien (approved):

PR #1636, opened 2022-10-13
Verify data source connection command

m1n0:
Still a little bit WIP, needs verification of all data sources, and I will also go through each data source and try to see whether there is a more meaningful query to run in each.


m1n0:
`Select 1` seems reasonable in all data sources, I added a skipif for spark_df as it doesnt really work with regular CLI anyway. Also added a fail test for the cli command. Ready for review now.

vijaykiran (approved):

PR #1638, opened 2022-10-14
fixing bq separate compute storage project

thiagodeschamps:
Soda-core after the 3.0.9 update was supposed to let us be able to run a query job in different project that the table is stored, but something was happening. Was observed that Soda was querying the table in the project that the Query Job was just supposed to run. After that, we came to a conclusion that, due to the project and the dataset be defined in the `QueryJobConfig` (see line 136 in the [commit](https://github.com/sodadata/soda-core/pull/1598/commits/a0cad80b3b44b3bc7ca574a6f3c85cda0bc37605)), and also the dataset be declared in the SQL statement, SQL built overrides the parameter `default_dataset` and ignores the `storage_project_id` declared, just letting the project_id which is the project that will run the query.

With this in mind, we made the following solution: redefine the `_create_table_prefix` method in the file bigquery_data_source.py, so it returns None when creating the table prefix, so the previous project.dataset will not be override by the SQL statement built.

p.s.: this problem only happens when using Soda CL because the SQL statement is built automatically

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
       - id: autoflake
         args: ["--in-place", "--remove-all-unused-imports"]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.0.0
+    rev: v3.1.0
     hooks:
       - id: pyupgrade
         args: [--py37-plus]
--- a/soda/bigquery/soda/data_sources/bigquery_data_source.py
+++ b/soda/bigquery/soda/data_sources/bigquery_data_source.py
@@ -114,8 +114,6 @@ class BigQueryDataSource(DataSource):
         self.project_id = data_source_properties.get("project_id", self.project_id)
 
         self.dataset = data_source_properties.get("dataset")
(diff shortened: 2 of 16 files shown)


vijaykiran:
@m1n0  can you review and merge this?

m1n0:
Thank you for testing and looking into this! 

I am thinking whether we could maybe do this a bit differently. We set the `schema` property manually (or one could say even "artificially") and that sets the prefix as well - [here](https://github.com/sodadata/soda-core/blob/04938d26d8ee495a5592debc806bd66a3d9eb656/soda/bigquery/soda/data_sources/bigquery_data_source.py#L117)

I am thinking whether removing this line would not have the same desired effect as your proposed change - but rather simplifying thing and not adding a method override. 

I do not currently have a BQ setup to test this super easily, do you perhaps have such setup, if so could you test that out? Thanks!

thiagodeschamps:
Hi, @m1n0!

That's great, really liked your solution, it's more simple and elegant. I changed here, tested and everything works completely fine! I've Just pushed the code also. Thank you very much for reviewing.

PR #1645, opened 2022-10-18
Change over time - add same day/month support

m1n0:


m1n0:
@vijaykiran it does not make much sense to test the actual cloud api request that gets the appropriate historical data, but I will still double check that this functionality works ok with a manual local test setup, I just havent had time to to get the historical data backfilled yet.

vijaykiran:
I think there is an extra `.antlr` folder the PR - can you delete that?

PR #1650, opened 2022-10-20
[cloud] don't send null file ref, when failed rows are disabled

vijaykiran:


vijaykiran:
Needs manual testing

m1n0 (approved):
I havent tested manually but code LGTM

PR #1652, opened 2022-10-22
Modify Test Column Names

tdstark:
### Description
In preparation for adding the Oracle dialect, I modified several test column names which leverage common reserved words, namely `size` and `date`. These were changed to `cst_size`, which can be read as customer size, and `date_updated`. I also modified the `sizeText` column to be `cst_size_text` so that wildcard tests would still work with `test_forbidden_columns_fail_matching_wildcard` using a similar pattern.

Note that this was tested using Postgres exclusively.

### Question
One outstanding question I have is that if this PR is pushed, we don't have a camel case column in the test tables anymore. I'm not sure if we doing any compatibility testing for column quoting. Do we need a camel case column?


vijaykiran:
@tdstark  thanks a lot for the PR, I'll take a look. We do want to test cases where we have to use quoted column names so the case is preserved.

tdstark:
> @tdstark thanks a lot for the PR, I'll take a look. We do want to test cases where we have to use quoted column names so the case is preserved.

Do we want to add a new column?

PR #1655, opened 2022-10-24
feat: allow use of in-check filters for distribution checks

tituskx:
This PR resolves the following: [Allow filters to be used on the column used in the distribution checks, not just on the DRO](https://sodadata.atlassian.net/browse/CLOUD-207)

This PR ensures that [in-check filters](https://docs.soda.io/soda-cl/filters.html#configure-in-check-filters) can be used for distribution checks. Previously it was only possible to implement [dataset filters](https://docs.soda.io/soda-cl/filters.html#configure-dataset-filters) for distribution checks. Specifically, this PR does the following

* In `distribution_check.py` we examine whether an in-check filter is used; if there is, it is added to the query `where` condition. If there are both a dataset filter and a in-check filter present we use both. Generally people shouldn't use dataset and in-check filters for the same checks, see [here](https://docs.soda.io/soda-cl/filters.html#configure-dataset-filters). However, it seems that the current behaviour in core is to apply both filters if they are provided (checked that this is true for `MetricChecks`) so that's why I implemented it in the same way here.
* In `sodacl_parser.py` the parsed filter is passed to the `DistributionCheckCfg` so that it can be used when the sql query is created.
* In `comparison.py` a new Exception is created that is raised when `test_data` is empty. This can happen if none of the data in the column satisfies the filter that is used (previously people got a pretty arcane `sklearn` error). This PR's implementation raises a `LoggableException` in `comparison.py` like we agreed with Vijay a couple of months ago; scientific raises `LoggableExceptions` that are then caught and logged by `soda-core`. We could also log this error higher up in the code, [here](https://github.com/sodadata/soda-core/blob/5640def6e2f0e107bfb427bd18ee4183afd5e4ba/soda/core/soda/execution/check/distribution_check.py#L60) for example. Let me know if that's what you would prefer.


bastienboutonnet:
Thanks a lot @tituskx I think this looks good and thanks for checking how soda-core does it on metric checks. I think you're correct in keeping the two behaviours consistent.

It would be nice if we could extend the tests. 

We have some tests that cover the parsing here: https://github.com/sodadata/soda-core/blob/main/soda/core/tests/data_source/test_distribution_check.py which I think would be a good place to add coverage. 

What do you think?

bastienboutonnet (commented):
Thanks a lot @tituskx I think this looks great and thanks for checking how soda-core does it on metric checks. I think you're correct in keeping the two behaviors consistent.

It would be nice if we could extend the tests.

We have some tests that cover the parsing here: https://github.com/sodadata/soda-core/blob/main/soda/core/tests/data_source/test_distribution_check.py which I think would be a good place to add coverage.

What do you think?

soda/scientific/soda/scientific/distribution/comparison.py
```diff
@@ -59,6 +63,12 @@ def __init__(
         dist_name: Union[str, None],
         data: List[Any],
     ):
+        if len(data) == 0:
```
baturayo, line 66: @tituskx not sure if we have tests for exceptions but I would add a test to cover this scenerio

soda/core/soda/execution/check/distribution_check.py
```diff
@@ -121,10 +121,20 @@ def sql_column_values_query(self, distribution_check_cfg, limit=1000000) -> str:
 
         partition_filter = self.partition.sql_partition_filter
         partition_str = ""
+
+        distribution_check_filter = distribution_check_cfg.filter
+
+        if distribution_check_filter:
```
baturayo, line 127: I would add a unit test to soda-core to test the `filter` on distribution checks. I will implement `sample` query to dist check and I want to make sure that my changes won't break anything. This part will mostly change in my future PR. However, I have the necessary info from this PR.

bastienboutonnet (approved):
I think this is ready to merge @tituskx. Thanks for your work on this!

PR #1666, opened 2022-11-02
feat: support sampling for distribution checks

baturayo:


soda/core/soda/execution/check/distribution_check.py
```diff
@@ -115,31 +115,36 @@ def get_log_diagnostic_dict(self) -> dict:
...
-            scan = self.data_source_scan.scan
-            resolved_filter = scan.jinja_resolve(definition=partition_filter)
-            partition_str = (
-                f"{partition_str} AND ({resolved_filter})" if partition_str else f"\nWHERE {resolved_filter}"
+        filter_clause = None
+        limit = int(1e6)
+
+        if partition_filter and distribution_check_filter:
+            self.logs.error(
+                "Cannot have both dataset filter and distribution check filter at the same time",
```
bastienboutonnet, line 131: Why are we strict about this?

bastienboutonnet (commented):
Looks good but I have one question about the error on not being able to have dataset and check filters at the same time. Just want to know what the thinking is because as far as I'm aware they could totally be appended as `and`s

soda/core/soda/execution/check/distribution_check.py
baturayo: For some reason, I was thinking that at soda-core it wasn't allowed to use both check filters and dataset filters at the same time. Indeed, it can be used together. I will fix it with `AND` operator

bastienboutonnet: Ok makes sense. At least for this check it works for sure since we're not doing crazy case and stuff. I think we don't even need to warn since people will know they have applied a filter from the dataset level due to the fact that they write:
`checks for dataset [filter_name]`

baturayo: done

bastienboutonnet (approved):
Looks good. Thanks @baturayo! Could you open a PR for docs.soda.io as well to document this?

PR #1671, opened 2022-11-07
feat: dask and pandas support

baturayo:
Dask and pandas support will enable soda-core to run checks on dask and pandas dataframes. The current python API will look like;

```python
import dask.datasets
import pandas as pd
from soda.scan import Scan

# Create soda scan object
scan = Scan()
scan.set_scan_definition_name("test")
scan.set_data_source_name("dask")

# Load timeseries data from dask datasets
df_timeseries = dask.datasets.timeseries().reset_index()
df_timeseries["email"] = "a@soda.io"

# Create an artificial pandas dataframe
df_employee = pd.DataFrame({"email": ["a@soda.io", "b@soda.io", "c@soda.io"]})

# Add dask dataframe to scan and assign a dataset name to refer from checks yaml
scan.add_dask_dataframe(dataset_name="timeseries", dask_df=df_timeseries)

# Add pandas dataframe to scan and assign a dataset name to refer from checks yaml
scan.add_pandas_dataframe(dataset_name="employee", pandas_df=df_employee)

# Define checks in yaml format
# alternatively you can refer to a yaml file using scan.add_sodacl_yaml_file(<filepath>)
checks = """
for each dataset T:
  datasets:
    - include %
  checks:
    - row_count > 0

profile columns:
  columns:
    - employee.%

checks for employee:
    - values in (email) must exist in timeseries (email) # Error expected
    - row_count same as timeseries # Error expected

checks for timeseries:
  - invalid_count(email) = 0:
      valid format: email
  - valid_count(email) > 0:
      valid format: email
  - duplicate_count(name) < 4:
      samples limit: 2
  - missing_count(y):
      warn: when > -1
  - missing_percent(x) < 5%
  - missing_count(y) = 0
  - avg(x) between -1 and 1
  - max(x) > 0
  - min(x) < 1:
      filter: x > 0.2
  - freshness(timestamp) < 1d
  - values in (email) must exist in employee (email)
"""

scan.add_sodacl_yaml_str(checks)

scan.set_verbose(True)
scan.execute()
```


soda/core/soda/execution/metric/numeric_query_metric.py
```diff
@@ -122,6 +124,23 @@ def set_value(self, value):
             self.value = int(value)
         elif isinstance(value, Decimal):
             self.value = float(value)
+        # Check if numpy is installed
+        elif find_spec("pandas") and find_spec("numpy"):
```
m1n0, line 128: I personally would prefer if pandas+numpy specific logic would be in a subclass or something like that instead of mixing it here. We do not exactly have a mechanism for that in the metric classes, I will suggest something if I come up with a something worthwhile considering.

examples/pandas_dask_example.py
```diff
@@ -0,0 +1,77 @@
+import dask.datasets
```
baturayo, line 1: Wdyt @vijaykiran @m1n0  about adding an `examples` folder to the main directory to keep python API examples? We can eventually add other data sources like spark for the contributors.

PR #1687, opened 2022-11-15
fix: ensure that profiling does not lowercase columns 

tituskx:
This PR resolves: [Profiling shows duplicate column names for Snowflake](https://sodadata.atlassian.net/browse/SAS-714) and [CLOUD-2123](https://sodadata.atlassian.net/browse/CLOUD-2123)

This PR implements the following changes:

* Remove the line `column_name = column_name.lower()` from [here](https://github.com/sodadata/soda-core/blob/5da87c4eb3f226719cf73a74c4b92684597322a5/soda/core/soda/execution/check/profile_columns_run.py#L346)
* Creates a new test table that has a capitalised column name in it
* Add a new unit tests that ensures that the running the profiling on the new test table returns the capitalised version (if the database does not make the columns lowercase)
* For unit tests that test column profiling, we now base the expected output on the way the datasource type handles lower/upper-case columns. Setting the expected output based on the datasource is currently done in the unit tests themselves. This seemed like the best approach as it's just a few lines of code and does not seem to warrant a separate fixture/expected output file.

@bastienboutonnet In profiling we also lowercase the table names, see [here](https://github.com/sodadata/soda-core/blob/5da87c4eb3f226719cf73a74c4b92684597322a5/soda/core/soda/execution/check/profile_columns_run.py#L383) and [here](https://github.com/sodadata/soda-core/blob/5da87c4eb3f226719cf73a74c4b92684597322a5/soda/core/soda/execution/check/profile_columns_run.py#L398), but it seems that that is the intended behaviour (I checked and core actually does the same when it returns the table names [here](https://github.com/sodadata/soda-core/blob/1af51163b06509f56a5d0692d17ce59fef9dfd4f/soda/core/soda/execution/check/profile_columns_run.py#L56)).


bastienboutonnet (approved):
Very nice @tituskx thanks!

bastienboutonnet:
> @bastienboutonnet In profiling we also lowercase the table names, see [here](https://github.com/sodadata/soda-core/blob/5da87c4eb3f226719cf73a74c4b92684597322a5/soda/core/soda/execution/check/profile_columns_run.py#L383) and [here](https://github.com/sodadata/soda-core/blob/5da87c4eb3f226719cf73a74c4b92684597322a5/soda/core/soda/execution/check/profile_columns_run.py#L398), but it seems that that is the intended behaviour (I checked and core actually does the same when it returns the table names [here](https://github.com/sodadata/soda-core/blob/1af51163b06509f56a5d0692d17ce59fef9dfd4f/soda/core/soda/execution/check/profile_columns_run.py#L56)).

and yes no need to change anything about tables at this point as it behaves as intended and I'm 99% sure matching on cloud is case-insensitive for those.

vijaykiran (approved):

soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -86,10 +91,18 @@ def test_profile_columns_text(data_source_fixture: DataSourceFixture):
     profiling.pop("dataSource")
     profiling.pop("table")
 
+    test_data_source = os.environ.get("test_data_source")
+
+    uppercase_column_name_databases = "snowflake"
```
vijaykiran, line 96: Are you this is only snowflake? I suspect on db2 and oracle  (and potentially sqlserver) we also have similar issues

tituskx: Hey @vijaykiran it does indeed seem that `db2` and `oracle` use uppercase column names as well, but we don't have unit tests for those yet right? I have added them to the list so that once we do those database types will not fail CI. 

For `sqlserver` we do have unit tests (and they passed), so it seems to case sensitive.

Thanks!

vijaykiran: we don’t have CI for oracle or db2, only local ones. Oracle CI will be added this sprint. For db2 the container creation is too slow, and we don’t have enough production usage yet. 

For now, I’ll merge it. If oracle CI fails, I’ll let you know.

PR #1688, opened 2022-11-15
Sqlserver: fix email regex, do not allow empty string

m1n0:


vijaykiran:
@m1n0  can you merge this with #1691  ?

PR #1691, opened 2022-11-15
Date format fixes

vijaykiran:


vijaykiran:
@m1n0  Can you check this and merge with #1688

m1n0 (approved):

m1n0:
this unfortunately does not work on sqlserver, we use a limited version there with double digits and that version would allow a date like `99-99-9999` which is nonsense. We need to redo sqlserver regexes as well to match these

m1n0 (changes requested):
sqlderver needs to be matched

m1n0:
@vijaykiran I looked into sqlserver and well...without OR in the patterns we would have to use OR between complete patterns. This means we cannot do `<one digit date>OR<two digit date with leading 1/2>OR<two digit starting with 3 but ending with 0/1/2><same for monh><year part>` but rather we would have to do permutations as we can only match the whole string...I pushed a commit which skips the date tests for sqlserver, the sqlserver version of formats just support xx.xx.xxxx

geertjan-garvis:
> @vijaykiran I looked into sqlserver and well...without OR in the patterns we would have to use OR between complete patterns. This means we cannot do `<one digit date>OR<two digit date with leading 1/2>OR<two digit starting with 3 but ending with 0/1/2><same for monh><year part>` but rather we would have to do permutations as we can only match the whole string...I pushed a commit which skips the date tests for sqlserver, the sqlserver version of formats just support xx.xx.xxxx

Thanks, this makes sense to me if can't fix this soon/properly. Currently data formats are broken for every single data source, we should not block that fix due to a different issue with sql server

m1n0 (approved):

PR #1698, opened 2022-11-22
Refactor duplicate check into two queries

m1n0:


soda/core/soda/execution/data_source.py
```diff
@@ -551,25 +551,50 @@ def sql_information_schema_columns(self) -> str:
...
+            f"""
+            WITH frequencies AS (
+                SELECT {column_names}, COUNT(*) AS frequency
+                FROM {table_name}
+                WHERE {filter}
+                GROUP BY {column_names})
+            SELECT {column_names}, frequency
+            FROM frequencies
+            WHERE frequency {'<=' if invert_condition else '>'} 1
+            ORDER BY frequency DESC"""
```
m1n0, line 593: @vijaykiran FYI I added order by to both duplicate queries, it seems to make sense especially for the failed rows one with limit to get the ones with most duplicates

vijaykiran (approved):

PR #1713, opened 2022-12-06
Fix for history loss when adding identity

vijaykiran:


m1n0:
@vijaykiran I still think this makes sense, or not?

vijaykiran:
@m1n0  Yes, but I based the PR on 3.0.13 - I'll open one against main.

PR #1727, opened 2022-12-13
Check attributes validation

m1n0:


soda/core/soda/common/attributes_handler.py
```diff
@@ -24,65 +24,95 @@ def validate(self, attributes: dict[str, any], schema: list[dict]) -> tuple[dict
...
                     self.logs.error(f"Unsupported attribute type '{schema_type}'.")
 
+        if not schema_type:
+            self.logs.error(f"Soda Cloud does not recognize '{key}' attribute name.")
+
         return is_valid
 
     def format_attribute(self, value: any):
         # Introduce formatting methods similar to validation methods if this gets more complex.
+        if isinstance(value, date):
```
m1n0, line 46: adding both date and datetime support here, its super easy so lets do both

soda/core/soda/execution/check/check.py
```diff
@@ -174,11 +175,11 @@ def create_identity(self, with_datasource: bool = False, with_filename: bool = F
         if isinstance(check_cfg.source_configurations, dict):
 
             identity_source_configurations = dict(check_cfg.source_configurations)
-            # The next lines ensures that configuration properties 'name', 'samples limit', 'identity' are ignored
-            # for computing the check identity
+            # The next lines ensures that some of the configuration properties are ignored for computing the check identity
             identity_source_configurations.pop("name", None)
             identity_source_configurations.pop("samples limit", None)
             identity_source_configurations.pop("identity", None)
+            identity_source_configurations.pop("attributes", None)
```
m1n0, line 182: removing attributes from identity

soda/core/soda/scan.py
```diff
@@ -479,19 +479,28 @@ def execute(self) -> int:
...
-
-                        attributes, invalid_attributes = attribute_handler.validate(attributes, attributes_schema)
-
-                        if invalid_attributes:
-                            self._logs.debug(f"Skipping invalid attributes '{invalid_attributes.keys()}'.")
-
-                        check.attributes = attributes
+                        check_attributes = check.check_cfg.source_configurations.get("attributes", {})
+
+                        if check_attributes:
```
m1n0, line 484: This is a bit clumsy:
1. I commented out the actual validation code so that current users can still use the attributes syntax without any effect, they just get set to the check.attributes property as-is for now
2. this whole section will need to be moved to earlier stage so that scan can be halted, this will be done later when cloud is ready

soda/core/soda/soda_cloud/soda_cloud.py
```diff
@@ -185,14 +202,10 @@ def get_historic_data(self, historic_descriptor: HistoricDescriptor):
         return {"measurements": measurements, "check_results": check_results}
 
     def is_samples_disabled(self) -> bool:
```
m1n0, line 204: please give extra 👀 on this is_samples_disabled and how organization config is handled now - nothing popped out from local testing and tests and we do have a test for this https://github.com/sodadata/soda-core/blob/081386dc6ebf99457b623749529c1e6030359d04/soda/core/tests/data_source/test_samples.py#L34 but this would be pretty bad to have a bug introduced

vijaykiran (approved):

PR #1735, opened 2022-12-15
fix: table and column inclusion/exclusion works correctly for profiling

tituskx:
This PR resolves: [Column inclusion/exclusion does not appear to work as expected for profiling](https://sodadata.atlassian.net/browse/CLOUD-2344)

Specifically, this PR fixes the following: 

* An `exclude` statement excludes all the matching tables, regardless of the column pattern
* An `include` statement with a wildcard `%`  includes all the columns for the matching tables, regardless of the column matching pattern. I.e. `include dim_customer_%.id` includes all the columns in `dim_customer_soda`, not just the `id` column

#### Database type datasource profiling run

1. In `__parse_profile_columns_section` we iterate over the profile statements. This hasn't changed.
2. In the `run` method of `profile_columns_run.py`:
    1. We parse the inclusion and exclusion profiling statements found in step 1, by calling [self.parse_profiling_expressions](https://github.com/sodadata/soda-core/pull/1735/files#diff-11e199cae4a683928a0f00656d6144923d023a74db2414799e665e87a3bf29eeR51). We iterate over every expression and split on `"."`, so that we have a pattern used to match tables , and a pattern used to match columns. We store the table name and column name patterns a dict and append it to the `parsed_profiling_expressions` list.
    2. We call the [get_tables_columns_metadata()](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR426) method. Here we create the query required to get the correct tables and columns, by calling [sql_get_tables_columns_metadata()](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR484). In `sql_get_tables_columns_metadata` we use [create_table_column_sql_filters()](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR454) to create a sql filter for each pattern (every include and exclude statement). We iterate over the patterns and use the table and column name patterns to create sql conditions. Specifically, every include/exclude statement is converted to a sql condition by doing combining the table name and column name filter. 
      After creating a sql condition for each statement, the conditions that correspond to include profiling statements are combined with `" OR ".join(include_sql_filter_clauses)`. We use `OR` here because only one of these conditions needs to be true. The conditions that correspond to exclude profiling statements are created using the same approach for the include statements but are prepended with the `NOT` operator. When all the exclude patterns have been converted to sql filters we concatenate them using `" AND ".join(exclude_sql_filter_clauses)`. We use `AND` here because none of these conditions should be true. After both the include patterns and exclude patterns have been converted to sql conditions they are joined with the other filters (for database and schema) using `" \n  AND ".join(filter_clauses)`. We then select the table names, column names, and datatypes that match this `WHERE` filter. This approach ensures that we only select table names and column names that match to at least one of the include statements and none of the exclude statements.
    3. After creating the query, it is executed in [get_tables_columns_metadata()](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR443). We create a `defaultdict` called `tables_and_columns_metadata`. The keys of `tables_and_columns_metadata` are the tables that were returned. The values of `tables_and_columns_metadata` are dictionaries with `column_name`: `column_data_type` key/value pairs. 
    4. We iterate over the `table_names` in `tables_and_columns_metadata` and use `customers_columns_metadata_result = tables_and_columns_metadata.get(table_name)` to get the corresponding columns.
    5. We iterate over the columns in `customers_columns_metadata_result` and profile them. We don't have to check whether `customers_columns_metadata_result` is `None` anymore, as there has to be at least one column present, or the sql query would not have returned the table name at all. Some changes were made to the code here but the functionality is the same.

With this approach we do not have to use separate queries to get the columns for each table, but instead grab the tables and columns that should be profiled in one go.

#### Spark datasource profiling run

The logic used for spark is the same as for database type datasources but instead we use python to do the pattern matching.

1. Same parsing of expressions as before 
2. In [get_tables_columns_metadata()](https://github.com/sodadata/soda-core/pull/1735/files#diff-45cfeae51060d9628d9a479b98fb3b3731af789fb5aae5a02710e091deaab431R321) we call [get_included_table_names()](https://github.com/sodadata/soda-core/pull/1735/files#diff-45cfeae51060d9628d9a479b98fb3b3731af789fb5aae5a02710e091deaab431R278) to get the table names that can be matched to at least one include profiling statement and cannot be matched to any exclude profiling statement with a `%` on the column side. We do so by iterating over the table names in the datasource and using `any(self.pattern_matches(table_name, include_pattern["table_name_pattern"]) for include_pattern in include_patterns)` to check that the table name is in an include statement, and `not any(self.pattern_matches(table_name, exclude_pattern["table_name_pattern"]) and exclude_pattern["column_name_pattern"] == "%" for exclude_pattern in exclude_patterns)` to ensure that the table name is not in an exclude statement where all the columns are excluded.
3. We iterate over every table in the [`included_table_names`](https://github.com/sodadata/soda-core/pull/1735/files#diff-45cfeae51060d9628d9a479b98fb3b3731af789fb5aae5a02710e091deaab431R336)
    1. We get the columns that are in this table. We then [iterate over every column and check whether there is ](https://github.com/sodadata/soda-core/pull/1735/files#diff-45cfeae51060d9628d9a479b98fb3b3731af789fb5aae5a02710e091deaab431R352)
        1. an include pattern that matches to both the current table and column
        2. an exclude pattern that matches to both the current table and column
    2. if the table and column are in a matching include pattern and not in a matching exclude pattern the column gets added to a `defaultdict` called `tables_and_columns_metadata`. The keys of `tables_and_columns_metadata` are the table names and the values are dictionaries with key-value pairs `column_name`: `column_data_type`.
3. We proceed similarly to the database type datasources in `profiling_columns_run.py`.

**Update 11-01-2023**

- The `table_names_only` optional argument in [get_tables_columns_metadata](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR431) is new and is used to indicate whether the function should behave like `get_table_names` (it's not used yet but will be if we replace `get_table_names`, https://sodadata.atlassian.net/browse/CLOUD-2625)
- We previously queried for the column `is_nullable` when we called [column_metadata_columns](https://github.com/sodadata/soda-core/pull/1735/files#diff-6d2074eed6eb32f5431671aada76698c8643690c0f80e1980fb12e1d77e160cbR130-R131) in profiling, see  [sql_get_table_columns](https://github.com/sodadata/soda-core/blob/b00e4ffebe787e9a2bdb35f0e624a59a6661ed6c/soda/core/soda/execution/data_source.py#L469). Since it does not seemed to be used, I did not add it to the columns that are queried for the new function `get_tables_columns_metadata` for profiling, see [tables_columns_metadata](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR323-R329) and [metadata_columns = ", ".join(self.tables_columns_metadata())](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR532)

- Instead of hardcoding `information_schema.columns` columns like `table_name` and `table_schema` directly in methods like [sql_get_table_columns](https://github.com/sodadata/soda-core/blob/main/soda/core/soda/execution/data_source.py#L440) they are now returned by static methods in the [datasource.py](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR307-R330). This means that we can override this for other datasources, [like so for db2](https://github.com/sodadata/soda-core/pull/1735/files#diff-82e2d2eb7e1162ccc59ab8feebb3132da577f076e8060aec6338d99767c82c42R96-R114) without having to override the methods that create the sql ([this method](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR484) creates the sql for profiling for all the different datasources). 

- I have added a method called `catalog_column_filter` in [datasource.py](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR480). This method ensures that we use [the database filter](https://github.com/sodadata/soda-core/pull/1735/files#diff-27ca57b93de9533353057240193320fefec9e43ca4c4efdb970bce150013b3eeR511) only if there is a `table_catalog`-like column in the `information_schema.columns` equivalent of the particular datasource. For [db2](https://www.ibm.com/docs/en/db2oc?topic=views-syscatcolumns) and [oracle](https://docs.oracle.com/database/121/REFRN/GUID-F218205C-7D76-4A83-8691-BFD2AD372B63.htm#REFRN20277), those columns do not exists and that's [why they override](https://github.com/sodadata/soda-core/pull/1735/files#diff-7928fb6ba5d7979930dd46d000d5896836a825792a3d3164a09ecdee5c2396fbR105) the `catalog_column_filter` to return `""`. It's not pretty so let me know if you see alternatives. The reason I added this part is so that we don't have to override the entire `sql_get_tables_columns_metadata` method for those datasources. It would be great if we can test db2 and oracle somehow to ensure that it works.

- For db2 and bigquery we hardcoded that the left and right hand side of the conditions in `information_schema.columns` lowercase where lowercase, see [here for an example](https://github.com/sodadata/soda-core/blob/94bdcc217b04f5fa9d84c416dc1a4307811d5e71/soda/db2/soda/data_sources/db2_data_source.py#L118). This PR does not do the same and instead uses the `default_casify_...` methods. Let me know if that's incorrect.

[CLOUD-2344]: https://sodadata.atlassian.net/browse/CLOUD-2344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


soda/core/soda/execution/check/profile_columns_run.py
```diff
@@ -46,96 +47,88 @@ def __init__(self, data_source_scan: DataSourceScan, profile_columns_cfg: Profil
...
 
-        if len(table_names) < 1:
+        if tables_columns_metadata is None:
             self.logs.warning(
-                f"No table matching your SodaCL inclusion list found on your {self.data_source.data_source_name} "
-                "data source. Profiling results may be incomplete or entirely skipped",
+                f"Your SodaCL profiling expressions did not return any existing dataset name + column name combinations for your '{self.data_source.data_source_name}' "
+                f"data source. Please make sure that the patterns in your profiling expressions define existing dataset name + column name combinations."
+                f" Profiling results may be incomplete or entirely skipped. See the docs for more information: \n"
+                f"https://docs.soda.io/soda-cl/profile.html#display-profile-information-in-soda-cloud",
```
baturayo, line 79: ask maybe @janet-can to create a dynamic link for the doc.

soda/core/soda/execution/check/profile_columns_run.py
```diff
@@ -46,96 +47,88 @@ def __init__(self, data_source_scan: DataSourceScan, profile_columns_cfg: Profil
...
-                            profile_columns_result_table,
-                        )
-                    except Exception as e:
-                        self.logs.error(
-                            f"Problem profiling numeric column {table_name}.{column_name}: {e}",
-                            exception=e,
-                        )
+            columns_metadata_result = tables_columns_metadata.get(table_name)
+            # perform numerical metrics collection
+            numerical_columns = {
```
baturayo, line 92: while changing this chunk, what about creating 2 methods called `handle_numerical_columns` and `handle_text_columns` to clean this big method.

soda/core/soda/execution/check/profile_columns_run.py
janet-can: Of course!  Here you go: https://go.soda.io/display-profile

soda/core/soda/execution/data_source.py
```diff
@@ -419,6 +425,89 @@ def get_table_columns(
...
+        sql_filters = []
+        for profiling_pattern in profiling_patterns:
+
+            table_name_pattern = profiling_pattern["table_name_pattern"]
+            column_name_pattern = profiling_pattern["column_name_pattern"]
+
+            sql_filter = (
+                f"({casify_function}(table_name) LIKE {casify_function}('{table_name_pattern}')"
+                f" AND {casify_function}(column_name) LIKE {casify_function}('{column_name_pattern}'))"
+            )
```
tituskx, line 461:
```suggestion
            sql_filter = (
                f"({self.default_casify_table_name()}(table_name) LIKE {self.default_casify_table_name()}('{table_name_pattern}')"
                f" AND {self.default_casify_column_name()}(column_name) LIKE {self.default_casify_column_name()}('{column_name_pattern}'))"
            )
```

soda/core/soda/execution/data_source.py
```diff
@@ -419,6 +425,89 @@ def get_table_columns(
...
+
+        if self.database:
+            filter_clauses.append(
+                f"{self.default_casify_sql_function()}({self.column_metadata_catalog_column()}) = '{self.default_casify_system_name(self.database)}'"
+            )
+
+        if self.schema:
+            filter_clauses.append(
+                f"{self.default_casify_sql_function()}(table_schema) = '{self.default_casify_system_name(self.schema)}'"
+            )
```
tituskx, line 494:
```suggestion
        if self.database:
            filter_clauses.append(
                f"{self.default_casify_system_name()}({self.column_metadata_catalog_column()}) = '{self.default_casify_system_name(self.database)}'"
            )

        if self.schema:
            filter_clauses.append(
                f"{self.default_casify_system_name()}(table_schema) = '{self.default_casify_system_name(self.schema)}'"
            )
```

soda/core/soda/execution/check/profile_columns_run.py
```diff
@@ -46,97 +47,80 @@ def __init__(self, data_source_scan: DataSourceScan, profile_columns_cfg: Profil
...
+            exclude_patterns=exclude_patterns,
+            query_name=f"profile-columns-get-table-and-column-metadata",
         )
 
-        if len(table_names) < 1:
+        if tables_columns_metadata is None:
             self.logs.warning(
-                f"No table matching your SodaCL inclusion list found on your {self.data_source.data_source_name} "
-                "data source. Profiling results may be incomplete or entirely skipped",
+                f"Your SodaCL profiling expressions did not return any existing dataset name + column name combinations for your '{self.data_source.data_source_name}' "
```
baturayo, line 79: I would use `and` instead of `+`

soda/core/soda/execution/data_source.py
```diff
@@ -298,6 +299,11 @@ def column_metadata_columns() -> list:
         """Columns to be used for retrieving column metadata."""
         return ["column_name", "data_type", "is_nullable"]
 
+    @staticmethod
```
baturayo, line 302: if it's smth static, we can make it class property.

soda/core/soda/execution/check/profile_columns_run.py
```diff
@@ -46,97 +47,80 @@ def __init__(self, data_source_scan: DataSourceScan, profile_columns_cfg: Profil
...
                             table_name,
                             profile_columns_result_table,
                         )
-                    except Exception as e:
-                        self.logs.error(
-                            f"Problem profiling text column {table_name}.{column_name}: {e}",
-                            exception=e,
+                    else:
+                        self.logs.info(
+                            f"Column '{table_name}.{column_name}' was not profiled because column data "
```
baturayo, line 116: maybe warning would be more accurate

tituskx: Got it will do!

soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -156,113 +169,218 @@ def test_profile_columns_all_tables_all_columns(data_source_fixture: DataSourceF
...
+        for index, profiled_table in enumerate(profiled_tables, 1)
+    }
+
+    test_data_source = os.environ.get("test_data_source")
+
+    test_data_source_uppercase = test_data_source in UPPERCASE_COLUMN_NAME_DATABASES
+    test_data_source_lowercase = test_data_source in LOWERCASE_COLUMN_NAME_DATABASES
+
+    if test_data_source_uppercase or test_data_source_lowercase:
+        casify_function = lambda x: x.upper() if test_data_source_uppercase else x.lower()
```
baturayo, line 377: why we don't use default_casify_function in data_source.py? Can we handle the casing in soda-core instead of tests?

soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -345,11 +463,16 @@ def test_profile_columns_capitalized(data_source_fixture: DataSourceFixture):
 
     test_data_source = data_source_fixture.data_source_name
 
-    lowercase_column_name_databases = ["postgres", "redshift", "athena"]
-    if test_data_source in lowercase_column_name_databases:
-        expected_column_name = "items_sold"
+    if test_data_source in LOWERCASE_COLUMN_NAME_DATABASES:
+        expected_column_name1 = "items_sold"
```
baturayo, line 467: I would call `default_casify_function` here as well. Smth like;

`expected_col_name1 = self.default_casify_function("items_sold")`

baturayo (approved):

bastienboutonnet (approved):
Looks good to me. 

PR #1747, opened 2022-12-30
feat: add row count to profiling

baturayo:


soda/core/tests/data_source/test_profile_columns.py
```diff
@@ -342,7 +343,7 @@ def test_profile_columns_capitalized(data_source_fixture: DataSourceFixture):
 
     column_profiles = profiling["columnProfiles"]
 
-    test_data_source = os.environ.get("test_data_source")
+    test_data_source = data_source_fixture.data_source_name
```
baturayo, line 346: irrelevant with PR but it was failing for me since I didn't set any `test_data_source` env variable. The modified version is more robust for not set env variables. (if test_data_source env var is null it means postgres)

bastienboutonnet (approved):

PR #1756, opened 2023-01-09
add dataset filter test case starting with %

baturayo:
Added an extra unit test to see if the following use case works for all data sources;

```yaml
for each dataset D:
  datasets:
    - include %rder  # it shouldn't match with any dataset since it doesn't have % sign in the end
```


bastienboutonnet (approved):
Looks good to me.

vijaykiran (approved):

vijaykiran (approved):

PR #1775, opened 2023-01-25
refactor: profiler implementation

baturayo:
The purpose of this PR is to refactor numeric and text column profiler implementation. The new approach has more modular and easy-to-maintain code. Lots of nested if/else and for loop logic have been refactored.

Apart from unit tests, to make sure that there is nothing changed in the cloud, I also made e2e tests, and the screenshots are attached. 

**Before Refactor**
before_refactor

**After Refactor**
after_refactor


bastienboutonnet (commented):
Good work here, I think this will make the code more maintainable and clear indeed. Most of all, I appreciate a lot the fact that we have a few more in-depth tests. Which brings me to the following request:
- I assume that the in-depth tests were added after the refactor were done since we didn't have much in there.
- The consequence is that, while they will help with future refactors and work, they probably won't have helped us making sure that the refactoring didn't change the outcome of profiling.

I think there's a simple solution:
- could you run profiling on a few of the tables of adventureworks with the unrefactored version of the profiling logic and one with the current changed on this branch.
- check the numbers match and drop a few screenshots in the PR description maybe?

baturayo:
> * The consequence is that, while they will help with future refactors and work, they probably won't have helped us making sure that the refactoring didn't change the outcome of profiling.

Actually, that was my main goal to add this test. In the previous code, some fields weren't tested. So, I added some extra checks that the outcome didn't change with my refactor.

> I think there's a simple solution:
> could you run profiling on a few of the tables of adventureworks with the unrefactored version of the profiling logic and > one with the current changed on this branch.
> check the numbers match and drop a few screenshots in the PR description maybe?

Definetely, will do it now

bastienboutonnet (approved):

tituskx (approved):
altın bok 😍

PR #1778, opened 2023-01-26
Core 248 generic check

m1n0:


soda/core/soda/execution/check/check.py
```diff
@@ -236,6 +235,7 @@ def get_cloud_dict(self):
             "location": self.check_cfg.location.get_cloud_dict(),
             "dataSource": self.data_source_scan.data_source.data_source_name,
             "table": Partition.get_table_name(self.partition),
+            "datasets": [Partition.get_table_name(self.partition)],
```
milanaleksic, line 238: This property `datasets` is not supported by backend API is not supported by backend and it will be ignored

soda/core/soda/execution/check/anomaly_metric_check.py
```diff
@@ -57,7 +57,6 @@ def __init__(
                 check_identity=self.create_identity(), limit=3
             )
             self.diagnostics = {}
-            self.cloud_check_type = "anomalyDetection"
```
milanaleksic, line 60: Anomaly detection check has different properties than a generic check, and just changing the type doesn't actually allow properties like `anomalyErrorSeverity` to get into our Backend.

soda/core/soda/execution/check/schema_check.py
```diff
@@ -191,33 +190,123 @@ def has_schema_violations(
...
+                {
+                    "type": "csv",
+                    "text": f"Column,Event,Details\n{change_events_text}",
+                    "title": "Diagnostics",
+                }
+            )
+
+        schema_diagnostics["preferredChart"] = "bars"
+        schema_diagnostics["valueLabel"] = f"{event_count} schema event(s)"
+        schema_diagnostics["valueSeries"] = {
```
milanaleksic, line 295: Warning: we are limiting `valueSeries` to 10 items max per test result

soda/core/soda/execution/check/change_over_time_metric_check.py
```diff
@@ -59,6 +59,7 @@ def evaluate(self, metrics: Dict[str, Metric], historic_values: Dict[str, object
             self.historic_diff_values = {
                 "historic_value": historic_value,
                 "metric_value": metric_value,
+                "delta": self.check_value,
```
milanaleksic, line 62: There is an inconsistency in properties you are pushing here compared to the published API of `ChangeOverTimeCoreCheckDiagnostics`. Only `delta` is correct. We have `historicAggregate` and `value`.
But, I think you are pushing this as a `generic` check now, yes? Therefore, all these properties will be lost

milanaleksic (commented):
I have taken a look and found API inconsistencies. I think you moved all the checks to use `generic`, but you haven't actually changed the RPC objects, so you still expect properties as in use-case specific classes like `ChangeOverTimeCoreCheckDiagnostics` and `AnomalyDetectionCoreCheckDiagnostics`, so data will be lost in those special cases. It could be this is still work in progress, perhaps?

Current API is always available on our API page https://cloud.soda.io/docs/api/index.html#/soda-core/sodaCoreInsertScanResults

soda/core/soda/execution/check/schema_check.py
```diff
@@ -191,33 +190,123 @@ def has_schema_violations(
...
+        event_count = len(change_events)
+
+        if change_events:
+            change_events_text = "\n".join(
+                f"{ch['column']},:icon-{value_type}: {ch['event']}, {ch['details']}" for ch in change_events
+            )
+            schema_diagnostics["blocks"].append(
+                {
+                    "type": "csv",
+                    "text": f"Column,Event,Details\n{change_events_text}",
```
milanaleksic, line 288: You will have to truncate this text to 2048 characters (or backend will do it for you). Alternatively, when you go over 2048 characters go for a `file` block instead.

soda/core/soda/execution/check/anomaly_metric_check.py
m1n0: removing anomaly checks from this PR, this will be dealt with separately

soda/core/soda/execution/check/change_over_time_metric_check.py
m1n0: removing CoT checks from this PR, this will be dealt with separately

soda/core/soda/execution/check/check.py
m1n0: removed

soda/core/soda/execution/check/schema_check.py
m1n0: removing schema checks from this PR, this will be dealt with separately

soda/core/soda/execution/check/schema_check.py
m1n0: removing schema checks from this PR, this will be dealt with separately

vijaykiran (approved):

PR #1780, opened 2023-02-01
fix: confidence bounds of anomaly detection sit too close together

tituskx:
This PR resolves: [anomaly score sends alerts when actual value is VERY close to predicted value](https://sodadata.atlassian.net/browse/CLOUD-2703)

This PR implements the following:

* Instead of rounding the uncertainty intervals' lower and upper bounds to ten decimals, they are rounded using `np.floor` and `np.ceil` respectively if the metric that we're classifying is an integer. Running the original time series that was causing issues now no longer causes any issues as can be seen on [dev](https://dev.sodadata.io/checks/c9457436-d78f-4ea4-bca6-27b718838614/metric-result?testResultId=6b9a7221-6117-4165-a173-85e9f476572b). The original check measurements that were used to create that plot can be found through [datadog](https://app.datadoghq.com/logs?query=host%3A10.0.5.176%20service%3Ashape-server-cloud%20%40aws.awslogs.logGroup%3Ashape-server-cloud&agg_m=count&agg_q=service&agg_t=count&cols=host%2Cservice&context_event=AYXJ9S5HAADNs5udqVYUIwPE&event=AgAAAYXJ9Ok1nWIKLgAAAAAAAAAYAAAAAEFZWEo5UzVIQUFETnM1dWRxVllVSXdPMgAAACQAAAAAMDE4NWNhOTgtMWFlOS00NTQzLWI0ZjctYjQxMzAzM2FmZTk2&index=&messageDisplay=inline&stream_sort=time%2Cdesc&to_event=AgAAAYXJ9OxtnWIKRgAAAAAAAAAYAAAAAEFZWEo5UzVIQUFETnM1dWRxVllVSXdQTwAAACQAAAAAMDE4NWNhOTgtMWFlOS00NTQzLWI0ZjctYjQxMzAzM2FmZTk2&viz=&from_ts=1674130266727&to_ts=1674130626727&live=false)

**Notes** 

This is not the final fix. Inspecting the response that was sent by the backend [for this dataset](https://dev.sodadata.io/checks/c9457436-d78f-4ea4-bca6-27b718838614/metric-result?testResultId=6b9a7221-6117-4165-a173-85e9f476572b) shows that the uncertainty bounds are too close together. This behaviour is caused by the fact that we use the new measurement (the value that we are classifying) to [fit the model](https://github.com/sodadata/soda-core/blob/a52b8b0f96d91c4639179fa8d58a8c7ab0558334/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py#L339). This means that the estimates of the prediction errors are way too conservative. I.e. doing something like 
```
m = Prophet()
m.fit(time_series)
predictions = m.predict(time_series)
```
means that the prediction intervals are extremely narrow and sit unrealistically close to the `real_data` column. If we would actually use an actual forecast, i.e. something like
```
m = Prophet()
m.fit(time_series[:-1])
future = m.make_future_dataframe(periods=1)
forecast = m.predict(future)
```
the prediction intervals around the training data used to fit the model are still very small, but the one used for the prediction are much wider (and would look a lot more realistic).

Using data that we're trying to predict to fit the model leads to unpredictable and strange behaviour but also means that it is too insensitive. Consider the following example where we're classifying the new datapoint for 20220-12-08 (records are in ascending order here)
Example 1 Incorrect
Going from an increase of 33.33% to 1100% in the number of rows doesn't trigger an alert because the datapoint that we're trying to classify is used to fit the model.

@bastienboutonnet I did not change the in-sample prediction to actual forecast in this PR, because the anomaly algorithm would become more sensitive and requires additional testing. If you want though, I'd be happy to add it in this PR as I think it's an improvement, but it would require more time.

### DIFF
--- a/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
+++ b/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
@@ -129,11 +129,15 @@ class PreprocessError(Exception):
 class ProphetDetector(BaseDetector):
     """ProphetDetector."""
 
+    # TODO: [CLOUD-2823] Deprecate use of class variable to indicate integer type metrics
+    integer_type_metrics = ["row_count", "duplicate_count", "missing_count", "invalid_count"]
+
     def __init__(
         self,
         logs: Logs,
         params: Dict[str, Any],
         time_series_data: pd.DataFrame,
+        metric_name: str,
         has_exegonenous_regressor: bool = False,
     ) -> None:
         """Constructor for ProphetDetector
@@ -166,6 +170,7 @@ class ProphetDetector(BaseDetector):
         self._is_trained: bool = False
         self._has_exogenous_regressor = has_exegonenous_regressor
         self.time_series_data = time_series_data  # this gets potentially rewritten when runnin skip measurements
+        self.uncertainty_bounds_require_integer_rounding: bool = metric_name in self.integer_type_metrics
 
         # public attrs
         self.model: Prophet
@@ -373,11 +378,18 @@ class ProphetDetector(BaseDetector):
         self.predictions["real_data"] = self.time_series["y"].reset_index(drop=True)
 
         self._logs.debug(f"Anomaly Detection: detecting anomalies for the last {self._n_points} points.")
-
         # round/trucate because floats are shit and cause precision errors
-        self.predictions["real_data"] = self.predictions["real_data"].round(10)
-        self.predictions["yhat_lower"] = self.predictions["yhat_lower"].round(10)
-        self.predictions["yhat_upper"] = self.predictions["yhat_upper"].round(10)
+        latest_measurement = self.predictions.iloc[-1]
+        no_integer_between_bounds = (
+            np.abs(np.ceil(latest_measurement["yhat_upper"]) - np.floor(latest_measurement["yhat_lower"])) < 2
+        )
+        if no_integer_between_bounds and self.uncertainty_bounds_require_integer_rounding:
+            self.predictions["yhat_lower"] = np.floor(self.predictions["yhat_lower"])
+            self.predictions["yhat_upper"] = np.ceil(self.predictions["yhat_upper"])
+        else:
+            self.predictions["real_data"] = self.predictions["real_data"].round(10)
+            self.predictions["yhat_lower"] = self.predictions["yhat_lower"].round(10)
+            self.predictions["yhat_upper"] = self.predictions["yhat_upper"].round(10)
 
         # flag data points that fall out of confidence bounds
         self.predictions["is_anomaly"] = 0
--- a/soda/core/soda/execution/check/anomaly_metric_check.py
+++ b/soda/core/soda/execution/check/anomaly_metric_check.py
@@ -111,7 +111,7 @@ class AnomalyMetricCheck(MetricCheck):
             self.logs.error(f"{SODA_SCIENTIFIC_MISSING_LOG_MESSAGE}\n Original error: {e}")
             return
 
-        anomaly_detector = AnomalyDetector(historic_measurements, historic_check_results, self.logs)
+        anomaly_detector = AnomalyDetector(historic_measurements, historic_check_results, self.logs, metric_name)
         level, diagnostics = anomaly_detector.evaluate()
         assert isinstance(diagnostics, dict), f"Anomaly diagnostics should be a dict. Got a {type(diagnostics)} instead"
 
--- a/soda/core/tests/data_source/test_profile_columns.py
+++ b/soda/core/tests/data_source/test_profile_columns.py
@@ -101,7 +101,7 @@ def test_profile_columns_numeric(data_source_fixture: DataSourceFixture):
     assert histogram["boundaries"] == [0.5, 3.3, 6.1]
 
     # TODO: Fix the histogram issue for mysql refer to CLOUD-2763
-    if data_source_fixture.data_source_name == "mysql":
+    if data_source_fixture.data_source_name in ["mysql", "dask"]:
         assert histogram["frequencies"] == [4, 2, 0]
     else:
         assert histogram["frequencies"] == [4, 0, 2]
--- a/soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
+++ b/soda/scientific/soda/scientific/anomaly_detection/anomaly_detector.py
@@ -87,8 +87,9 @@ class AnomalyHistoricalMeasurements(BaseModel):
 
 
 class AnomalyDetector:
-    def __init__(self, measurements, check_results, logs: Logs):
+    def __init__(self, measurements, check_results, logs: Logs, metric_name: str):
         self._logs = logs
+        self.metric_name = metric_name
         self.df_measurements = self._parse_historical_measurements(measurements)
         self.df_check_results = self._parse_historical_check_results(check_results)
         self.params = self._parse_params()
@@ -103,6 +104,7 @@ class AnomalyDetector:
             logs=self._logs,
             params=self.params,
             time_series_data=feedback.df_feedback_processed,
+            metric_name=self.metric_name,
             has_exegonenous_regressor=feedback.has_exegonenous_regressor,
         )
         df_anomalies = detector.run()
--- a/soda/scientific/tests/anomaly_detection/anomaly_detector_test.py
+++ b/soda/scientific/tests/anomaly_detection/anomaly_detector_test.py
@@ -127,7 +127,7 @@ LOGS = Logs(logging.getLogger(__name__))
 def test_anomaly_detector_evaluate(historical_measurements, historical_check_results, expectation):
     from soda.scientific.anomaly_detection.anomaly_detector import AnomalyDetector
 
-    detector = AnomalyDetector(historical_measurements, historical_check_results, logs=LOGS)
+    detector = AnomalyDetector(historical_measurements, historical_check_results, logs=LOGS, metric_name="avg_length")
     _, diagnostic = detector.evaluate()
     assert diagnostic["value"] == expectation["value"]
     assert diagnostic["anomalyProbability"] == pytest.approx(expectation["anomalyProbability"])


soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
```diff
@@ -129,11 +129,14 @@ class PreprocessError(Exception):
 class ProphetDetector(BaseDetector):
     """ProphetDetector."""
 
+    integer_type_metrics = ["row_count", "duplicate_count", "missing_count", "invalid_count"]
```
bastienboutonnet: I'm a little curious about the reasoning behind this decision. Can you give a few pointers as to why we should only perform the rounding in the new way to integer metrics rather than to all?

bastienboutonnet (commented):
Stellar investigation of the issue and overall, something that looks like a good proposal. 

I have a question to understand one of the fundamental choices you made. I'm curious about the assumption.

More generally, thanks a lot for the additional analysis and thoughts on the current approach and the shortcomings of including the last point in the prediction.

This point came up a few times already and my thoughts on this are very much of a resounding "let's definitely investigate this". As I think I've mentioned already to both you and @baturayo the current implementation was a conscious choice. One that I saw on my testing a while back would be appropriate in order to be on the side of caution.

However, it seems in some cases, especially when data is really stable that this almost leads to the opposite effect and makes the model too sensitive if I understand your analysis correctly. That's **very** interesting.

Which, I think, contrary to what you say in the description, this PR **is** the final fix for this issue, we should nonetheless investigate revising the assumption this model has been running under. 

Can you create a ticket where you outline what you wrote in the PR description? If you have some code already or anything that would make it easier for us to jump into the issue again, do add it to our research repo somewhere. If not, a reference to the PR + your write up and any additional leads would be great.

If we are to change the approach, I want to make sure we go through it methodically, compare the performance, understand the compromises etc. But I'm super up for it!

soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
tituskx: Rounding like this completely messes up the uncertainty interval, so I don't think it's something we should do unless necessary. This type of aggressive rounding could hurt us when we're predicting floats.

Thinking about I'll make a small adjustment for integers as well and only round if there's no integer between the upper and lower bound.

tituskx:
> Which, I think, contrary to what you say in the description, this PR is the final fix for this issue, we should nonetheless investigate revising the assumption this model has been running under.

@bastienboutonnet to be honest I disagree because the problem lies with the way we construct uncertainty intervals around our predictions (this PR is just an ad-hoc solution to a bigger problem). In any case I agree that we can tackle the big problem in another ticket 👍 . I've added an issue to next sprint [Re-evaluate anomaly detection algorithm and remove future datapoints from current fitting dataset](https://sodadata.atlassian.net/browse/CLOUD-2795).

tituskx:
@baturayo I reimplemented `integer_type_metrics = ["row_count", "duplicate_count", "missing_count", "invalid_count"]` part instead of checking whether the metric value is an integer. The reason is that we interpolate between dates by taking the mean so that that `row_count` can actually be a float. Once we deal with [Re-evaluate anomaly detection algorithm and remove future datapoints from current fitting dataset](https://sodadata.atlassian.net/browse/CLOUD-2795) this type of rounding will no longer be necessary and we can deprecate the logic that you did not like.

soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
baturayo: I would add a TODO with the issue link to not forget to delete the hardcoded part.

bastienboutonnet: Just so you know, that didn't really answer my question. In other words why is it only necessary with ints? Why is it safer than with floats?

baturayo (approved):

bastienboutonnet:
> to be honest I disagree because the problem lies with the way we construct uncertainty intervals around our predictions (this PR is just an ad-hoc solution to a bigger problem).

@tituskx I think not including the last measurement, which is a new feature and new prediction approach, could still run in a situation where by chance we have those values close to each other. It certainly makes it less likely but it's not a guarantee right?

In that case, we'll always have to think of a corner-case fix. 

Am I missing something?

tituskx:
> @tituskx I think not including the last measurement, which is a new feature and new prediction approach, could still run in a situation where by chance we have those values close to each other. It certainly makes it less likely but it's not a guarantee right?
> 
> In that case, we'll always have to think of a corner-case fix.

I think it would extremely unlikely for the bounds to be close together and not include an integer, but let's see!

> Am I missing something?

Maybe yes so let me address it here and also respond to the comment that you made in [Re-evaluate anomaly detection algorithm and remove future datapoints from current fitting dataset](https://sodadata.atlassian.net/browse/CLOUD-2795?focusedCommentId=20968) at the same time.

Why the results that I showed above should worry us is that there seems to be a fundamental difference between the way the confidence intervals are constructed for future datapoint vs historical datapoints. The "prediction intervals" for the historical data sit way too close together. 

Therefore, it seems that we're not just adding a datapoint to our training data, thereby risking overfitting. What actually seems to happen is that the model removes some uncertainty from the prediction intervals for the historical data, that it normally adds when it's forecasting. I.e. it seems to recognise that historical data (maybe based on the timestamp) was used to fit the model and therefore should be treated differently. I assume that you would agree that that's not what we want right? 

To make sure that we're on the same page; it's not like we're using the same dataset to train and evaluate an image classifier. What seems to happen is that there's something **fundamentally different** in the **way** the model creates prediction intervals when it concerns historical data.

soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
tituskx: It's necessary for row count because we need a way to interpret the prediction interval when it doesn't contain any integers. I.e. the confidence interval [10.5, 10.6] will classify any next observation as an anomaly if the next observation is an integer. So we need to round to [10, 11] (or do something else) to have a working implementation.

For floats that is not the case. Whatever your prediction interval looks like there's always a next observation that can be contained by it.

Also imagine that you have the values 14.2, 10.55, 10.65, 10.2, 10.35, 15.2, 10.2, 10.9, 11.1 leading to a prediction interval of [10.5, 10.6] (didn't do the math just an example). If you round downwards and upwards you get the confidence interval [10, 11] which is not sensitive to the small difference in magnitude of the previous values.

bastienboutonnet:
> I think it would extremely unlikely for the bounds to be close together and not include an integer, but let's see!

But not impossible. That's an important nuance and is unrelated to whether you think the current approach is fundamentally flawed or not. I want to make sure we don't pile things on top of each other.

> Why the results that I showed above should worry us is that there seems to be a fundamental difference between the way the confidence intervals are constructed for future datapoint vs historical datapoints. The "prediction intervals" for the historical data sit way too close together.

> Therefore, it seems that we're not just adding a datapoint to our training data, thereby risking overfitting. What actually seems to happen is that the model removes some uncertainty from the prediction intervals for the historical data, that it normally adds when it's forecasting. I.e. it seems to recognise which historical data (maybe based on the timestamp) was used to fit the model and therefore should be treated differently. I assume that you would agree that that's not what we want right?

Well, yes, that's likely to happen when data doesn't move much. And it was, again, by design to make it less sensitive. My assumption, and I already communicated it several times but no worries, is that I explicitly wanted the last point to be in the fitting.

What my assumption is here is that I don't care about the model forecasting. I care that it tells me how certain it is of any given point. 

However, I admit that when data does not change at all, it becomes quickly too confident and that leads to the problem we have here. But that doesn't make this approach bad fundamentally. It makes it work as expected. 

Could we have seen this coming, maybe, but in the testing I did before this didn't come up. And that is fine, this case is actually kind of a corner case, and as with everything in software or data corner-cases are a thing we cannot avoid.


> What actually seems to happen is that the model removes some uncertainty from the prediction intervals for the historical data, that it normally adds when it's forecasting. I.e. it seems to recognise which historical data (maybe based on the timestamp) was used to fit the model and therefore should be treated differently. I assume that you would agree that that's not what we want right?

I wonder how that can be. We ask of the model to always fit and predict the whole timeframe. Therefore I don't think that can happen. But I'm curious to expand on why you think that. I would however encourage to take this discussion internally as this is out of the scope of this PR.

What I want to clarify, is that there is nothing wrong with the current approach. It works as expected given the assumptions that were thrown into the design. Now, in light of this evidence, we may decide that we want to approach it differently. I already conceded to wanting to do so, but we will have to do due diligence and testing.

soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
bastienboutonnet: Ok that all makes sense now. Thanks for the explanation. So for integers I think this is totally legit. 

That being said, I'm not sure I love the idea of listing the metrics here. We'll always fall behind + it mixes knowledge from another domain (core in that case). Do we have a way to simply check whether the actual measurements time series is made of integers. I forgot if that would be the case but if so, I'd prefer that we go with that. If we must rely on the metric, I think that actually we will want to work with core so that they can communicate about the metric's type. I think it should be ok to get this from them and have cloud be ok to store it. In fact, it could help for a lot of other things. More than happy to look into putting up this series of feature requests with the teams if we deem it necessary.

Right now, though, if listing the metrics is our ownly option. Please add a `# TODO` comment that refers to a JIRA issue that needs to be created to refactor this when/if cloud and core agree to support datatype info storage.

tituskx: @bastienboutonnet No I don’t think so I’m afraid, also see my [comment to Baturay](https://github.com/sodadata/soda-core/pull/1780#issuecomment-1415481250). The reason is that we use interpolation between values so row counts can be floats. Looking at both the historical check results and historic measurements there seems to be no field that indicates type.

bastienboutonnet (approved):
Let's go!

vijaykiran:
@tituskx  Can you please squash and update the commit message with the description? I thin the description in the PR is valuable and should be in the commit log as well.

PR #1789, opened 2023-02-14
Generic check type for Schema check

m1n0:


vijaykiran:
@m1n0  Is this ready ?

m1n0:
This is now ready for review, its a significant improvement of what is sent to Cloud to be rendered, both warn and fail info is sent and also displayed in the cli output. The only change from the outside pov is changed api of schema check, schema changes are captured in an encapsulated object instead of globally - this is so that both warn and fail events can be captured. Please merge if all looks good @vijaykiran

vijaykiran (approved):

PR #1791, opened 2023-02-14
pin duckdb version

baturayo:


vijaykiran (approved):

JorgeGarciaIrazabal:
@baturayo what was the reason to pin duckdb? duckdb 0.7 has some nice features regarding to schema names and registered datasets.

PR #1793, opened 2023-02-14
Invalid values/format/regex for validity checks

m1n0:


vijaykiran (approved):

vijaykiran:
@m1n0 looks like SQLServer doesn't like the changes - can you fix it?

PR #1796, opened 2023-02-15
Data source utils: support config as string

m1n0:


soda/core/soda/core/api/data_source_utils.py
```diff
@@ -6,14 +8,20 @@ class DataSourceUtils:
     EXPERIMENTAL: Utility API for DataSource and connections.
     """
 
-    def __init__(self, config_path: str):
+    def __init__(self, config_path: str | None = None, config_string: str | None = None):
         """
         All the features for config files (e.g. env vars) are supported.
         :param config_path: The director path containing the data source config yaml files.
```
vijaykiran, line 14: Can you update the function doc string too?

vijaykiran (changes requested):

soda/core/soda/core/api/data_source_utils.py
m1n0: 🤦 updated

vijaykiran (approved):

PR #1799, opened 2023-02-16
fix: skip/ignore measurements does not work

tituskx:
This PR resolves: [skip/ignore measurements does not seem to work](https://sodadata.atlassian.net/browse/CLOUD-2787)

Specifically, this PR fixes the following bugs:

1. Skip measurements are not used because the feedback provided in the check results uses camel case, i.e., `skipMeasurements`, while the `BaseDetector` looks for the snake case `skip_measurements`
2. When the skip measurements are used by the `BaseDetector`, future measurements are skipped instead of previous measurements ("ignore previous values" actually meant "ignore future values"). If bug 1. would not have been present, using "ignore previous values" would have shut down anomaly detection.

**Notes**

* This PR adds a new unit test called `test_prophet_model_skip_measurements` that ensures that skipping measurements works
* The new code was also tested on dev, see [this dataset](https://dev.sodadata.io/checks/7bbc31df-41aa-4ce1-9364-f24e4c9c286c/metric-result?testResultId=890d1a75-989d-4906-ac87-df6a071a551b)
* @bastienboutonnet during the sprint planning we discussed adding functionality to automatically change the camel case to snake case in the pydantic models, like we do in rca. However, since camel case is used in lots of other places during anomaly detection (not just in the `skip_measurements` method) I think it makes more sense to tackle them all in another ticket.


soda/scientific/tests/anomaly_detection/anomaly_detector_test.py
```diff
@@ -132,3 +152,176 @@ def test_anomaly_detector_evaluate(historical_measurements, historical_check_res
...
+                    0: "2023-02-15 11:00:00",
+                    1: "2023-02-16 11:00:00",
+                },
+                "skipMeasurements": {0: None, 1: nan},
+            },
+            id="previousAndThis",
+        ),
+    ],
+)
+def test_prophet_model_skip_measurements(
```
baturayo, line 310: It's great that you added this test 👍

soda/scientific/tests/anomaly_detection/anomaly_detector_test.py
```diff
@@ -132,3 +152,176 @@ def test_anomaly_detector_evaluate(historical_measurements, historical_check_res
...
+            "warning_lower_than_or_equal": "warning_lower_than_or_equal",
+            "warning_greater_than_or_equal": "warning_greater_than_or_equal",
+            "critical_lower_than_or_equal": "critical_lower_than_or_equal",
+            "critical_greater_than_or_equal": "critical_greater_than_or_equal",
+        }
+    },
+}
+
+
+@pytest.mark.parametrize(
```
baturayo, line 210: I would create an anomaly_detection_assets.py folder to get the static dictionaries from there to make it more organized.

soda/scientific/soda/scientific/anomaly_detection/models/base.py
```diff
@@ -56,40 +56,30 @@ def _eliminate_measurements(
...
-        has_two_previouses = False
         is_inclusive = False
         max_index = []
-        if "skip_measurements" in self._data_df.columns:
+        if "skipMeasurements" in self._data_df.columns:
+            self._data_df = self._data_df.sort_values(by="ds", ascending=True)
+            self._data_df = self._data_df.reset_index(drop=True)
             skip_previous_inclusive_indexes = self._data_df.index[
-                self._data_df["skip_measurements"] == "previousAndThis"
+                self._data_df["skipMeasurements"] == "previousAndThis"
```
baturayo, line 65: instead of changing `skip_measurements` to `skipMeasurements` I would do the contrary. It's more pythonic.

soda/scientific/tests/anomaly_detection/anomaly_detector_test.py
tituskx: Yes makes sense. 👍  Shall we do that when we start refactoring/creating unit tests for anomaly detection? I.e. this ticket: https://sodadata.atlassian.net/browse/CLOUD-2804

soda/scientific/soda/scientific/anomaly_detection/models/base.py
tituskx: @baturayo yes totally agree! However, see this note in the PR description:

> during the sprint planning we discussed adding functionality to automatically change the camel case to snake case in the pydantic models, like we do in rca. However, since camel case is used in lots of other places during anomaly detection (not just in the skip_measurements method) I think it makes more sense to tackle them all in another ticket.

I think it makes sense to tackle all the camelcase stuff in another ticket

baturayo: indeed makes sense to tackle it in another ticket.

baturayo (approved):
Nice cleanups😍 LGTM!

bastienboutonnet (approved):

vijaykiran (approved):

PR #1802, opened 2023-02-16
fix: return yhat not trend as anomalyPredictedValue

tituskx:
Hey @bastienboutonnet came across this just now. Guess this is incorrect or am I mistaken?


bastienboutonnet (approved):
Looks good. Dunno how we came to that must have been a typo. Good think is that we weren't really using or showing any of this in the output at this stage but good catch.

Issue #1800, opened 2023-02-16
Wrong failed query on numeric query


JorgeGarciaIrazabal:
Soda report a wrong query to retrieve the failed records over a numeric query.

How to reproduce:
```
import duckdb
import pandas as pd
from soda.sampler.sampler import Sampler
from soda.scan import Scan

duckdb_conn = duckdb.connect(database=":memory:")


class QuerySampler(Sampler):
    def store_sample(self, sample_context):
        print("query:", sample_context.query)
        output = duckdb_conn.execute(sample_context.query).df()
        print("values:", set(output.value))


df = pd.DataFrame({"value": [1, 2, 3, 4, 5, 6]})
duckdb_conn.register("my_table", df)

scan = Scan()
scan.add_configuration_yaml_str(
    f"""
data_source duckdb_db:
  type: duckdb
  database: duckdb
"""
)


scan.set_data_source_name("duckdb_db")
scan.add_duckdb_connection(duckdb_conn, data_source_name="duckdb_db")
scan.add_sodacl_yaml_str(
    """
checks for my_table:
- invalid_count(value) = 0:
   valid max: 4
   valid min: 2
   """
)
scan.sampler = QuerySampler()
scan.execute()

```

Current output (from `QuerySampler`)
```
SELECT * FROM my_table 
 WHERE NOT value IS NULL AND NOT value >= 2.0 AND value <= 4.0 
 LIMIT 100
values: {1}

```

Expected output:

```
query: SELECT * FROM my_table 
 WHERE NOT value IS NULL AND NOT (value >= 4.0 AND value <= 3.0)
 LIMIT 100
values: {1, 5, 6}
```

I was able to reproduce this with Postgressql too

jmarien:
SODA-1441

JorgeGarciaIrazabal:
I think the problem is here: https://github.com/sodadata/soda-core/blob/main/soda/core/soda/execution/metric/numeric_query_metric.py#L312

I can create a PR today if that helps.

JorgeGarciaIrazabal:
fixed on #1811 closing this issue

PR #1804, opened 2023-02-17
fix fail query generation when multiple conditions exist

JorgeGarciaIrazabal:
wrap missing condition, valid_condition and invalid_condition when negating with "NOT"

This solves #1800


soda/core/tests/data_source/test_samples.py
```diff
@@ -327,3 +327,39 @@ def test_sample_limit_default(check: str, has_sample_query: bool, data_source_fi
 
         limit_keyword = data_source_fixture.data_source.LIMIT_KEYWORD
         assert f"{limit_keyword} {DEFAULT_FAILED_ROWS_SAMPLE_LIMIT}" in sample_queries[0]
+
+
+def test_sample_with_multiple_value_condition(data_source_fixture: DataSourceFixture):
+    """
+    Tests failed rows queries.
+    Tests query returns failed query includin all not in valid values.
```
JorgeGarciaIrazabal, line 335:
```suggestion
    Tests query returns failed query including all not in valid values.
```

vijaykiran:
See https://github.com/sodadata/soda-core/pull/1811

PR #1807, opened 2023-02-20
fix: do not return `anomalyProbability`

tituskx:
This PR resolves: [Anomaly detection, anomaly probability exceeds 1](https://sodadata.atlassian.net/browse/CLOUD-2802)

* This PR removes the `anomalyProbability` field from the anomaly detection output (it's not used anywhere except for reporting purposes)
* This [dataset on dev](https://dev.sodadata.io/checks/a3460ae9-3e37-4a1c-890b-666eac7804a2/metric-result?testResultId=f8637c4d-62a0-4fd2-9336-d031748be5cb) was used to test that the anomaly detection module still works correctly without the `anomalyProbability` field (also tested that the module still sends out alerts and it does)

@bastienboutonnet I have added some questions and comments here: [Anomaly detection, anomaly probability exceeds 1](https://sodadata.atlassian.net/browse/CLOUD-2802)

### DIFF
--- a/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
+++ b/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
@@ -349,30 +349,6 @@ class ProphetDetector(BaseDetector):
         self.predictions = self.model.predict(self.time_series)
         self._is_trained = True
 
-    @staticmethod
-    def _derive_anomaly_probability(
-        predictions: pd.DataFrame,
-        anomaly_flag_col: str = "is_anomaly",
-        real_data_colname: str = "real_data",
-    ) -> pd.DataFrame:
-        predictions["anomaly_probability"] = 0
-        anomaly_directions = {
-            "above": {anomaly_flag_col: 1, "confidence_col_name": "yhat_upper"},
-            "below": {anomaly_flag_col: -1, "confidence_col_name": "yhat_lower"},
-        }
-
-        interval_range = predictions["yhat_upper"] - predictions["yhat_lower"]
-        for _, params in anomaly_directions.items():
-            # TODO: We might want to revisit the maths here. Might make sense to sigmoidize around
-            # a dynamic acceptance threshold.
-            predictions.loc[predictions[anomaly_flag_col] == params[anomaly_flag_col], "anomaly_probability"] = abs(
-                (predictions[real_data_colname] - predictions[params["confidence_col_name"]]) / interval_range
-            )
-        # TODO: Add a failsafe for a case where we might still have an inf value
-        # this means we'll check if any rows has inf and then if so, we replave to prob 0
-        # is_anomaly should then also be 0 etc. Test thoroughly.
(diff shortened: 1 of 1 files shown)


soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
```diff
@@ -396,12 +396,13 @@ def detect_anomalies(self):
         self.predictions.loc[self.predictions["real_data"] > self.predictions["yhat_upper"], "is_anomaly"] = 1
         self.predictions.loc[self.predictions["real_data"] < self.predictions["yhat_lower"], "is_anomaly"] = -1
 
+        # TODO: [CLOUD-2990] re-enable once we compute uncertainty intervals correctly
         # derive anomaly probability (by normalising delta between real and predicted)
-        self._derive_anomaly_probability(self.predictions, anomaly_flag_col="is_anomaly", real_data_colname="real_data")
+        # self._derive_anomaly_probability(self.predictions, anomaly_flag_col="is_anomaly", real_data_colname="real_data")
```
bastienboutonnet, line 401: are we sure the backend doesn't use that field to determine whether something is anomalous or not?

Are we sure **we** don't use it in the feedback?

bastienboutonnet (commented):
fundamental question about whether it's actually safe to deprecate

soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py
tituskx: hey @bastienboutonnet 

> are we sure the backend doesn't use that field to determine whether something is anomalous or not?

Backend works without this field, see this [dataset on dev](https://dev.sodadata.io/checks/434b239d-da81-4262-936c-973d6495f4ee/metric-result?testResultId=badd6a54-1eaf-472b-a8fc-24ba5706776f) that was created running anomaly detection using the branch of the current PR

> Are we sure we don't use it in the feedback?

 Yes! In fact I'll remove the comment about the [gnarly conversion](https://github.com/sodadata/soda-core/blob/6a361a3eb84daf19eacbb75ef5b9518b6400cda7/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py#L416-L419) as well (I guess it refers to something that is no longer there?). As you can see in the [_criticality_threshold_calc code](https://github.com/sodadata/soda-core/blob/6a361a3eb84daf19eacbb75ef5b9518b6400cda7/soda/scientific/soda/scientific/anomaly_detection/models/prophet_model.py#L439-L450) we only use `y_upper` and `y_lower` together with the criticality threshold which [is hardcoded to 0.1](https://github.com/sodadata/soda-core/blob/6a361a3eb84daf19eacbb75ef5b9518b6400cda7/soda/scientific/soda/scientific/anomaly_detection/detector_config.yaml#L28)

bastienboutonnet (approved):
Good cleanup!

vijaykiran (approved):

PR #1811, opened 2023-02-22
Fix condition for fail queries 

vijaykiran:


JorgeGarciaIrazabal:
@vijaykiran I think most of the test errors are because some datasource connections don't have a `cursor`. Is there a way I can help with this issue?

m1n0 (approved):

m1n0:
thanks for this improvement @JorgeGarciaIrazabal ! I was able to fix the test by using already fetched data instead of re-running the sample queries with a cursor manually. Merging

JorgeGarciaIrazabal:
amazing, thank you so much @m1n0!

PR #1847, opened 2023-03-23
feat: derive bins and weights for categorical DRO in SQL warehouse

tituskx:
This PR resolves: [Rework categorical DRO generation to use SQL instead of in-memory python](https://sodadata.atlassian.net/browse/IA-153)

This PR

* Implements the use of a group by query in `update-dro` to get the counts of categorical values and deprecates the old `value_counts = data.value_counts()` logic
* Creates two new unit tests:
    * `test_cli_update_distribution_file_categorical`. The output of this unit test was generated using the main branch to ensure that the changes in this PR do not give different results
    * `test_cli_update_distribution_errors`. Checks whether we log the correct error message when the column just has null values 

Notes

* Enabling/creating unit tests for `update-dro` for data sources other than postgres seems to suggest that the functionality does not work for `dask`, `sqlserver`, `vertica`, and `spark` (the tests fail). I propose we handle that in another ticket as it does not have to do with the current changes (the tests fail on main too when enabled). If you agree I'll create a new ticket
* The results using the group by query are not ordered (contrary to the `value_counts = data.value_counts()` results). This means that generating from the DRO will lead to different results as the seed + the different order will lead to different results. While a previously failing check should still fail (unless it's p-value is close to the significance level), the p-value might be different. I think this implementation makes most sense because ordering the group by query is not necessary, based on the way [we sample categorical data](https://github.com/sodadata/soda-core/blob/08a5339e49edc4a416cb159a29e8743d82557916/soda/scientific/soda/scientific/distribution/utils.py#L77), and could slow down the query. @bastienboutonnet Let me know if you prefer to have the exact same results as on main, would be an easy change.


soda/core/soda/cli/cli.py
```diff
@@ -317,25 +317,33 @@ def update_dro(
...
+                column_values = [row[0] for row in rows]
 
             if not column_values:
                 logging.error(
                     f"""{column_name} column does not have any data! To generate a distribution reference object (DRO) your column needs to have more than 0 rows!"""
                 )
                 return
 
-            if all(i is None for i in column_values):
+            if (distribution_type == "categorical" and all(row[0] is None for row in column_values)) or (
```
bastienboutonnet, line 344: when creating long ifs, I like to load them into a variable with an easy to read name. For example
`is_categorical_and_blah blah = distribution_type == 'categorical' and blah_blah is True`

soda/core/tests/cli/test_cli_update_distribution_file.py
```diff
@@ -38,4 +43,104 @@ def test_cli_update_distribution_file(data_source_fixture: DataSourceFixture, mo
...
+    reason="Other data sources are not supported",
+)
+@pytest.mark.parametrize(
+    "distribution_type",
+    [
+        pytest.param("continuous", id="continuous distribution type"),
+        pytest.param("categorical", id="categorical distribution type"),
+    ],
+)
+def test_cli_update_distribution_errors(
```
baturayo, line 115: good one 👍

soda/scientific/soda/scientific/distribution/generate_dro.py
```diff
@@ -126,13 +125,10 @@ def generate_continuous_dro(self) -> DRO:
...
-        data = pd.Series(self.data)
-        value_counts = data.value_counts()
-
-        labels = value_counts.index.to_numpy()
-        weights = value_counts.to_numpy()
-        weights = normalize(weights)
-        return DRO(weights=weights.tolist(), bins=labels.tolist())
+        labels = [row[0] for row in self.data]
+        label_counts = [row[1] for row in self.data]
+        weights = [label_count / sum(label_counts) for label_count in label_counts]
```
baturayo, line 130: what if `sum(label_counts)=0`? Can it be the case?

soda/core/tests/cli/test_cli_update_distribution_file.py
```diff
@@ -2,15 +2,20 @@
...
 from helpers.fixtures import test_data_source
 from helpers.mock_file_system import MockFileSystem
+from ruamel.yaml import YAML
 
 
 @pytest.mark.skipif(
-    test_data_source != "postgres",
-    reason="Run for postgres only as nothing data source specific is tested.",
+    test_data_source not in ["postgres", "snowflake", "bigquery", "mysql"],
+    reason="Other data sources are not supported",
```
baturayo, line 18: why we don't support other data sources? We just do `select * from where or group by` right?

bastienboutonnet (commented):
Approach looks good to me @tituskx I have a few general questions though:

- The issue mentions that we should leave the python method for datasources in which we can't perform the SQL approach. Did you miss that perhaps?
  - I think in theory it wouldn't hurt (see my last bullet)
- On the ordering, if you think it definitely has no impact then I would say no need to bother indeed.
- For the datasources you mention in the PR description, those are all experimental and community driven so indeed it's probably the case that they were never tested. However, no one thought that it might be a good idea to add a warning about db support. So I have a question:
  - is it the case that the old python approach also didn't work for these dbs? If "it didn't work" then we should throw a warning to day db is not supported. If it did work (or could) how about we leverage the old python code path?

I appreciate that the warning + mapping of supported data sources may be a little OOS. If I were you, I would consider implementing it as I think it's a simple dict "if not in this list say something" kind of thing. But if you'd like to move on and do it in another PR please make a ticket. I think we should be able to even complete it in this sprint given our current velocity.

soda/scientific/soda/scientific/distribution/generate_dro.py
tituskx: I believe they can never be zero ever or else this line would have thrown an error: [ if not column_values](https://github.com/sodadata/soda-core/pull/1847/files#diff-bb85d418b19f0af8eb3a93cbf5963e15aeaabb3bced924989cece5888979c7e4R338-R342)

soda/core/tests/cli/test_cli_update_distribution_file.py
tituskx: Yeah I don't know why actually, I'll have a look to see how hard it would be to fix!

soda/core/tests/cli/test_cli_update_distribution_file.py
```diff
@@ -38,4 +43,104 @@ def test_cli_update_distribution_file(data_source_fixture: DataSourceFixture, mo
...
+    "distribution_type",
+    [
+        pytest.param("continuous", id="continuous distribution type"),
+        pytest.param("categorical", id="categorical distribution type"),
+    ],
+)
+def test_cli_update_distribution_errors(
+    data_source_fixture: DataSourceFixture, mock_file_system: MockFileSystem, distribution_type
+):
+    table_name = data_source_fixture.ensure_test_table(null_test_table)
```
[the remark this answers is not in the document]
tituskx: Yes makes sense 👍

soda/core/soda/cli/cli.py
tituskx: Yes makes sense will change 👍

tituskx:
> Approach looks good to me @tituskx I have a few general questions though:
> 
> * The issue mentions that we should leave the python method for datasources in which we can't perform the SQL approach. Did you miss that perhaps?
>   
>   * I think in theory it wouldn't hurt (see my last bullet)

Hey @bastienboutonnet thanks for the review. No didn't miss that but since there seems to be no support for non-database datasources it would not have been used if I kept it in. I can add it in if you think that makes sense, but my personal opinion is that it might make more sense once we start adding support for non-database datasources. That way there's no unused code laying around + it would be a bit easier to add this functionality when we know where it will be used and how.

> * On the ordering, if you think it definitely has no impact then I would say no need to bother indeed.

Cool, I'll do some more tests to double check!

> * For the datasources you mention in the PR description, those are all experimental and community driven so indeed it's probably the case that they were never tested. However, no one thought that it might be a good idea to add a warning about db support. So I have a question:
>   
>   * is it the case that the old python approach also didn't work for these dbs? If "it didn't work" then we should throw a warning to day db is not supported. If it did work (or could) how about we leverage the old python code path?
> 
> I appreciate that the warning + mapping of supported data sources may be a little OOS. If I were you, I would consider implementing it as I think it's a simple dict "if not in this list say something" kind of thing. But if you'd like to move on and do it in another PR please make a ticket. I think we should be able to even complete it in this sprint given our current velocity.

[This test that was already present](https://github.com/sodadata/soda-core/pull/1847/files#diff-1efbd7cdb05c3bdfb308ad8ec6791d31fd6d4bffd662e858dfd726f23e27e7ccR17-R18) fails on main too (after enabling it for earlier mentioned datasources. I'm not sure whether the data sources are not supported though or whether it has to do with the testing. I'll figure that out and handle it in this ticket (mapping + message etc) 👍

tituskx:
@bastienboutonnet @baturayo Update:

* Figuring out how to test locally took too long. All the data sources were [already using SQL](https://github.com/sodadata/soda-core/pull/1847/files#diff-bb85d418b19f0af8eb3a93cbf5963e15aeaabb3bced924989cece5888979c7e4L320-L321), so I don't think the current change (which just adds the group and order by) will break support for previously supported data sources that are not tested.
* Added order by to the group by query. Since we're going to implement sampling in this PR: [feat: implement optional sample field for DRO update #1848
](https://github.com/sodadata/soda-core/pull/1848), we should get the results in descending order to get the most prevalent values first. 
* Add messages to `update_dro` and `DistributionCheck` if the functionality is not tested for the respective data source. 
* Refactored the unit tests

bastienboutonnet (approved):
Great work @tituskx 

baturayo (approved):
Let's Give Titus Magic

PR #1850, opened 2023-03-24
fix: improve error handling when invalid metric is provided to AnomalyMetricCheck

tituskx:
This PR resolves: [Fix error handling when invalid metric is provided to AnomalyMetricCheck](https://sodadata.atlassian.net/browse/IA-37)

This PR: 

* Ensures that `self.skip_anomaly_check = True` is set if an exception occurs
* Improves the error message when an incorrect metric is used in the check


bastienboutonnet (approved):
A true delighter of users! Thanks @tituskx 

Issue #1692, opened 2022-11-16
Feature: User-defined checks using a sql file


ssulca:
**Goal**

Support custom metrics using  **sql_file key**

This would allow reference a relative file, 
instead of including all your customized SQL queries in the custom metrics in the scan YAML file. 
Similar to [soda-sql](https://docs.soda.io/soda-sql/sql_metrics.html#custom-metrics-using-file-reference)

i.e. 
```yaml
checks for dim_product:
  - product_stock >= 50:
      product_stock sql_file: product_stock.sql
```
and `product_stock.sql` would be:
```sql
SELECT COUNT(safety_stock_level - days_to_manufacture)
FROM dim_product
```

jmarien:
SODA-1300

ssulca:
Hi, I would like contribute with this issue, 
In `soda/core/soda/sodacl/sodacl_parser.py` in line [568](https://github.com/sodadata/soda-core/blob/main/soda/core/soda/sodacl/sodacl_parser.py#L568)
adding a new check like 
```python
elif configuration_key.endswith("query") or configuration_key.endswith("sql_file"):
```
In case of `sql_file` key, read from relative path using  `self.path_stack.file_path` from  `soda.common.parser.Parser` 
and using `soda.common.file_system.file_system` for get directory name relative to sodacl check file and read sql file

ssulca:
#1859

PR #1852, opened 2023-03-29
feat: user defined checks using sql file

ssulca:
This PR resolves: #1692 

This PR 
* Implements a new check using sql_file key as 
   
   ```python
    elif configuration_key.endswith("query") or configuration_key.endswith("sql_file"):
   ```
   In case of `sql_file` key, reading from relative path using `self.path_stack.file_path` from `soda.common.parser.Parser` and `using soda.common.file_system.file_system` for get directory name relative to sodacl check file and read sql file (similar to `soda-sql`)
* Migrate `test_sql_metric_with_sql_file` unit test from `soda-sql`


vijaykiran:
https://github.com/sodadata/soda-core/pull/1859

PR #1884, opened 2023-06-05
Update spark_data_source.py with configuration param

Clusiv:
Added param 'configuration' to support additional spark configurations like 'spark.executor.memory':'512m'


soda/spark/soda/data_sources/spark_data_source.py
```diff
@@ -59,6 +60,7 @@ def hive_connection_function(
         port=port,
         database=database,
         auth=auth_method,
+        configuration=configuration,
```
vijaykiran, line 63: Can you please use `kwargs` instead? Otherwise it will break other connection functions

vijaykiran (changes requested):
Please see my comment about kwarts ...

m1n0:
hi @Clusiv , could you please do the requested changes from above please?

m1n0:
Indeed @Rhiyo , thanks for pointing that out, closing

Issue #1887, opened 2023-06-08
Athena: Incorrect catalog column name for column metadata.


bichitra95:
For the Athena data source, the method `column_metadata_catalog_column()` return the schema column name instead of catalog column name. [Code Reference]( https://github.com/sodadata/soda-core/blob/e0d8f4056e1e212f9f55dd9b5230c95307b0cae9/soda/athena/soda/data_sources/athena_data_source.py#LL104C9-L104C41)

So while doing the scan for an Athena table, It's not able to fetch the associated columns for the given table.

jmarien:
SODA-1757

PR #1888, opened 2023-06-08
Athena catalog column name to default

bichitra95:
Related Issue: https://github.com/sodadata/soda-core/issues/1887

While profiling for a table, we are fetching the columns metadata for the table [here](https://github.com/sodadata/soda-core/blob/e0d8f4056e1e212f9f55dd9b5230c95307b0cae9/soda/core/soda/execution/check/profile_columns_run.py#L30). 
The query that generates for fetching the columns is of follows,

```
SELECT table_name, column_name, data_type 
FROM information_schema.columns 
WHERE (((table_name LIKE 'junaid_test') AND (lower(column_name) LIKE lower('%')))) 
  AND lower(table_schema) = '<database>' 
  AND lower(table_schema) = '<schema>'
ORDER BY ORDINAL_POSITION
```

So removed the override method to use the parent method which returns `table_catalog` instead of `table_schema`.


m1n0:
Thanks for reporting and fixing this @bichitra95 ! I apologize for the delay, this has slipped our attention for too long. Would you be able to resolve the conflicts please so that we can merge this? 🙏

m1n0:
hi @bichitra95 , are you pushing old commits to this branch? In any case I see a bunch of unrelated changes in this PR, nothing related to the original PR title, could you elaborate please?

m1n0:
OP is not responding and keeps pushing unrelated changes to this branch. Closing

PR #1952, opened 2023-10-24
Initial contracts transformation proposal

tombaeyens:


soda/contracts/design_decisions.md
```diff
@@ -0,0 +1,52 @@
+# Soda data contract YAML format
+
+### No data source & connection details
+
+We leave datasource, connection and schema out of the contract files
+  * Pro
+    * The same contract file can be used to run in CI/CD
+    * Leveraging the existing Soda configuration files
+  * Con: 
+    * more CLI args
```
bastienboutonnet, line 10: this can easily be addressed with a project config file where defaults could be held for example.

soda/contracts/design_decisions.md
```diff
@@ -0,0 +1,52 @@
...
+    * less repeatable
+    * less sharing of the sharable connection details & profiles
```
bastienboutonnet: I'm not sure I understand that issue so maybe let's discuss

soda/contracts/design_decisions.md
```diff
@@ -0,0 +1,52 @@
...
+We translate Soda data contract YAML format to SodaCL YAML string first and then feed the SodaCL YAML 
+string into a Soda scan.
+
+Pros:
+* Composable.  
+* Easier to build.
+* Users can review the intermediate SodaCL and debug that based on the SodaCL docs.
+
+Cons: 
+* No native error messages on the contract YAML lines.
```
bastienboutonnet, line 43: - is there any way in which we could somehow solve this technically (not that we **have** to but just thinking outloud.
- do we think we'd ever actually want to maintain a separate parser (like a real parser) so that we can go direct from the contract format into the structures that core then uses, giving us back more control and less compilations. 

just thinking outloud as a "might it be a good idea in the future" kind of thing.

soda/contracts/design_decisions.md
```diff
@@ -0,0 +1,52 @@
...
+* Extra 'compilation' step
+
+### New YAML framework
+
+See `soda/contracts/soda/contracts/yaml.py` 
+Is that a better YAML abstraction?
+It's intended for reading, not writing. Should we add ability to write on this same framework?
+For now we write using plain dicts/lists.  There is also the unpack() method.
+But full mutable data structures would require overloading the muting operators like eg __setitem__ etc
```
bastienboutonnet, line 52: I am not sure what you mean by write, but we can go over that in our chat

soda/contracts/README.md
```diff
@@ -0,0 +1,48 @@
...
+| `checks` | SodaCL checks.  See [SodaCL docs](https://docs.soda.io/soda-cl/metrics-and-checks.html) | list of SodaCL checks |
+
+# Column keys
+
+| Key | Description | YAML data type |
+| --- | ----------- | -------------- |
+| `name` | Name of the column as in the SQL engine. | string |
+| `data_type` | Name of the data type as in the SQL engine | string |
+| `not_null` | `not_null: true` creates a missing values check | boolean |
+| `unique` | `unique: true` creates a uniqueness check | boolean |
```
janet-can, line 47: Does this translate into a duplicate check, or a new uniqueness check? (As it once existed in Soda SQL.)

soda/contracts/design_decisions.md
```diff
@@ -0,0 +1,52 @@
...
+### Keys without spaces nor variables
+
+No spaces in keys.  No parsing of keys.  No variable parts in keys except for the column names.
+
+* Pro:
+  * More JSON compliant
+  * More validation from JSON schema
+  * More expected and in line with people's expectations
+* Con:
+  * Not similar to SodaCL
```
janet-can, line 28: For what it's worth, from a user perspective, having to mentally shift from SodaCL syntax to sort-of-SodaCL-syntax for data contracts could be annoying. If the syntax validation is already a challenge for users (myself included, and as evidenced by the forever absent extra indent or colon), we ought to consider pushing for the same syntax to reduce even this small amount of friction.

soda/contracts/tests/contracts/contract_execution_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+    source_yaml_string = dedent(
+        f"""
+        dataset: {table_name}
+
+        columns:
+          - name: id
+            data_type: character varying
+            unique: true
+            attributes:
```
janet-can, line 24: Attributes? Tell me more! :)

soda/contracts/design_decisions.md
janet-can: If I can chime in from a support perspective, I am understanding data contracts as kind of a layer on top of SodaCL so that contracts get "translated" to SodaCL, then SodaCL gets "translated" to SQL queries. If that approaches a reasonable description of things, then errors for contracts ought to be distinguishable from SodaCL errors so as to be able to troubleshoot. I.e. If I make a syntax error or enter an invalid value in data contract form and I get a None Type error or something like that, is it because Soda "mistranslated" the contract into SodaCL, or is it because I made a typo in my contract?

soda/contracts/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+checks:
+  - avg(distance) between 400 and 500
+```
+
+# Top level keys
+
+| Key | Description | Required | YAML data type |
+| --- | ----------- | -------- | -------------- |
+| `name` | Name of the dataset as in the SQL engine. | Required | string |
```
bastienboutonnet, line 36: I would advise for a more specific name here. Judging by your description this is the dataset name. I would therefore expect the key to be called `dataset` no?

soda/contracts/README.md
```diff
@@ -0,0 +1,107 @@
...
+| `missing_*` | Ensures a missing values check and uses all the `missing_*` configurations | Optional | list of strings or numbers |
+| `valid_*` * `invalid_*` | Ensures a validity check and uses all the `valid_*` and `invalid_*` configurations | Optional | list of strings or numbers |
+| `unique` | `unique: true` ensures a uniqueness check | Optional | boolean |
+
+# Contract to SodaCL translation
+
+## The schema check
+
+A contract verification will always check the schema.  The contract schema check will verify that the list of columns
+in the database dataset matches *exactly* with the columns in the contract file.  All columns listed in the contract
```
bastienboutonnet, line 56: so this basically leads to a schema check that warns of `any` changes happen correct? In fact, maybe if you can show me what CL the example above would lead to that would be even easier to talk about it (as I won't review the code -- I leave that to Milan)

Which would also mean that the index (order) of the columns is also checked? If so, I don't know if I really like that. I find that more often than not, index (column position) changes are not an issue, however, they happen very often.

If the answer to the above question was going to be yes, then I was wondering if we could somewhere let people tell us what kind of schema check they want (or at least how strict). But let's see what the answer to the question is.

soda/contracts/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+"Enforcement" of a contract comes down to verifying that a certain dataset (like eg a table) complies with the specification in
+the contract file.  When the contract does not comply, the data owner and potentially the consumers should be notified.
+
+In your python (virtual) environment, ensure that the libraries `soda-core` and `soda-core-contracts` are available
+as well as the `soda-core-xxxx` library for the SQL engine of your choice.
+
+To verify if a dataset complies with the contract, here's the code snippet.
+
+```python
```
bastienboutonnet, line 86: is is the case that we won't offer a CLI option to enforce a contract?

bastienboutonnet (commented):
Hey Tom, nice one, I think it's mostly good from my side, but I did have a few questions which might just be in the domain of clarifications.

soda/contracts/README.md
tombaeyens: Good point as I was in doubt.  I first called it `dataset` as you suggest.  

i wanted to make it consistent with the columns. In the context of a column, it's normal to call it `name`.  Not sure if that needs consistency.  Also realized that all the properties on file level are related to the dataset.  I also considered `dataset_name`.  

Other option is to use `physical_name` in case we also want non-tech names in there later.

soda/contracts/README.md
tombaeyens: > a schema check that warns of any changes happen correct?

Warns if there is any difference between the contract schema and the measured schema of the table.  (It is not related change over time)

> Which would also mean that the index (order) of the columns is also checked?

At the moment, the index is not checked.  I agree that this is too volatile and not a common cause of issues.

soda/contracts/README.md
tombaeyens: It translates to the existing SodaCL `duplicate_count(column_name)` check

soda/contracts/design_decisions.md
tombaeyens: I think the better JSON schema support and user expectations for YAML in this case win from the consistency with SodaCL.

I also don't think that engineers will switch regularly between Soda Contract YAML and SodaCL YAML.  I would think it's going to be mostly new users that will use the contract language and only that.

soda/contracts/tests/contracts/contract_execution_test.py
tombaeyens: These have to be first implemented on Soda Cloud.  Atm, they are not as far as I know.  I will remove them from the code/tests

soda/contracts/README.md
bastienboutonnet: I guess it depends how much in the future we're talking about. IMO the programmatic way is shitty to maintain cos you have to reapeat all this boilerplate code which, unless you have a good reason to go for programmatic I think most users would choose CLI.

To me it also seems rather trivial to add a new command or option to the scan command to also do the translation step. But maybe I don't see something

soda/contracts/README.md
bastienboutonnet: > Warns if there is any difference between the contract schema and the measured schema of the table. (It is not related change over time)

Ah you're not leveraging the schema check we have in core I see. Ok I guess then all my questions go away :)

soda/contracts/README.md
bastienboutonnet: At this point and given unsure directions of what to add and so on, I'd advocate for the most specific name for now `dataset_name`. Then later if we want to let other things in then we'll likely not clash and I'm sure that if we want to introduce abstractions like `phsycal_name` or whatever we can find a way to reason about defaults and so on.

What do you think?

soda/contracts/design_decisions.md
janet-can: They won't regularly switch? Even though a user can add SodaCL checks to the same file?

janet-can:
Something to note: we collectively made a decision with Milan L. and other back in November 2022 to remove `missing format` as a documented configuration key.  

> [missing metrics](https://docs.soda.io/soda-cl/missing-metrics.html) behave the same way as [validity metrics](https://docs.soda.io/soda-cl/validity-metrics.html) - so missing format is not intended to accept some sort of pattern, but rather one of the [pre-defined formats](https://docs.soda.io/soda-cl/validity-metrics.html#list-of-valid-formats)  (i.e. be consistent with invalid_* checks) That being said, we discussed that using a missing format in such way is equivalent to using a validity check with a format, so we dont really need to document that one.
So the proposal is:
remove missing format from the missing metrics page

See thread, incl. Jira ref: https://sodadata.slack.com/archives/C02J6Q493PY/p1667416384017759

Given that the data contracts YAML borrows from SodaCL constructs, should we also match the experience by not documenting the `missing_format` configuration key?

PR #1973, opened 2023-12-14
GCP Delegate Authentication support

nathadfield:
When authenticating against GCP it is also possible to use a sequence of accounts through which impersonation can occur.  For example, `my-account` -> `first-account` -> `second-account`.

In this scenario, `my-account` is granted `Service Account Token Creator` against `first-account` which, in turn, also has the same role associated against `second-account`.

The GCP [impersonated_credentials](https://google-auth.readthedocs.io/en/master/reference/google.auth.impersonated_credentials.html) module can receive the argument `delegates` and is expecting a sequence of SA strings.

By providing a list of delegates in the Soda configuration file as shown below this can then be extracted from `data_source_properties` and submitted for authentication.

```
data_source my_dataset:
  type: bigquery
  connection:
    project_id: my_project
    use_context_auth: True
    impersonation_account: second-account@second-project.iam.gserviceaccount.com
    delegates: ["first-account@first-project.iam.gserviceaccount.com"]
```

### DIFF
--- a/soda/bigquery/soda/data_sources/bigquery_data_source.py
+++ b/soda/bigquery/soda/data_sources/bigquery_data_source.py
@@ -104,10 +104,16 @@ class BigQueryDataSource(DataSource):
 
         if self.data_source_properties.get("impersonation_account"):
             self.logs.info("Using impersonation of Service Account.")
+            if self.data_source_properties.get("delegates"):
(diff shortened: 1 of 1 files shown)


m1n0:
[parallel PR](https://github.com/sodadata/soda-core/pull/1976) to run the CI passed, merging

PR #1975, opened 2023-12-15
fix anomaly detection frequency aggregation bug

baturayo:


bastienboutonnet (approved):
Good to go, see with @m1n0 to get it merged

PR #1978, opened 2023-12-20
Enhance scan script, split execute to smaller functions.

duynguyenhoang:
Hello Soda team!

We are evaluating and adapting Soda as our main data quality tool. This is our first PR.

What include in the PR:
- Split `execute` function to smaller functions. It also helps us to write test easier.
- Allow us to calculate metrics value only without running validation. It is quite useful for us to do dry-run before actually validate data.


m1n0:
Hi! Thanks for the contribution! Could you please explain in a bit more detail how exactly this helps you? I get the split, but I would appreciate a bit more context and possibly a bit more thought into designing the split, thanks!

PR #1982, opened 2024-01-04
Hive data source improvements

robertomorandeira:
Hello Soda team!

We are evaluating Soda with the Hive data source. We have detected some issues that are included in the PR:

- PyHive log is too verbose
- Describe Table is failing in Hive
- Add support for Kerberos


m1n0 (approved):
Hello @robertomorandeira, thanks for the contribution! 

I do not have a hive+kerberos testing env but the change looks good

PR #1984, opened 2024-01-05
Document group by example for Soda Core with failed rows check

janet-can:
Adding documentation for how to use Soda Core and a failed rows check to group check results by category.


examples/group-by.md
```diff
@@ -0,0 +1,47 @@
...
+```
+![group-by-1](/docs/assets/images/group-by-1.png){:height="600px" width="600px"}
+2. Build a query to select groups with the relevant aggregations.
+```sql
+SELECT country, AVG(age) as avg_age
+FROM Customers
+GROUP BY country
+```
+![group-by-2](/docs/assets/images/group-by-2.png){:height="600px" width="600px"}
+3. Add a common table expression (CTE) to identify the "bad" group (where the average age is less than 25) from among the grouped results.
```
m1n0, line 21: CTE is not really necessary here, 
```
SELECT country, AVG(age) as avg_age
	FROM Customers
	GROUP BY country
    HAVING AVG(age) < 25
```

would work as well, and its a more efficient compute and memory wise. If CTE is used intentionally then I have no objections, just pointing this out.

janet-can: Oh, cool, that's a good idea. It's not necessary, just something that emerged out of the original version ages ago. I'll adjust as you suggest, then merge (if I have that permission).

m1n0 (approved):

Issue #1985, opened 2024-01-09
samples limit is not honored by failed rows checks


denisdallinga:
We discovered an issue with sample limits not being applied when using failed row checks.

```
checks for my_db.my_table [daily]:
    - failed rows:
        samples limit: 5
        # Any row that matches this dataset filter fails
        fail condition: 1 = 1
        name: test check
```
Some things we've tried with a similar configuration as above:
- Even though the samples limit is explicitly defined in the check, the sampler collects the complete dataset, not applying the limit of 5 as specified in the definition.
- The bug is apparent when using both the `fail condition`  and `fail query` .
- If the samples limit is left out of the definition, still the whole dataset is collected, and not the default of 100 rows as mentioned in the documentation.
- Other  types of checks do correctly apply the samples limit, we've verified this with `failed_count`  and `missing_count` checks. The `LIMIT` can be seen being applied when inspecting the SQL queries generated by Soda.

Confirmed not working on Soda Core `3.1.2`

[Mention](https://soda-community.slack.com/archives/C038FFU79J5/p1703695567778029) in soda-core channel in the soda community slack

## To reproduce the issue
```python
from pyspark.sql import SparkSession
from soda.scan import Scan
from soda.sampler.sampler import Sampler, SampleContext

# custom sampler to print the amount of rows being sampled
class CountRowsSampler(Sampler):
    def store_sample(self, sample_context: SampleContext):
        print(f'sample size: {len(sample_context.sample.get_rows())}')

spark_session = SparkSession.builder.config('spark.app.name', 'pytest-spark').getOrCreate()

# 101 entries, to see that the sampler does not honor the explicit(5) or default limit(100)
data = [['1']] * 101
df = spark_session.createDataFrame(data=data, schema=['number_one'])
df.createOrReplaceTempView('dataframe')

# setup the scan
scan = Scan()
scan.add_spark_session(spark_session=spark_session)
scan.set_data_source_name('spark_df')

check = """
checks for dataframe:
    - failed rows:
        samples limit: 5
        # Any row that matches this dataset filter fails
        fail condition: 1 = 1
"""
scan.add_sodacl_yaml_str(check)
scan.sampler = CountRowsSampler()

scan.execute()
```

prints:
```
sample size: 101
```

tools-soda:
SAS-2664

denisdallinga:
I made a first attempt at a partial fix for this: https://github.com/sodadata/soda-core/pull/1986

[...]

PR #1986, opened 2024-01-11
Apply sample limit to UDFailedRowsExpressionQuery

denisdallinga:
Github Issue: [#1985](https://github.com/sodadata/soda-core/issues/1985)

This changes the way samples are collected from `UserDefinedFailedRowsExpressionQueries`. In order to apply the samples limit given in the check configuration, or apply the default samples limit, the failed rows expression needs to fire off an additional SampleQuery. The SampleQuery executes a copy of the query, appending a LIMIT clause.

### DIFF
--- a/soda/core/soda/execution/check/user_defined_failed_rows_expression_check.py
+++ b/soda/core/soda/execution/check/user_defined_failed_rows_expression_check.py
@@ -81,6 +81,7 @@ class UserDefinedFailedRowsExpressionCheck(Check):
             scan = self.data_source_scan.scan
             condition = scan.jinja_resolve(definition=partition_filter, location=self.check_cfg.location)
             sql += f"\n      AND ({condition})"
+
         return sql
 
     def get_log_diagnostic_dict(self) -> dict:
--- a/soda/core/soda/execution/query/query.py
+++ b/soda/core/soda/execution/query/query.py
@@ -160,7 +160,7 @@ class Query:
         finally:
             self.duration = datetime.now() - start
 
-    def store(self):
+    def store(self, allow_samples=True):
         """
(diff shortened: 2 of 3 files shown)


NathanBick:
@denisdallinga I'm working on [a PR](https://github.com/sodadata/soda-core/pull/1991) to address this issue. Maybe we can team up to get more momentum on this issue?

m1n0 (changes requested):
Thanks for this contribution! It looks like this works, but I would suggest a few changes to make it consistent with how other check work:
- use `.fetchone()` method of the query class to get the result
- set the check metric value 
- if there are results and check has samples limit >0 then create a new sample query object with limit

You can check duplicate check query that does exactly that for inspiration, it should be pretty much 100% copy-paste except for the sql itself. https://github.com/sodadata/soda-core/blob/main/soda/core/soda/execution/query/duplicates_query.py#L95-L107

One note - please keep in mind that different data sources use different syntax for limit - TOP/FETCH FIRST, and it's even different syntax (you dont just append it to the end of the query) - so the way to go here is to:
- refactor [`get_failed_rows_sql`](https://github.com/sodadata/soda-core/blob/main/soda/core/soda/execution/check/user_defined_failed_rows_expression_check.py#L70) to take the sql string itself from the data_source class
- override that sql generation in each datasource that does not use LIMIT - the ones that need that override have the `LIMIT_KEYWORD` attribute overridden

Lastly - this should be covered by tests as its error-prone with all the variances in how limit is done across data sources, there already is a test for how the limit keyword is applied [here](https://github.com/sodadata/soda-core/blob/main/soda/core/tests/data_source/test_samples.py#L252), it just needs to be adjusted to include user defined check with expression.

I hope this is not too much!

PR #1994, opened 2024-01-23
bug: handle null values for continuous dist (#165)

baturayo:
cherry-picked from soda-library


dirkgroenen (commented):
I'l have to blindly approve here, as I am missing the context for these changes, but given you've already verified this with Titus on Soda Library I'm pretty sure to trust all is good.

Issue #2007, opened 2024-02-04
How to connect to Dremio Cloud?


JorisTruong:
Hi,

I am trying to try out `soda-core` with Dremio Cloud. According to [Soda's documentation](https://docs.soda.io/soda/connect-dremio.html#connection-configuration-reference), the connection configuration should look like this:
```
data_source my_datasource_name:
  type: dremio
  host: 127.0.0.1
  port: 5432
  username: simple
  password: simple_pass
  schema: public
```

The [Dremio's documentation](https://docs.dremio.com/cloud/sonar/client-apps/python#config-file-options) suggests to use `data.dremio.cloud` as the host and a token instead of username/password.

Any advice? I also have installed the ODBC driver for Arrow Flight SQL as it looks like this is what is used the the [source code](https://github.com/sodadata/soda-core/blob/70b8753d24f9a8e9de5912874eaf9ed5dc60538c/soda/dremio/soda/data_sources/dremio_data_source.py#L65), but it has not been working for me so far.

Here is a code snippet with the error log:
```
scan = Scan()
scan.set_data_source_name("cost_analysis")
scan.add_configuration_yaml_str(
    """
    data_source cost_analysis:
      type: dremio
      host: data.dremio.cloud
      port: 443
      token: <personal access token>
"""
)
scan.add_sodacl_yaml_str(sodacl_yaml_str)
scan.execute()
scan.assert_no_checks_fail()
```

`Encountered a problem while trying to connect to dremio: can only concatenate str (not "int") to str`

tools-soda:
SAS-2800

m1n0:
Hi! This is currently not supported, but it should be a fairly straightforward change. Would you be open to help and contribute token support?

JorisTruong:
Closed as #2009 has been merged

PR #2009, opened 2024-02-06
Dremio Token Support

JorisTruong:
Related to #2007.

I am not sure how to add tests though, given that Dremio Cloud is the SaaS version of Dremio. Here is what I tried using my own token:

image

Using a wrong token would raise this error: `pyodbc.Error: ('HY000', '[HY000] [Apache Arrow][Flight SQL] (100) Flight RPC failed with message:  (100) (SQLDriverConnect); [HY000] [Apache Arrow][Flight SQL] (100) Flight RPC failed with message:  (100)')`

Please advise if you have ideas, would gladly add some tests.


m1n0 (approved):
Thank you for the prompt contribution @JorisTruong ! 

As for tests - I don't think this small change would justify spinning up a Dremio instance just to test this one connectivity option - we already have adding Dremio container and running tests on in in the backlog so that will suffice for now.

PR #2012, opened 2024-02-08
Update oracle_data_source.py

vinod901:
Accommodate the Oracle data source to connect with `host`, `port` and `service_name` passed to it as different properties in the soda config file. We can still pass `connectstring` property.

Now, we can do
```yml
data_source xxxxxx :
  type: oracle
  username: xxxxxxx
  password: xxxxxxxx
  connectstring: {oracle_connectstring}
```
or
```yml
data_source xxxxxx :
  type: oracle
  username: xxxxxxx
  password: xxxxxxxx
  host: {host_address}
  port: {port_number}
  service_name: {service_name}
```
`soda-core` will create the `connectstring` from the given `{host_address}`, `{port_number}` and the `{service_name}` and connects to the data source

> Default value of `port` property is set to `1523`


m1n0 (approved):

m1n0:
Thanks for contributing this @vinod901 !

PR #2023, opened 2024-02-21
fix: timezone mismatch between the recent and historical ad results

baturayo:
Together with timezone, I also fixed the log messages to not do 2 cherry-pick for such a small stuff


bastienboutonnet (approved):

PR #2025, opened 2024-02-27
feat: in anomaly detection simulator use soda core historic check results endpoint instead of test results

baturayo:
- Fix integer detection
- Use sodaCoreCheckIdentities endpoint to get check ids for simulator


bastienboutonnet (approved):

PR #2027, opened 2024-02-29
feat: implement daily and monthly seasonality to external regressor …

baturayo:
- refactor the existing external regressor
- Instead of using the delta between yhat and y, only use the binary value
- Add three potential external regressors and each regressor can only have binary values
      - yearly
      - monthly
      - weekly 
- Implement custom external regressor for monthly and yearly


bastienboutonnet (approved):

PR #2033, opened 2024-03-11
Add parenthesis to check filter

abargar:


abargar:
Note that the quotations-around-column-name formatting caused inconsistent behavior with the SparkDF DataSource. This was easily fixable in this case, but might be good to note in the following doc:
Screenshot 2024-03-12 at 15 44 44

The formatting was fixed in the commit `dda387f`.

abargar:
Hey! Just checking in. Anything I need to do to help see this PR through? :)

m1n0:
hi @abargar , isn't the same issue present with the metric filter as well [here](https://github.com/sodadata/soda-core/blob/main/soda/core/soda/execution/metric/numeric_query_metric.py#L109)?

PR #2045, opened 2024-03-23
Bump dev requirements

m1n0:


m1n0:
downgrading tox and removing usage of 4x version as 4x was messing up other requirements

PR #2047, opened 2024-03-27
Rename argument in set_scan_results_file method

### LINKED ISSUE #2040: Change argument name in set_scan_results_file method
I think both the function and its argument having the same name can make confusions here. I'm suggesting to change the argument name to something like `scan_results_file`.

https://github.com/sodadata/soda-core/blob/a1a200862e79af4f01847241133d88a1260568b8/soda/core/soda/scan.py#L127-L128

ozgenbaris1:
The argument name `set_scan_results_file` in the `set_scan_results_file` method was confusing as it shared the same name as the function itself. This commit renames the argument to `scan_results_file` for improved clarity and consistency. Resolves #2040

### DIFF
--- a/soda/core/soda/scan.py
+++ b/soda/core/soda/scan.py
@@ -124,8 +124,8 @@ class Scan:
         global verbose
         verbose = verbose_var
 
-    def set_scan_results_file(self, set_scan_results_file: str):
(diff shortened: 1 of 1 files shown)


m1n0:
Thanks @ozgenbaris1 !

PR #2063, opened 2024-04-18
Update db2_data_source.py

4rahulae:
This is needed in order to fix the connectivity issues for db2 from soda-core.

Please add this change as part of the soda-core release

### DIFF
--- a/soda/db2/soda/data_sources/db2_data_source.py
+++ b/soda/db2/soda/data_sources/db2_data_source.py
@@ -65,11 +65,12 @@ class Db2DataSource(DataSource):
         self.username = data_source_properties.get("username")
         self.database = data_source_properties.get("database")
         self.schema = data_source_properties.get("schema")
+        self.verify_ssl = data_source_properties.get("security")
         self.update_schema(self.schema)
 
     def connect(self):
         conn_str = (
-            f"DATABASE={self.database};HOSTNAME={self.host};PORT={self.port};UID={self.username};PWD={self.password}"
+             f"DATABASE={self.database};HOSTNAME={self.host};PORT={self.port};UID={self.username};PWD={self.password};SECURITY={self.verify_ssl}"
         )
         self.connection = ibm_db_dbi.connect(conn_str)
         return self.connection
--- a/soda/db2/soda/data_sources/db2_data_source.py
+++ b/soda/db2/soda/data_sources/db2_data_source.py
@@ -69,7 +69,7 @@ class Db2DataSource(DataSource):
         self.update_schema(self.schema)
 
     def connect(self):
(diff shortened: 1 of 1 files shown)


4rahulae:
once the approvals are done and the changes are done, how soon can this be added to the new release?

4rahulae:
@m1n0 @vijaykiran Hi Milan as per our discussions earlier can you please review this PR and approve it and merge it for release?

dirkgroenen:
Hi @4rahulae, thanks for your contribution 🙌   @m1n0 is currently traveling back from our company offsite, so he'll likely not reach out today.  

I'm myself not able to test this at the moment, but could you assist by verifying that your changes do not impact configurations where `security` is not set?  According to the [documentation](https://www.ibm.com/docs/en/db2/11.5?topic=properties-connectionstring) I can see that it only accepts `SSL` as a valid value, so I would like to make sure this does not introduce backwards compatability issues for anyone who does not define `security` in their `configuration.yml`.  

(For example a possible scenario could be that Python translates it into an empty string or None, triggering DB2 to fail)

4rahulae:
> Hi @4rahulae, thanks for your contribution 🙌 @m1n0 is currently traveling back from our company offsite, so he'll likely not reach out today.
> 
> I'm myself not able to test this at the moment, but could you assist by verifying that your changes do not impact configurations where `security` is not set? According to the [documentation](https://www.ibm.com/docs/en/db2/11.5?topic=properties-connectionstring) I can see that it only accepts `SSL` as a valid value, so I would like to make sure this does not introduce backwards compatability issues for anyone who does not define `security` in their `configuration.yml`.
> 
> (For example a possible scenario could be that Python translates it into an empty string or None, triggering DB2 to fail)

m1n0:
thanks for the contrib @4rahulae ! I would suggest a bit more conservative way to do this though - pythons `get` without a second param returns NULL, so the `VERIFY_SSL=` will contain nothing, I would suggest only appending it to the connection string if it does have a non-null value, would you agree?
Also, I think it would make sense to match the yaml key we read with the connection string argument, so read `verify_ssl` instead of `security` if that makes sense.

m1n0:
hi, I don't know what the default value for `SECURITY` is, so maybe I would suggest a more safe/generic way to do this:
```
self.security = data_source_properties.get("security")

conn_str =  f"DATABASE={self.database};HOSTNAME={self.host};PORT={self.port};UID={self.username};PWD={self.password}"

if self.security is not None:
  conn_str += f";SECURITY={self.security}"

what do you think?

4rahulae:
Hi @m1n0 added changes as per your suggestion and tested them successfully on my local, please review the changes and let me know if it looks good to be merged?

m1n0:
Thanks for the contrib, merging!

PR #2095, opened 2024-06-07
Add session token parameter to athena data source connection

ruixuantan:


m1n0:
Thanks for the contrib!

PR #2097, opened 2024-06-12
Python 3.11 full support

m1n0:


m1n0:
Hi! Thanks for the suggestion. We have already tried to push this through internally but we need to address an issue with dask-sql - https://github.com/dask-contrib/dask-sql/issues/1346
Python 3.11 is otherwise fully supported by Soda, only CI does not run with it. Feel free to look into that Dask issue, otherwise I will update this once we have internal resources to address it.

PR #2105, opened 2024-06-25
Do not invoke failed rows API when no failed rows

jzalucki:


soda/core/tests/helpers/test_scan.py
```diff
@@ -80,9 +83,9 @@ def enable_mock_soda_cloud(self) -> MockSodaCloud:
             self._configuration.soda_cloud = MockSodaCloud(self)
         return self._configuration.soda_cloud
 
-    def enable_mock_sampler(self) -> MockSampler:
-        if not isinstance(self._configuration.sampler, MockSampler):
-            self._configuration.sampler = MockSampler()
+    def enable_mock_sampler(self, SamplerClass: Type[T] = MockSampler) -> T:
```
m1n0, line 86: clever! the only suggestion I have is snake_case, the var name should be snake_case, although it is a bit confusing then to call 
```
self._configuration.sampler = sampler_class()
```
but imho it is clearer for pythonic readability

Also, I think I would use type instead of isinstance so that it does not match subclasses as that can get annoyingly hard to debug https://stackoverflow.com/questions/152580/whats-the-canonical-way-to-check-for-type-in-python

m1n0 (changes requested):
LGTM, I like the mock request/response setup that might be useful for other stuff later.

I only had one small suggestion for the mock sampler enable

soda/core/tests/helpers/test_scan.py
jzalucki: Fixed.

m1n0 (approved):

jzalucki:
License/CLA problem branch recreated `CLOUD-7845/http-no-failed-rows-`

PR #2111, opened 2024-06-26
Skip validation if variables used for between threshold

jzalucki:


m1n0 (approved):
LGTM, maybe just change that log to a debug instead of info? I don't think it's particularly useful for non-verbose users

jzalucki:
License/CLA problem branch recreated `SAS-2427/between-threshold-error-with-variables-`

PR #2128, opened 2024-07-05
More date format regex corrections

pholser:
* Previously, a two-digit month of "00" would have been admitted as valid
* Legitimate year values, e.g. "1899", would have bee disallowed.


m1n0 (commented):
Thanks for the contribution! Looks good to me.

PR #2131, opened 2024-07-12
Remove spaces at the beginning and at the end of the query.

jzalucki:
The issue here is a trailing "\n" sign in the query.

### DIFF
--- a/soda/core/tests/data_source/test_group_evolution.py
+++ b/soda/core/tests/data_source/test_group_evolution.py
@@ -27,3 +27,24 @@ def test_group_evolution(data_source_fixture: DataSourceFixture):
     scan.execute()
 
     scan.assert_all_checks_pass()
+
+def test_group_evolution_query_multiline(data_source_fixture: DataSourceFixture):
+    table_name = data_source_fixture.ensure_test_table(customers_test_table)
+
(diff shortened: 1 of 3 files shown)


soda/core/tests/data_source/test_group_evolution.py
```diff
@@ -27,3 +27,29 @@ def test_group_evolution(data_source_fixture: DataSourceFixture):
     scan.execute()
 
     scan.assert_all_checks_pass()
+
+
+@pytest.mark.skipif(
+    test_data_source not in ["postgres", "bigquery", "spark_df", "oracle"],
```
m1n0, line 33: You can do this using the datasource casify methods like here https://github.com/sodadata/soda-core/blob/c510e0b1a65ad13c0852d1a4decd86483a31b229/soda/core/tests/data_source/test_for_each_dataset.py#L43-L57

m1n0 (changes requested):
We should be able to run the test on all datasources

soda/core/tests/data_source/test_group_evolution.py
jzalucki: Fixed.

m1n0 (approved):

PR #2133, opened 2024-07-15
ISO 8601 date should accept 24-hr times

pholser:


m1n0 (approved):

m1n0:
Thanks for the contrib, merging

PR #2137, opened 2024-07-25
Optional status and type field for contracts in Atlan

bichitra95:
Atlan API now accepts contracts without `type` and `status` fields as mandatory fields. 
 
**type**: If the `type` field is absent in the contract, Atlan will try to fetch the type with the help of the qualifiedName provided for supported dataset types. If there are multiple supported datasets present in Atlan for the given qualifiedName, the API will through an error. 

**status**: status is optional now. The default value of status is DRAFT if not present in the contract.


tombaeyens:
I have applied this change in the new contracts7 branch which will be the next contracts release

https://github.com/sodadata/soda-core/tree/contracts7

PR #2144, opened 2024-08-06
Don't use strtobool from distutils

### LINKED ISSUE #2091: distutils is deprecated. distutils.util.strtobool needs replacement.
As of Python 3.10, distutils has been deprecated, `strtobool` should be replaced in https://github.com/sodadata/soda-core/blob/main/soda/core/soda/telemetry/soda_telemetry.py

PEP 632 recomends copying the implementation into your own code, or reimplementing:
https://www.python.org/dev/peps/pep-0632/#migration-advice

akx:
Inspired by [this question](https://stackoverflow.com/q/78838254/51685) – one shouldn't need `distutils`/`setuptools` just to see if a string feels truthy.

The set of truthy-feeling strings was borrowed from `setuptools`.

Fixes #2091.


Swastik-SQED:
`strtobool` is also used in env_helper.py and should also be replaced,
https://github.com/sodadata/soda-core/blob/main/soda/core/soda/common/env_helper.py#L4

It would be great if this PR can also address env_helper

akx:
@Swastik-SQED Sure – this PR is open for maintainers to edit as they see fit. That file was added 2 months ago, this PR was created nearly a year ago... I didn't have a crystal ball at that point :)

AndreGodinho7:
I saw that https://github.com/sodadata/soda-core/pull/2439 was merged, any release planned for soda v3 to be compatible with python 3.12? @paulteehan @m1n0

PR #2156, opened 2024-09-03
Dataset level configuration for attributes and samples columns.

jzalucki:


jzalucki:
Do not port to soda-core as its doesn't bring much value.

m1n0:
@jzalucki wouldn't it still make sense here? Attributes still work in Cloud-less scenario, and samples columns too as they can be used in http sampler or custom sampler

jzalucki:
During the process it turned out it won't be that easy to port it here. If we think it's still worth to add it to soda-core we can create a separate task for that.

PR #2160, opened 2024-09-10
Add support for Azure SQL, Synapse, and Microsoft Fabric and extend support for SQL Server

sdebruyn:
This PR adds support for Microsoft Fabric Lakehouses and Data Warehouses as a data source and along with it extends support for the existing SQL Server data source.

What this PR does/includes:

## Extend authentication methods for SQL Server
This should make a lot of T-SQL users happy. I am also one of the lead maintainers for the dbt adapters which do the same thing. So basically I added the features which the dbt adapters support to this data source as well.

The data source now supports all the common authentication methods users of SQL Server (on-prem), Azure SQL DB, and Azure SQL Server Managed Instance use. This has probably also opened the door to compatibility with Azure Synapse Dedicated SQL Pools (because of the data types they probably should use the dbt-fabric adapter).

## Create a new Fabric data source as a child of the SQL Server data source
The Fabric data source is different from the one for SQL Server in which data types are allowed. Furthermore it inherits from SQL Server.

## No failing tests :)
Fabric T-SQL surface area is a bit smaller than the one in SQL Server. SQL Server was skipped on a lot of tests and this will unfortunately be the same for Fabric at this moment.
```
pytest soda/core/tests/
...
343 passed, 40 skipped, 750 warnings in 101.46s (0:01:41)
```
I'd be happy to work with you to add this to your CI but then we'd need to work together through Slack to configure the cloud infrastructure.


m1n0 (approved):
Thanks for this, the authentication work is very much appreciated! I just bumped tbump/version, ran sqlserver tests locally and this LGTM, merging!

PR #2162, opened 2024-09-12
Loosening duckdb dependency to less than 1.1.0

tombaeyens:


jzalucki (approved):

jochenchrist:
DuckDB releases [1.2.0](https://duckdb.org/docs/installation/?version=stable&environment=python).

Maybe you could loosen this dependency even further.

m1n0:
duckdb 1.2.0+ unfortunately requires some extra work, so merging this as-is for now. Feel free to open PR that adds 1.2.0+ support

sammyzord:
Hi all, DuckDB 1.0.0 is subject to a vulnerability, namely CVE-2024-41672. The patched version is 1.1.0, I saw in the preceding messages that allowing for 1.2.0+ would require substantial rework, but is it the same for version==1.1.0?

Thanks in advance!

Issue #2064, opened 2024-04-23
Yaml emitter error while executing scans concurrently


Electrogenic:
Hi Soda team.

## Background
We are using the `soda-core-trino` version `3.3.1` library inside of a `Fast API` server. The following code works when handling a single request at a time. However, when multiple requests come in at the same time (even just 2), we are running into the Yaml emitter error found in the stacktrace below. I'm wondering if there might be some thread safety issues going on within the ruaml yaml emitter, or in the way it is being used.

```python
scan = Scan()
scan.set_data_source_name('default')
# Trino data source is configured
scan.add_configuration_yaml_str('<REDACTED>')
# I simplified the checks for this example, but I'm using about 240 in actuality.
checks = """
    checks for my_table:
    - schema:
        name: Check that field foo is present
        fail:
            when required column missing:
            - foo
    """
scan.add_sodacl_yaml_str(checks)
soda_scan_status_code = scan.execute()
```

## Stacktrace
```Traceback (most recent call last):
   File "/workspace/app/routes/v1/contracts/routes.py", line 419, in execute_soda_scann
     soda_scan_status_code = scan.execute()
   File "/workspace/.venv/lib/python3.10/site-packages/soda/scan.py", line 638, in execute
     self.scan_results = self.build_scan_results()
   File "/workspace/.venv/lib/python3.10/site-packages/soda/scan.py", line 72, in build_scan_results
     checks = [check.get_dict() for check in self._checks if check.outcome is not None and check.archetype is None]
   File "/workspace/.venv/lib/python3.10/site-packages/soda/scan.py", line 72, in <listcomp>
     checks = [check.get_dict() for check in self._checks if check.outcome is not None and check.archetype is None]
   File "/workspace/.venv/lib/python3.10/site-packages/soda/execution/check/check.py", line 363, in get_dict
     "identity": self.create_identity(with_datasource=True, with_filename=True),
   File "/workspace/.venv/lib/python3.10/site-packages/soda/execution/check/check.py", line 234, in create_identity
     identity_source_configurations_yaml = to_yaml_str(identity_source_configurations)
   File "/workspace/.venv/lib/python3.10/site-packages/soda/common/yaml_helper.py", line 13, in to_yaml_str
     return YamlHelper.to_yaml(yaml_object)
   File "/workspace/.venv/lib/python3.10/site-packages/soda/common/yaml_helper.py", line 24, in to_yaml
     cls.__yaml.dump(yaml_object, stream)
[...]

PR #2163, opened 2024-09-13
Fix YamlHelper for thread safety in multi-threaded environments.

NiklasA:
### Overview
This pull request addresses the thread safety issue with YamlHelper as reported in [Issue #2064](https://github.com/sodadata/soda-core/issues/2064). The modifications ensure that YamlHelper can be safely used in multi-threaded environments, preventing potential data races and concurrency errors.

@m1n0 Can you please review the changes and provide feedback or approve the merge if everything looks in order. Thank you!

### DIFF
--- a/soda/core/soda/common/yaml_helper.py
+++ b/soda/core/soda/common/yaml_helper.py
@@ -1,5 +1,5 @@
 from ruamel.yaml import YAML, StringIO
-
+import threading
 
 def create_yaml() -> YAML:
     yaml = YAML()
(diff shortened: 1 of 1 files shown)


m1n0:
hi @NiklasA, thanks for looking into this and thanks for your patience! I am thinking that opening a thread for this is a bit of an overkill. I looked into it myself and it seems to me that all that needs to happen is to have a new yaml instance instead of reusing the same one like this https://github.com/sodadata/soda-core/pull/2188 - my local testing confirms it working, so I merged it to unblock a release, please test it and get back to me. We can definitely revisit your proposal for the fix if it still does not fix the issue

NiklasA:
Hi @m1n0 ,

thank you for your detailed response! We’ve tested your approach and compared it to our existing implementation. While your solution works well, we’ve observed that its performance is approximately 10% slower than our current approach (within our existing load tests).

For context, we are using the `datacontract-cli` as part of our data ingestion processes, where we need to verify a significant amount of data. This might represent an edge case, but in these scenarios, performance is critical for us.

We appreciate your input and are happy to discuss potential optimizations or alternative approaches that might bridge this gap. Let us know your thoughts!

Best regards
Niklas

NiklasA:
**FYI:** @PascalRosenberger, @ManbirP and @SebastianHirsch

PR #2165, opened 2024-09-18
Comparison row count check secondary datasource filter fix

asantoz:
# Description
This PR fix the bug on `RowCountComparisonCheck` in case of a secondary table datasource with filter compare, the current bug ignores the partition_sql filter.

# Changes

- Fix partition creation on `RowCountComparisonCheck` adding the partition config


m1n0 (approved):
Thanks for the fix, this LGTM, merging!

PR #2176, opened 2024-10-04
Fix FATAL: unsupported startup parameter: options

RekunDzmitry:
Sometimes PostgresDataSource fails on connection to a Postgres database with an error - FATAL: unsupported startup parameter: options.
The goal of that PR is to fix that behavior and remove options from the connection if they are not supported.


RekunDzmitry:
Hey! Just checking in. Anything I need to do to help see this PR through? :)

m1n0:
Thanks for the contrib!

PR #2202, opened 2025-01-23
chore: update ruamel.yaml version range to >=0.17.0,<0.19.0

syou6162:


m1n0:
Thanks for the contrib, but I looked into the failing tests and they uncover why we have this pinned to 0.18. I cannot dive deeper into seeing how could we resolve this right now, feel free to suggest a fix. Running postgres and mysql tests locally will replicate the issue for you.

PR #2199, opened 2025-01-22
Release opentelemetry version contraint (#2192)

ghjklw:
Release soda-core version contraint for opentelemetry as there is no reason to support Python 3.7 and this is creating dependency conflicts with many other libraries.

Closes #2192

PR #2212, opened 2025-02-24
Fix opentelemetry dependency version  (#2192)

### LINKED ISSUE #2192: Outdated opentelemetry dependency creating conflicts
The currently supported versions of [opentelemetry](https://pypi.org/project/opentelemetry-api/#history) are up to 1.22 which are now a year old. This was last changed in march this year via #2043.

The reason I've stumbled across this is due to [Airflow no longer supporting opentelemetry <1.22](https://github.com/apache/airflow/pull/44113) explicitly in their most recent release.

The package version limit committed to in soda-core in March '24 appears to maintain support for Python 3.7, which makes it more of a tricky discussion I'm sure, however not moving with later versions of other python packages (opentelemetry is just the current problem) will cause many conflicts with other packages in the future I'm sure.

It would be nice to move on from 3.7 which will mean many other packages (e.g markupsafe #2103) can also be bumped.

### LINKED ISSUE #2213: Potential Redundant Constraint in `opentelemetry-api` Dependency
In [PR #2199](https://github.com/sodadata/soda-core/pull/2199), the `opentelemetry-api` dependency is specified as `opentelemetry-api~=1.16.0,<2.0.0`.  

I believe this might be a typo, as `~=1.16.0` already implies `<2.0.0`, making the `<2.0.0` constraint redundant.  

Perhaps the intended specification was `opentelemetry-api>=1.16.0,<2.0.0`.  

The current constraint (`~=1.16.0`) is causing conflicts with `apache-airflow==2.10.4+astro.1` in my project. Could you please review and consider adjusting it?

ghjklw:
Fix a mistake I made in #2199 

The previous PR set "opentelemetry-api~=1.16.0,<2.0.0", but this still only allows max version to be 1.16.*.
For example, the airflow conflict apache-airflow 2.10.4+astro.1 depends on opentelemetry-api>=1.24.0 is still present.

This PR fixes it by setting the dependency to "opentelemetry-api>=1.16.0,<2.0.0"

Thanks @lewis-anderson53 for pointing it out!

Fixes #2192
Fixes #2213


ruimgf (approved):

m1n0 (approved):
thanks, looks good to me!

m1n0:
released in 3.5.1 last week https://github.com/sodadata/soda-core/releases/tag/v3.5.1

PR #2218, opened 2025-03-08
Remove dependency of soda-core-spark-df on pyspark

### LINKED ISSUE #2217: soda-core-spark-df unnecessary dependency on PySpark
Soda-core-spark-df declares a dependency on `pyspark>=3.4.0`. Looking through the code, it seems to be used only for typing, and should therefore be declared as a dev (eventually optional/extra) dependency.

This makes it unfortunately impossible with `databricks-connect`: https://docs.databricks.com/aws/en/dev-tools/databricks-connect/python/install#install-the-databricks-connect-client-with-venv

ghjklw:
Soda-core-spark-df declares a dependency on pyspark>=3.4.0, but pyspark used only for typing, and should therefore be declared as an extra dependency to avoid unnecessary dependency conflicts.

Closes #2217

### DIFF
--- a/soda/spark_df/setup.py
+++ b/soda/spark_df/setup.py
@@ -8,12 +8,14 @@ description = "Soda Core Spark Dataframe Package"
 
 requires = [
     f"soda-core-spark=={package_version}",
-    "pyspark>=3.4.0",
 ]
+extras = {"pyspark": ["pyspark>=3.4.0"]}
+
 # TODO Fix the params
 setup(
     name=package_name,
     version=package_version,
     install_requires=requires,
+    extras_require=extras,
     packages=find_namespace_packages(include=["soda*"]),
 )
--- a/soda/spark_df/soda/data_sources/spark_df_connection.py
+++ b/soda/spark_df/soda/data_sources/spark_df_connection.py
@@ -1,9 +1,13 @@
-from pyspark.sql.session import SparkSession
+from typing import TYPE_CHECKING
+
 from soda.data_sources.spark_df_cursor import SparkDfCursor
+
+if TYPE_CHECKING:
+    from pyspark.sql.session import SparkSession
 
 
 class SparkDfConnection:
-    def __init__(self, spark_session: SparkSession):
+    def __init__(self, spark_session: "SparkSession"):
         self.spark_session = spark_session
 
     def cursor(self) -> SparkDfCursor:
--- a/soda/spark_df/soda/data_sources/spark_df_contract_data_source.py
+++ b/soda/spark_df/soda/data_sources/spark_df_contract_data_source.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 import logging
 import re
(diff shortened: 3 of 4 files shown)


m1n0:
Thanks for the contrib @ghjklw. The pinned version ensures that the expected programmatic api (i.e. classes in this cases) is present. Version we are using is 2 years old, do you think that is still too much to ask?

ghjklw:
Hi @m1n0, I think that in this case the version requirement is perfectly reasonable, no issue with that 🙂

The specific challenge I'm facing is that the packages `pyspark` and `databricks-connect` both provide the module `pyspark` and are incompatible as documented in https://docs.databricks.com/aws/en/dev-tools/databricks-connect/python/install#install-the-databricks-connect-client-with-venv and .https://docs.databricks.com/aws/en/dev-tools/databricks-connect/python/troubleshooting#conflicting-pyspark-installations

To be fair, this is much more of a Databricks issue than a Soda issue, but it's also much easier to fix on this side 😉 In practice, when I override `soda-core-spark-df` dependency on `pyspark`, it works flawlessly.

By "ensures that the expected programmatic api (i.e. classes in this cases) is present", do you mean:
- when running unit/integration tests? In this case I believe my PR covers it as we still install PySpark to run the tests
- at runtime? If that is the case, I think (personal opinion 😉) that the solution has stronger consequences than the risk it mitigates. The only PySpark methods used the I found are: `SparkSession.sql`, `DataFrame.collect`, `DataFrame.count`, `DataFrame.offset`, `DataFrame.limit`, `DataFrame.schema.fields`. Of these, only `offset` is somewhat new (version `3.4.0`).

  Do you think it might be a good enough solution to update the documentation to say to install `soda-spark-df[pyspark]` unless you have an existing `pyspark>=3.4.0` installation? This would work for 99% of persons, but those who need the flexibility on how they manage their PySpark environment would still be able to do so. You could even add a specific note about `databricks-connect`.

m1n0:
The `offset` method is what caused the pin to be `>= 3.4.0`. I don't think just having it in docs would be a good-enough UX, especially coming back from the more strict one that we have now. The pin was motivated by multiple users coming to us with `.cursor() is not defined` issues.

I would rather move forward and not back on this, how about and extra that removes the pyspark dependency? Doing that and adding it to docs would not break existing user experience and allow users in scenarios like yours to move forward without friction

AndreGodinho7:
any updates on this issue? what is the way now to install soda v3 with databricks connect? @m1n0 @ghjklw

devlucasc:
Any updates on this? I tried to use `datacontracts-cli` on Databricks serverless, but datacontracts-cli depends on `soda-core-spark-df` that has the dependency on `pyspark` and it can cause dependency conflicts due to spark connect as well. I could ran by installing `databricks-connect` package right after the instalation of datacontracts-cli, but it would be nice if it is possible to move this `pyspark` dependency to optional extras. Thanks

PR #2219, opened 2025-03-10
Validity check: fix complex mixed rules query

m1n0:


mivds (commented):
Change looks good, but not sure the test cases are actually covering the new functionality. The `customers_test_table` has 4 distinct values in the `cat` column:
- `"HIGH"`
- `"MEDIUM"`
- `"LOW"`
- `None`

Suggest adding one of those first three in the `invalid values` condition, e.g. with max length 6. That should then result in some failing checks.

m1n0:
Good idea @mivds, thanks! I updated the tests

PR #2221, opened 2025-03-11
allow user to override scope for Microsoft Entra ID access token

sdebruyn:
I am the contributor of the Microsoft Entra ID auth support in Soda.

Currently I am trying to connect from a Fabric Notebook to my SQL Server Managed Instance. But I made a logical mistake in my first version of the code below where I always assumed that in Fabric Notebooks you would connect a Fabric Data Warehouse or Lakehouse.

The code below changes that and tries to be smart about the scope for the access token. The user can pass a `scope` to indicate the scope for the access token. If not, it will try to figure out the scope based on the hostname.


m1n0 (approved):
LGTM, thanks for the follow up!

PR #2231, opened 2025-04-09
Use pydantic models for data source config

m1n0:
First stab at using models for data source:
- no drastic changes, I contained myself to minimal scope so that we can review, merge and iterate quickly
- no actual validation using pydantic is done yet
- the packages are a mess, a more generic approach will follow

after all of the above is done, I will start with actual extra data source implementation. I want to get this at least somewhat right so that we don't pass around dicts for no reason, and so that the implementation is not too scattered. I am not super happy about how the three different classes compose a data source but I do get the intention to separate the mode from the connection and dialect. I will try to iterate on this design so that it's more intuitive to walk through and work with.

### DIFF
--- a/soda-core/src/soda_core/cli/handlers/data_source.py
+++ b/soda-core/src/soda_core/cli/handlers/data_source.py
@@ -5,7 +5,7 @@ from typing import Optional
 
 from soda_core.cli.exit_codes import ExitCode
 from soda_core.common.logging_constants import Emoticons, soda_logger
-from soda_core.common.yaml import YamlSource
+from soda_core.common.yaml import DataSourceYamlSource
 
 
 def handle_create_data_source(data_source_file_path: str, data_source_type: str) -> ExitCode:
@@ -47,7 +47,9 @@ def handle_test_data_source(data_source_file_path: str) -> ExitCode:
     soda_logger.info(f"Testing data source configuration file {data_source_file_path}")
     from soda_core.common.data_source_impl import DataSourceImpl
 
-    data_source_impl: DataSourceImpl = DataSourceImpl.from_yaml_source(YamlSource.from_file_path(data_source_file_path))
+    data_source_impl: DataSourceImpl = DataSourceImpl.from_yaml_source(
+        DataSourceYamlSource.from_file_path(data_source_file_path)
+    )
     error_message: Optional[str] = (
         data_source_impl.test_connection_error_message()
         if data_source_impl
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -42,18 +42,10 @@ class DataSourceConnection(ABC):
         """
         if self.connection is None:
             try:
-                self._log_connection_properties_excl_pwd(self.connection_properties)
+                logger.debug(f"'{self.name}' connection properties: {self.connection_properties}")
                 self.connection = self._create_connection(self.connection_properties)
             except Exception as e:
                 logger.error(msg=f"Could not connect to '{self.name}': {e}", exc_info=True)
-
-    def _log_connection_properties_excl_pwd(self, connection_yaml_dict: dict):
-        dict_without_pwd = {
-            k: v
-            for k, v in connection_yaml_dict.items()
-            if not "password" in k and not "pwd" in k and not "key" in k and not "secret" in k
-        }
-        logger.debug(f"{self.name} connection properties: {dict_without_pwd}")
 
     def close_connection(self) -> None:
         """
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -6,6 +6,7 @@ from typing import Optional
 
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_results import QueryResult, UpdateResult
+from soda_core.common.exceptions import DataSourceConnectionException
 from soda_core.common.logging_constants import soda_logger
 from soda_core.common.sql_dialect import SqlDialect
 from soda_core.common.statements.metadata_columns_query import (
@@ -15,6 +16,7 @@ from soda_core.common.statements.metadata_columns_query import (
 from soda_core.common.statements.metadata_tables_query import MetadataTablesQuery
 from soda_core.common.yaml import YamlObject, YamlSource
 from soda_core.contracts.contract_verification import DataSource
+from soda_core.model.data_source.data_source import DataSourceBase
 
 logger: logging.Logger = soda_logger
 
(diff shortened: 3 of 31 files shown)


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -42,19 +42,11 @@ def open_connection(self) -> None:
         """
         if self.connection is None:
             try:
-                self._log_connection_properties_excl_pwd(self.connection_properties)
+                logger.debug(f"'{self.name}' connection properties: {self.connection_properties}")
                 self.connection = self._create_connection(self.connection_properties)
             except Exception as e:
                 logger.error(msg=f"Could not connect to '{self.name}': {e}", exc_info=True)
 
-    def _log_connection_properties_excl_pwd(self, connection_yaml_dict: dict):
```
m1n0, line 50: pydantic secret field takes care of not logging passwords

soda-core/src/soda_core/cli/handlers/data_source.py
```diff
@@ -47,7 +47,9 @@ def handle_test_data_source(data_source_file_path: str) -> ExitCode:
     soda_logger.info(f"Testing data source configuration file {data_source_file_path}")
     from soda_core.common.data_source_impl import DataSourceImpl
 
-    data_source_impl: DataSourceImpl = DataSourceImpl.from_yaml_source(YamlSource.from_file_path(data_source_file_path))
+    data_source_impl: DataSourceImpl = DataSourceImpl.from_yaml_source(
+        DataSourceYamlSource.from_file_path(data_source_file_path)
```
m1n0, line 51: the only point of DataSourceYamlSource subclass is to have the type constant set by default so that we don't manipulate the state of the object somewhere between parsing and creating data structures

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -24,60 +26,43 @@ class DataSourceImpl(ABC):
...
             PostgresDataSourceImpl,
         )
 
-        return PostgresDataSourceImpl(
-            data_source_yaml_source=data_source_yaml_source,
-            name=name,
-            type_name=type_name,
-            connection_properties=connection_properties,
-        )
+        return PostgresDataSourceImpl(data_source_model=data_source_model)
```
m1n0, line 56: proper factory TBD

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -42,19 +42,11 @@ def open_connection(self) -> None:
         """
         if self.connection is None:
             try:
-                self._log_connection_properties_excl_pwd(self.connection_properties)
+                logger.debug(f"'{self.name}' connection properties: {self.connection_properties}")
```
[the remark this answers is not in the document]
m1n0: pydantic secret field type takes care of this

soda-core/src/soda_core/common/data_source_impl.py
nielsn: There's a nice little pattern using a _registry class var and __init_subclass__ for this

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -24,60 +26,43 @@ class DataSourceImpl(ABC):
...
-            name=data_source_name,
-            type_name=data_source_type_name,
-            connection_properties=connection_properties,
+        from pydantic import Field, TypeAdapter
+        from soda_postgres.model.data_source.postgres_data_source import (
+            PostgresDataSource,
         )
 
+        DataSourceModel = Annotated[Union[PostgresDataSource], Field(discriminator="type")]
+        adapter = TypeAdapter(DataSourceModel)
```
nielsn, line 42: Nice, had never really checked out what the `TypeAdapter` was for.

soda-core/src/soda_core/common/yaml.py
```diff
@@ -20,6 +21,10 @@ def __init__(self):
         self.ruamel_yaml_parser.preserve_quotes = True
 
 
+class FileType(str, Enum):
+    DATA_SOURCE = "data source"
```
nielsn, line 25: NIT: "data_source"?

soda-core/src/soda_core/common/yaml.py
```diff
@@ -165,6 +174,13 @@ def parse(self) -> Optional[YamlObject]:
                 )
 
 
+class DataSourceYamlSource(YamlSource):
+    def __init__(self, file_path: Optional[str] = None, yaml_str: Optional[str] = None):
+        super().__init__(file_path=file_path, yaml_str=yaml_str)
+
+        self.file_type: str = FileType.DATA_SOURCE
```
nielsn, line 181: I think we might avoid this by using the `__init_subclass__` dunder on the `YamlSource` class. Then you could just have a `class DataSourceYamlSource(YamlSource, file_type=FileType.DATA_SOURCE)` without anything else.

soda-postgres/src/soda_postgres/model/data_source/postgres_connection_properties.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    ssl_cert: Optional[str] = Field(None, description="Path to SSL client certificate")
+    ssl_key: Optional[str] = Field(None, description="Path to SSL client key")
+    ssl_root_cert: Optional[str] = Field(None, description="Path to SSL root certificate")
+
+    # Connection options
+    connection_timeout: Optional[int] = Field(None, description="Connection timeout in seconds")
+
+
+class PostgresConnectionPassword(PostgresConnectionPropertiesBase):
```
nielsn, line 35: I would have put all of the different auth options as subclasses of a specific `PostgresConnectionAuthProperties` base class and used that one within a top level `PostgresConnectionProperties` class.  

In this case you'll end up with trouble in composing stuff whenever a second property with variations (i.e. subclasses) comes along...

nielsn (approved):

soda-core/src/soda_core/common/yaml.py
m1n0: unsure actually, I just copied the value from here without change https://github.com/sodadata/soda-core/pull/2231/files#diff-327af2890051407092017844c1b242ab19ce67de0a69b9b69f81b133df954726L29

soda-core/src/soda_core/common/yaml.py
m1n0: ok I didn't know about that, and it prompted me to read a bunch on meta classes. Cool stuff 😅

soda-postgres/src/soda_postgres/model/data_source/postgres_connection_properties.py
m1n0: Indeed, adjusted

PR #2232, opened 2025-04-09
DTL-325: contracts CLI revision part 2

nielsn:
### Tweaked parameter verification

- Fail fast 
- Improved messages
- Introduced exceptions to communicate failed authentications
- Changed low level request logging for failed requests from `critical` to `debug` to enable it only when running `-v`

### Unblock verification on agent for remote contracts

- Hotfix missing `metadata.source.filepath` value by adding a fixed string for now. Later development will fetch existing metadata along with the contract, so we can reuse it. Right now we cannot pass this information all the way down to the place where the call to Soda Cloud is being made.
- Implement actual feedback required have the proper exit code (`4` vs other values) for remote agent verification.


soda-core/src/soda_core/cli/handlers/contract.py
```diff
@@ -36,10 +36,13 @@ def handle_verify_contract(
...
-        contract_yaml_sources += [
-            YamlSource.from_str(soda_cloud_client.fetch_contract_for_dataset(dataset_identifier))
-            for dataset_identifier in dataset_identifiers
-        ]
+        for dataset_identifier in dataset_identifiers:
+            contract = soda_cloud_client.fetch_contract_for_dataset(dataset_identifier)
+            if not contract:
+                return (
+                    ExitCode.LOG_ERRORS
+                )  # TODO: verify this is OK as it breaks the "multiple contracts" functionality
```
m1n0, line 44: do we need to resolve this dilemma since it applies to local verification as well and therefore is a very fundamental decision?

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -476,40 +490,43 @@ def verify_contract_on_agent(
...
+            log_records=None,
+        )
+
         can_publish_and_verify, reason = self.can_publish_and_verify_contract(
             data_source_name, dataset_prefix, dataset_name
         )
         if not can_publish_and_verify:
             if reason is None:
                 logger.error(f"Skipping contract verification because of an error (see logs)")
+                verification_result.sending_results_to_soda_cloud_failed = True
```
m1n0, line 526: I am not pushing for strict immutable objects but these instance manipulations which significantly change the logic flow ten to become very murky over time. Would it make sense to remove this flag from ContractVerificationResult, create a sublcass of it which would be e.g. CloudContractVerificationResult, control the flow based on class instead of a bool flag and return a new instance of CLoudContractVerificationResult which just takes the non-cloud ContractVerificationResult one into constructor?

The downside is that it would eliminate the need of the flag, making the cloud subclass empty (nothing added compared to the non-cloud one) but it would make the ContractVerificationResult a single purpose class. Wdyt?

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -536,108 +553,101 @@ def verify_contract_on_agent(
...
-                scan_id=scan_id, blocking_timeout_in_minutes=blocking_timeout_in_minutes
-            )
+        if response.status_code != 200:
+            logger.error("Remote contract verification failed.")
+            verification_result.sending_results_to_soda_cloud_failed = True
+            return verification_result
+
+        if not scan_id:
+            logger.warning("Did not receive a Scan ID from Soda Cloud")
+            return verification_result
```
m1n0, line 563: There's a quite few `return` scenarios here, with quite a lot of logic/setup/datastructures in between, would it make sense to refactor these bits out into separate functions so that the overall flow here is more intuitive to read? e.g. verify contract command dict setup+getting a response could be a candidate that would become a one-liner here, the logging logic below as well, wdyt?

m1n0 (approved):
A few code design remarks, but none blocking a merge if we don't have the time to address them

soda-core/src/soda_core/common/soda_cloud.py
nielsn: This can definitely be cleaner, but I was counting on just using Exceptions to deal with this kind of "information transfer". 

Right now this flag is the only thing that allows us to interpret the result as either exit code `3` or `4`.
In a better world we just catch exceptions and translate those into exit codes when using CLI. Because the flag itself is already semantically incorrect.

So yes: this should be better, but I think regular exception support will fix that already without the need for more subclassing.

PR #2235, opened 2025-04-11
Datasource package loader/factory

m1n0:
This currently breaks the tests as adding the init dunder files creates explicit packages and the current structure results in conflicting package names making tests impossible to load. Functionality and CLI itself works, a refactor of the structure will follow but I did not want to create a huge PR that does both.


soda-core/src/soda_core/__init__.py
```diff
@@ -0,0 +1,3 @@
+from soda_core.plugin_loader import load_data_source_plugins
+
+load_data_source_plugins()
```
m1n0, line 3: this one ensures no registration in packages will be needed

soda-core/src/soda_core/model/data_source/data_source.py
```diff
@@ -14,8 +15,13 @@ class DataSourceBase(
...
     name: str = Field(..., description="Data source name")
-    type: str = Field(..., description="Data source type")
+    type: Literal["data_source"]
     # Alias connection -> connection_properties to read that from yaml config
     connection_properties: DataSourceConnectionProperties = Field(
         ..., alias="connection", description="Data source connection details"
     )
+
+    @classmethod
+    def get_class_type(cls) -> str:
```
m1n0, line 25: the only thing that is a bit hacky - to get the simple "postgres" string for the registering. That string is available in the plugin register thingie but it's not passed along into the init subclass

m1n0:
@nielsn merging to unblock other work, please do review and I can adjust afterwards

nielsn:
Had some feedback to discuss after stand-up, can still look into it then

PR #2242, opened 2025-04-16
Support remote contract verification without publishing

nielsn:
- Pass publish flag value to `soda_cloud` call
- When not publishing, call `sodaCoreTestContract` API instead

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -489,7 +489,10 @@ class SodaCloud:
         return allowed, reason
 
     def verify_contract_on_agent(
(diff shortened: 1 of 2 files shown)


LaurenDebruyn (commented):
https://github.com/sodadata/soda-core/pull/2242/files#diff-9add66a9e380e084cee7fb5275ec38d9671d0ed4fffca33f3d56fdf0b2d103dbR564-R567
```python
if response.status_code != 200:
  logger.error("Remote contract verification failed.")
  verification_result.sending_results_to_soda_cloud_failed = True
  return verification_result
```
Technically this can never be `True` when we don't publish the results.

LaurenDebruyn (approved):

PR #2244, opened 2025-04-22
Removed default casify from test suite

tombaeyens:


soda-snowflake/tests/data_sources/snowflake_data_source_test_helper.py
```diff
@@ -49,7 +49,9 @@ def _create_data_source_yaml_str(self) -> str:
         """
 
     def create_test_schema_if_not_exists_sql(self) -> str:
-        return f"CREATE SCHEMA IF NOT EXISTS {self.dataset_prefix[1]};"
+        sql_dialect: "SqlDialect" = self.data_source_impl.sql_dialect
+        schema_name: str = self.dataset_prefix[1]
```
m1n0, line 53: we access the dataset prefixes over the codebases like this by index, it will work fine for the most part but perhaps this could be a dialect method that takes in the prefixes and returns the one that should be the schema to make it safer

tombaeyens: Good suggestion!  I ll see if I can do that in a subsequent PR

PR #2248, opened 2025-04-23
V4 check filters

tombaeyens:


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -206,6 +206,11 @@ def log_summary(self, logs: Logs) -> None:
         for diagnostic in self.diagnostics:
             logger.info(f"  {diagnostic.log_line()}")
 
+    def get_numeric_diagnostic_value(self, diagnostic_name: str) -> any:
```
m1n0, line 209: @tombaeyens is there any guarantee there is only one numeric diagnostic present?

tombaeyens: @m1n0 No guarantee but this is a helper method for testing.  So first one is good for that use case.  I could also move this method to a test helper class if that keeps things cleaner.  Or we could name it `get_first_numeric_diagnostic_value`.  Any preference?

m1n0: how about both:
- move to test helper
- change the name to include "first"
- add a method comment saying what it is useful for

m1n0 (approved):

PR #2250, opened 2025-04-24
Pass variables when executing on agent

LaurenDebruyn:


soda-core/src/soda_core/cli/soda.py
```diff
@@ -240,7 +240,7 @@ def handle(args):
         contract_file_paths = args.contract
         data_source_file_path = args.data_source
 
-        exit_code = handle_test_contract(contract_file_paths, data_source_file_path)
+        exit_code = handle_test_contract(contract_file_paths, {}, data_source_file_path)
```
nielsn, line 243: Not sure how this one is supposed to behave... If it doesn't require the variables to be set then we shouldn't bother passing it into the `handle_test_contract` function at all. Just pass empty dict in the underlying function (or nothing at all).

If they are required for syntax validation, then we're missing the `--set` CLI setup

LaurenDebruyn: How I understand the test contract command, it would need the variables to validate the syntax because it mostly follows the verification flow. So that means we're still missing the `--set` CLI setup. Not sure if there is a ticket for this, but I can implement it as a follow-up.

PR #2252, opened 2025-04-24
V4 NOW variable updates

tombaeyens:
Wait for this PR to be merged first as it branches of that branch: https://github.com/sodadata/soda-core/pull/2248

It's only the last 3 commits that should be reviewed.  Other commits have already been merged to v4.

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -7,3 +7,18 @@ SODA_CLOUD_HOST=dev.sodadata.io
 #   2) Test if you have permissions. By default it should.  If not, ask around.
 SODA_CLOUD_API_KEY_ID=
 SODA_CLOUD_API_KEY_SECRET=
+
+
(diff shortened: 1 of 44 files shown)


tombaeyens:
@m1n0 After our discussion, made the following updates:

* Mocked the time.  I created a CurrentTime class for this.  Turns out that datetime.now is based on C and therefore normal mocking doesnt work.  It's either a 17K lib or 50 lines of code in our own lib.  Opted for the 50 lines of code.  In the main code we now need to use `CurrentTime.now()` instead of datetime.now() to get the current time.

* Stopped leaking the resolved and soda variables into ContractImpl

* Added ContractYaml.data_timestamp and used that to pass over ContractImpl to Soda Cloud

* Checked the timestamp format of NOW variables.

PR #2255, opened 2025-04-26
V4 Changed reference data dataset to qualified name

tombaeyens:
FYI, the docs will be updated in another PR


soda-core/src/soda_core/common/dataset_identifier.py
```diff
@@ -10,11 +17,29 @@ def __init__(self, data_source_name: str, prefixes: list[str], dataset_name: str
         self.dataset_name = dataset_name
 
     @classmethod
-    def parse(cls, identifier: Optional[str]) -> "DatasetIdentifier":
-        if not identifier:
+    def read(cls, yaml_object: YamlObject, yaml_key: str, required: bool = True) -> Optional[str]:
```
nielsn, line 20: I don't like having to deal with `YamlObject` instances here. 

This class should be kept clean, just dealing with strings at most.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -127,9 +127,8 @@ def __init__(
                 },
             )
 
-        self.dataset: Optional[str] = (
-            self.contract_yaml_object.read_string("dataset") if self.contract_yaml_object else None
-        )
+        self.dataset: Optional[str] = DatasetIdentifier.read(yaml_object=self.contract_yaml_object, yaml_key="dataset")
```
nielsn, line 130: I see why  you want to encapsulate the repetitive "fetch a field and parse it as a DatasetIdentifier" but I just don't think both responsibilities belong in the `DatasetIdentifier` class.

Getting a value from a `YamlObject` given a key seems basic enough to belong in the `YamlObject` class itself?

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -356,28 +352,9 @@ def __init__(self, variable_name: str, variable_yaml_object: YamlObject):
 
 class ValidReferenceDataYaml:
     def __init__(self, valid_reference_data_yaml: YamlObject):
-        dataset: any = valid_reference_data_yaml.read_value("dataset")
-        is_list_str: bool = isinstance(dataset, list) and all(isinstance(e, str) for e in dataset)
-        self.dataset: str | list[str] | None = dataset if isinstance(dataset, str) or is_list_str else None
+        self.dataset: Optional[str] = DatasetIdentifier.read(yaml_object=valid_reference_data_yaml, yaml_key="dataset")
```
nielsn, line 355: Same as previous remark

nielsn (changes requested):

soda-core/src/soda_core/contracts/impl/contract_yaml.py
tombaeyens: Moved DatasetIdentifier.read to YamlObject.read_dataset_identifier.  Better indeed 👍   

@nielsn Does that resolve all 3 comments?

tombaeyens:
Assuming comments are resolved ok. Waiting longer makes other merges harder...

PR #2257, opened 2025-04-28
V4 mocktime

tombaeyens:
I found a way to mock time.  But it has some quircks you need to know

Turns out that datetime.datetime.now cannot be mocked because it has native impl.

And it is possible to overwrite datetime.datetime with a class that inherits from datetime.datetime.

Especially the ordering of imports in the conftest.py is important.  I renamed the file to `a_time_machine.py` so that precommit would order the imports in the right order.

This solution requires that datetime.datetime is replaced with the `class TimeMachineDatetime(datetime.datetime)` ** before ** any `import datetime from datetime` is perfomed by the production code.


nielsn (approved):

soda-core/tests/helpers/a_time_machine.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TimeMachineDatetime(datetime.datetime):
+    NOW: Optional[datetime.datetime] = None
+
+    @classmethod
+    def now(cls, tz=None) -> datetime.datetime:
+        return cls.NOW if cls.NOW else super().now(tz)
+
+
+datetime.datetime = TimeMachineDatetime
```
m1n0, line 15: please move the comment from the PR into the code here. It will get forgotten in PR, but will be super handy if/when we need to change this code

soda-core/tests/features/test_dataset_filter.py
```diff
@@ -81,9 +81,11 @@ def test_dataset_filter(data_source_test_helper: DataSourceTestHelper):
               - missing:
     """
 
+    time_machine.set_now(t1)
+
     # On the first time partition t1 (16th) the filter should pass
     contract_verification_result_t1: ContractVerificationResult = data_source_test_helper.assert_contract_pass(
-        test_table=test_table, contract_yaml_str=contract_yaml_str, provided_now_timestamp=t1
```
m1n0, line 86: Can we get rid of the provided_now_timestamp parameter completely since it's not needed any more? Or is it someething you would like to use for backfilling/freshness or anything like that?

tombaeyens: provided_now_timestamp is completely replaced with the line above `time_machine.set_now(t1)`  No need to keep the provided now timestamp for other use cases later.

soda-core/tests/helpers/a_time_machine.py
tombaeyens: done 👍

m1n0 (approved):

PR #2259, opened 2025-04-28
Datasources: pass through connection properties as kwargs, add magic

m1n0:
- Passes kwargs as-is to driver
- Adds a bit of magic to handle secrets and connection dicts to simplify datasource specific code
- Adds support to use different name in yaml key and what is passed to the driver for controlled UX


nielsn (commented):
Very nice cleanup

PR #2262, opened 2025-04-30
V4 Added multi column duplicate checks

tombaeyens:
@m1n0 What I want you to review specifically here is how the `DuplicateCheckYamlParser` and `DuplicateCheckParser` do the split between single column and multi column check implementations.  So that we're aligned if this is a good mechanism to get different behaviors for the same check type.


tombaeyens:
I'll be fixing the sonar problems later in a separate PR.  First I want to merge things so that I don't get merge conflicts

PR #2264, opened 2025-04-30
V4 Added aggregate check type

tombaeyens:
Docs see https://github.com/sodadata/soda-core/blob/v4-numeric/docs/pages/aggregate_check.md


tombaeyens:
Added metric expression check type to this PR as well to avoid conflicts

m1n0 (approved):
lgtm, please do check the snowflake tests

tombaeyens:
@m1n0 snowflake seems to be installation issue, not the PR
```
FAILED soda-core/tests/components/test_data_source_api.py::test_data_source_env_var_resolving - ImportError: Data source type 'postgres' not available. Make sure to install the required plugin, e.g. `pip install soda-postgres`
```

docs/pages/freshness_check.md
```diff
@@ -0,0 +1,19 @@
+# Freshness check
+
+### Verify the freshness of the data
+
+```yaml
+dataset: postgres_adventureworks/adventureworks/advw/dim_employee
+
+checks:
+  - freshness:
```
m1n0, line 9: so freshness only works on dataset level, not supported to set it on the column ?

docs/pages/freshness_check.md
```diff
@@ -0,0 +1,19 @@
...
+      column: created_at
+      threshold:
+        must_be_less_than: 2
+```
+
+`column` configures a timestamp column in the dataset. (required) 
+
+`unit` configures the unit of measure for the threshold.  Can be `minute`, `hour` or `day` (optional, default is hour) 
+
+`now_variable` The name of the NOW variable. `var.` does not have to be included. (optional, default uses current time from `NOW` in the soda namespace)
```
m1n0, line 19: wouldn't it make sense to let the user pick the namespace? This way `soda.` vars cannot be used here

docs/pages/numeric_check.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Configure a check filter
+
+Apply the check only to a subset of the data.
+
+```yaml
+dataset: postgres_adventureworks/adventureworks/advw/dim_employee
+
+columns:
+      - numeric:
```
m1n0, line 30: wrong indentation

soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -208,7 +208,12 @@ def log_summary(self, logs: Logs) -> None:
 
     def get_numeric_diagnostic_value(self, diagnostic_name: str) -> any:
```
m1n0, line 209: I don't see any usage of this method in this PR or on the main (v4) branch, is it used anywhere?

m1n0 (approved):
LGTM, just a few small remarks/questions

soda-core/src/soda_core/contracts/contract_verification.py
tombaeyens: removed it.  good catch.

I moved this one and forgot to delete it.  Also renamed it to shorter version to make tests more readable.

docs/pages/freshness_check.md
tombaeyens: Indeed.  I would  think it's best to have it only on dataset level.  If we ever want special Soda Cloud editor support for the single or first freshness check.  Seems better if that is on the predictable dataset level instead of searching for it in the columns.  Also freshness may later involve multiple columns.  That could also get more consistent with dataset level checks.

docs/pages/freshness_check.md
m1n0: but why is soda. namespace forbidden?

PR #2276, opened 2025-05-12
Support multiple datasource connections

m1n0:


nielsn (approved):
Still in doubt if this is the way to go forward (and if we should not introduce an additional CLI argument with its own configuration file instead)

m1n0:
we went with explicit arguments for other data sources

PR #2277, opened 2025-05-12
DTL-548: Raise more understandable error when an invalid regex is passed

nielsn:
This PR introduces a new `InvalidRegexException`, subclass of the `InvalidContractException` tree (not sure if it belongs there). 

It will be raised whenever a runtime exception is raised BY THE DATASOURCE mentioning the regex.
This approach is good because:
- the datasource decides what regex to accept
- can be tailored to each datasource individually

But it is bad because:
- it is tailored to each datasource individually, so implementations are very dependent on the data source (and we should consider moving this as part of an interface implemented by the datasource)

Alternative: implement this kind of verification (if we want it) in the `RegexFormat` class and apply it during its `read` method? This does mean Python regexes will be THE standard for regexes.

### DIFF
--- a/soda-core/src/soda_core/common/exceptions.py
+++ b/soda-core/src/soda_core/common/exceptions.py
@@ -1,4 +1,5 @@
 from soda_core.common.dataset_identifier import DatasetIdentifier
+from typing import Optional
 
 
 class SodaCoreException(Exception):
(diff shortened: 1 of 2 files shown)


m1n0:
I hope I get the comment right, but I would not try to reconcile python and warehouse regexes, they are not the same implementation and some (e.g. MS SQL) don't support regexes out of the box at all.

PR #2299, opened 2025-06-04
Fix authentication inside Fabric Notebooks

sdebruyn:
Microsoft updated their mssparkutils package to be named notebookutils and also changed the scopes it can take to get a credential. This PR makes Soda compatible again with the current version of Fabric. To auth inside a Fabric Notebook you would just put authentication to "FabricSpark".


m1n0 (approved):
Thanks for the update @sdebruyn !

PR #2300, opened 2025-06-05
DTL-739: Fix broken `caplog` tests

nielsn:
Explicitly calling the `configure_logging` was breaking the logging setup due to all handlers being overridden etc. Not cool.

The way to do it: you should manually set up the logging, outside of the facade calls.

```python
from soda_core import configure_logging
from soda_core.contracts import verify_contracts_locally

configure_logging(verbose=True)  # Defaults to `False`

print("Running contract verification...")

res = verify_contracts_locally(
    dataset_identifiers=["bus_nienu_doggo/nyc/public/bus_breakdown_and_delays"],
    data_source_file_path="/home/niels/Code/soda/soda-core/configuration.yml",
    soda_cloud_file_path="/home/niels/Code/soda/soda-core/soda-cloud.yml",
)
```

### DIFF
--- a/soda-core/src/soda_core/__init__.py
+++ b/soda-core/src/soda_core/__init__.py
@@ -1,3 +1,8 @@
+from soda_core.common.logging_configuration import configure_logging
 from soda_core.plugins import load_plugins
+
+__all__ = [
+    "configure_logging",
+]
 
 load_plugins()


m1n0:
hmm isn't that a bit too technical stuff leaking from the API? I haven't encountered a lib requiring to explicitly configure logging for programmatic use 🤔

PR #2307, opened 2025-06-10
Fix diagnostics, support date-like in freshness

m1n0:


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1082,7 +1082,7 @@ def _build_diagnostics_json_dict(check_result: CheckResult) -> Optional[dict]:
         else None
     )
 
-    check_result_value: Optional[Number] = check_result.get_threshold_value()
+    check_result_value: Number = check_result.get_threshold_value()
```
nielsn: I would have actually kept the default handling `or 0` here, as right now it is acceptable from the `soda-core` domain that a check result has no threshold value (if not evaluated).

OR we should introduce a new status that informs us if the check has been evaluated or not, that way we can handle this part more cleanly (and correct)

soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -280,12 +281,18 @@ def _log_console_format(cls, n: Number) -> str:
...
         if self.threshold_metric_name and self.diagnostic_metric_values:
             v = self.diagnostic_metric_values.get(self.threshold_metric_name)
             if isinstance(v, Number):
                 return v
 
+        #  TODO: this default is here only because check.diaignostics.value is a required non-nullable field in the api.
+        # Re-think this when we switch to the new diagnostics format and/or make sure all check types produce a mandatory threshold metric+value pair.
+        # A poor default, but it really needs to be a number and not None or Nan.
+        # It's guaranteed to be 0 for non-evaluated checks, which is not ideal, but better than stopping ingestion.
+        return 0
```
nielsn: Yeah, see first comment.

nielsn (approved):

soda-core/src/soda_core/common/soda_cloud.py
tombaeyens: On the topic of required information in the API pushed to Soda Cloud: The implementations are all written from the assumption that unknown errors can occur in queries.  If unknown errors in executing the queries or metrics, then a check remains NOT_EVALUATED.  In that case it can be that the diagnostics are not present either.  That's something to keep in mind and if necessary, update the Soda Cloud backend API accordingly I think.

soda-core/src/soda_core/contracts/impl/check_types/schema_check.py
```diff
@@ -239,10 +239,20 @@ def __init__(
...
         are_columns_out_of_order: bool,
     ):
+        schema_events = sum(
+            (
+                len(expected_column_names_not_actual),
+                len(actual_column_names_not_expected),
+                len(column_data_type_mismatches),
+            )
+        )
+        diagnostic_metric_values = {"schema_events": schema_events}
```
tombaeyens: Consider renaming "schema_events" to "schema_events_count".  But only if you make other changes too.  Not worth to update PR for this thing only.

tombaeyens (approved):

PR #2314, opened 2025-06-11
DTL-807: Floor the current freshness in seconds so it returns an `int`

nielsn:
I think this makes the most sense given the unit. This also matches what Soda Cloud is expecting.

Adjusted tests accordingly.


tombaeyens (approved):
lgtm.  although i wonder how we can prevent these in the future.

nielsn:
> lgtm. although i wonder how we can prevent these in the future.

Meh, I feel like this is just a thought exercise really. Thinking about the interface and sticking to it in the implementation.

PR #2316, opened 2025-06-12
V4 failedrowsext

tombaeyens:


soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -203,6 +204,12 @@ def test_connection_error_message(self) -> Optional[str]:
     def build_data_source(self) -> DataSource:
         return DataSource(name=self.name, type=self.type_name)
 
+    def qualify_dataset_name(self, dataset_identifier: DatasetIdentifier) -> str:
+        assert dataset_identifier.data_source_name == self.name
```
m1n0: asserting like this seems unexpected, shouldn't this be a custom exception with a meaningful message?

soda-core/src/soda_core/common/extensions.py
```diff
@@ -11,7 +9,6 @@ def find_class_method(cls, module_name: str, class_name: str, method_name: str)
             module = import_module(module_name)
             class_ = getattr(module, class_name)
             return getattr(class_, method_name)
-        except AttributeError as e:
-            raise ExtensionException(
-                message=f"Feature '{class_name}.{method_name}' requires the Soda Extensions to be installed."
-            )
+        except (AttributeError, ModuleNotFoundError) as e:
+            # Extension not installed
```
m1n0: when would this happen, why do we change it to silently continue? I see the usage below, which tries to load the extension with a hardcoded module/class/method names, but that will be replaced by registering the extension using entry points. I would therefore keep the exception raising here and temporarily catch the exception and silently continue in the ContractVerificationHandler

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1196,3 +1198,25 @@ def _append_exception_to_cloud_log_dicts(cloud_log_dicts: list[dict], exception:
     exc_cloud_log_dict["index"] = len(cloud_log_dicts)
     cloud_log_dicts.append(exc_cloud_log_dict)
     return cloud_log_dicts
+
+
+class FailedRowsDiagnostic:
```
m1n0: I would use pydantic to skip having to create the constructors, especially the duplicate ones in the subclasses. Actually - all the 4 constructors are identical, there is no need to repeat them in the subclasses

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -28,6 +29,11 @@ def quote_default(self, identifier: Optional[str]) -> Optional[str]:
             else None
         )
 
+    def build_fully_qualified_sql_name(self, dataset_identifier: DatasetIdentifier) -> str:
```
m1n0, line 32: wouldn't it make sense to make this a method on the dataset identifier for more portability as we discussed?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -49,6 +52,28 @@
 logger: logging.Logger = soda_logger
 
 
+class ContractVerificationHandler:
+    @classmethod
+    def instance(cls, identifier: Optional[str] = None) -> Optional[ContractVerificationHandler]:
+        # TODO: replace with plugin extension mechanism
+        create_method: Callable[..., Optional[ContractVerificationHandler]] = Extensions.find_class_method(
```
m1n0: as mentioned above - I would leave the exception in the find_class_method and just catch it here

m1n0 (changes requested):

soda-core/src/soda_core/common/data_source_impl.py
tombaeyens: Added an error message, but kept the AssertionException because it's a bug, not a user error.  Can change to SodaException if you think that is better

soda-core/src/soda_core/common/extensions.py
tombaeyens: Removed extentions.py file and inlined the class instantiation into the calling class.

soda-core/src/soda_core/common/soda_cloud.py
tombaeyens: removed the constructors, replacing them with `pass` and to my surprise it still worked!  Nice

soda-core/src/soda_core/common/sql_dialect.py
tombaeyens: makes sense, but that has to be done in a lot of places.  planned to do that later all in 1 go.  ok?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
tombaeyens: Inlined the initialization

soda-core/src/soda_core/common/data_source_impl.py
m1n0: There is a lot of debate whether exceptions should be used for anticipated user flows or not, but using them for bugs is imho always valid

m1n0 (approved):
thanks for following up on the comments! The non-pydantic dto's are not great but we can resolve later.

soda-core/src/soda_core/common/data_source_impl.py
tombaeyens: In this case for sure because the code is executing a query on the wrong data source.  Can't be good :)

tombaeyens:
I know I won't convince, but still would like to point out that for me it's more nuanced.  I think you are assuming full knowledge of the pydantic framework.  For me, it's not always clear from reading the pydantic classes how the translation is done.

I think plain dict code can be write at least as easy to read in a way that the interface is still very clear.  And the code is easy to follow / debug.

PR #2317, opened 2025-06-13
Default names

tombaeyens:


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -241,6 +241,27 @@ def log_table_row(self) -> dict:
...
+        "invalid": "No invalid values",
+        "duplicate (col)": "No duplicate values",
+        "duplicate (multi-col)": "No duplicate values",
+        "aggregate": "Metric function meets threshold",
+        "metric": "Metric meets threshold",
+        "failed_rows": "No rows violating the condition",
+    }
+
+    def _get_name_with_default(self, check: Check) -> str:
+        if isinstance(check.name, str):
```
nielsn: Should probably be `if check.name and isinstance(check.name, str)` to deal with the `Optional[str]` properly?

nielsn (changes requested):
Nice, good clean up!

soda-core/src/soda_core/contracts/contract_verification.py
tombaeyens: What would that first `check.name and` add to it?  I think only diff would be that an empty string is resulting in false.

nielsn: Oh you're completely right. It would deal with `check` being `None` but that won't happen here. My bad, this is indeed all good!

nielsn (approved):

PR #2325, opened 2025-06-27
Fixing diagnostics

tombaeyens:
Also deleted the numeric check stuff as it was renamed to aggregate


tombaeyens:
Merging this should be synced with some updates on the backend.  @milanaleksic Can you let us know when the backend updates are ready?

m1n0 (approved):

milanaleksic:
BE side is merged, taking the latest request example as it is defined there, although Benjamin still needs to approve to make sure we are 100% aligned

tombaeyens:
Was related to DTL-887

PR #2328, opened 2025-07-02
Enabled failed rows diagnostics upload for source queries + dwh fixes

tombaeyens:


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -1151,9 +1151,22 @@ def build_identity_path(self) -> str:
         return "/".join(parts)
 
     def _build_definition(self) -> str:
+        source_contract_yaml_dict: dict = self.contract_impl.contract_yaml.contract_yaml_object.yaml_dict
+
+        contract_dict: dict = {}
+        if "filter" in source_contract_yaml_dict:
+            contract_dict["filter"] = source_contract_yaml_dict.get("filter")
```
m1n0: Filter is already parsed someplace else, we shouldn't access the yaml dict directly like this any more at this stage.

tombaeyens: Seems not that easy: The pydantic model object doesn't seem to be available in the ContractImpl nor the ContractYaml.  I think we need to review and decide cause now we have multiple models and we are in transition half way.  I don't think it's good to switch to pydantic if we don't have the error messages and the other aspects like docs and locations covered.  

Also the all-or-nothing parsing approach has to be reviewed and the decision has to be rolled out.   Meaning that with pydantic the idea is that if the parsing doesn't succeed anywhere, there is no model object at all.  Whereas with the current approach, the model is parses everything that matches the expected schema.  Which is a bit different logic.

Maybe that is not to the point... do you think i should do the effort to bring the pydantic model object to that place and double check the logic in case the parsing doesn't work out?

m1n0 (approved):

PR #2336, opened 2025-07-10
Obsl 548 add bigquery support

m1n0:


Niels-b:
PR is ready to review. All tests seem to pass in CI (including new tests on `v4` branch). 

Note: can't add @m1n0 to review as he created the PR. Also added @mivds. Do we add anyone else?

mivds:
> PR is ready to review. All tests seem to pass in CI (including new tests on `v4` branch).
> 
> Note: can't add @m1n0 to review as he created the PR. Also added @mivds. Do we add anyone else?

I think @m1n0 and I will be the main reviewers for now. @nielsn might also be interested, but he's on holiday currently.

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -4,13 +4,49 @@
 from typing import Optional
 
 
+class BaseSqlExpression:
```
nielsn, line 12: Too lazy to fully figure it out on a mobile phone screen, but why is the introduction of this concept required?

nielsn (approved):

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -0,0 +1,97 @@
...
+    def get_location(self) -> str:
+        location = self.data_source_model.connection_properties.location
+        if location is None:
+            if self.cached_location is not None:
+                location = self.cached_location
+            else:
+                result = self.execute_query("SELECT @@location")
+                location = result.rows[0][0]
+                logger.info(f"Detected BigQuery location: {location}")
+                self.cached_location = location
```
mivds: Suggest moving up the `cached_location` to avoid nested ifs.


```suggestion
        if self.cached_location is not None:
            location = self.cached_location
        elif self.data_source_model.connection_properties.location is not None:
            location = self.data_source_model.connection_properties.location
        else:
            result = self.execute_query("SELECT @@location")
            self.cached_location = location = result.rows[0][0]
            logger.info(f"Detected BigQuery location: {location}")
```

soda-core/src/soda_core/common/sql_ast.py
Niels-b: We are using this to build an SQL tree. This enables us to get the "context" (i.e. parent nodes in the tree) when generating SQL. 
This is especially relevant in the BigQuery `_build_tuple_sql` function that needs know when it is in a `DISTINCT`. Because BigQuery cannot use STRUCTS/ARRAYS/... with a DISTINCT.
Building the syntax tree seems like the most elegant solution. We need the `BaseSqlExpression` for the `SELECT`, `FROM`,... classes as they do not inherit from `SqlExpression` and we wanted to keep it that way.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -0,0 +1,97 @@
...
+class BigQuerySqlDialect(SqlDialect):
+    DEFAULT_QUOTE_CHAR = "`"
+
+    def default_casify(self, identifier: str) -> str:
+        return identifier.upper()
+
+    def _build_tuple_sql(self, tuple: TUPLE) -> str:
+        if tuple.check_context(DISTINCT):
+            return self._build_tuple_sql_in_distinct(tuple)
+        return f"[{super()._build_tuple_sql(tuple)}]"
```
mivds: Are the square brackets needed here? Unless I'm mistaken this evaluates to both brackets & parentheses, i.e. `[(value1, value2, ...)]`

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _build_tuple_sql_in_distinct(self, tuple: TUPLE) -> str:
+        return f"TO_JSON_STRING(STRUCT({super()._build_tuple_sql(tuple)}))"
```
mivds, line 74: I think the `TO_JSON_STRING` here isn't necessary. Looking at [the BigQuery docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_distinct) `DISTINCT(<struct type>)` should work.

mivds: Larger concern though is that this expression doesn't do what is intended. From what I can tell this is being created using syntax like `DISTINCT(TUPLE(...))`, but I don't think it works that way. At least not in BigQuery. The `DISTINCT` applies to the entire context, you can't select only a part of it. So something like `SELECT(DISTINCT(TUPLE("col1", "col2")), "col3")` would give incorrect results.

To give an example:
```sql
SELECT DISTINCT (table_schema, table_name), column_name FROM region-europe-west1.INFORMATION_SCHEMA.COLUMNS;
```

Results in:
```
[{"v":{"f":[{"v":"dev_mvds"},{"v":"SODATEST_CustomersProfilingCapitalized_8652307e"}]}}]	ITEMS_SOLD
[{"v":{"f":[{"v":"dev_mvds"},{"v":"SODATEST_CustomersProfilingCapitalized_8652307e"}]}}]	CST_Size
```
Note how there is a duplicate returned for the struct entry.


I would rethink the approach here. I'm not sure this covers everything, but I think introducing new operators like `SELECT_DISTINCT` and `COUNT_DISTINCT` makes more sense. That makes it clear the `DISTINCT` is operating on the full argument set, rather than a subset.

I believe that also means you don't need to know if you are within `DISTINCT` context or not. So the whole tracking parents inside the `SqlExpression` tree probably isn't necessary. At least that would be my expectation because the query below just works as expected:
```sql
SELECT DISTINCT table_schema, table_name FROM region-us.INFORMATION_SCHEMA.TABLES;
```

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
+        expression: str = self.build_expression_sql(matches.expression)
+        return f"REGEXP_CONTAINS({expression}, r'{matches.regex_pattern}')"
+
+    def supports_varchar_length(self) -> bool:
+        return False
+
+    def quote_column(self, column_name: str) -> str:
+        return self.quote_default(column_name)
```
mivds: Shouldn't the base class already use `self.quote_default`?

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def sql_expr_timestamp_literal(self, datetime_in_iso8601: str) -> str:
+        return f"timestamp('{datetime_in_iso8601}')"
+
+    def sql_expr_timestamp_truncate_day(self, timestamp_literal: str) -> str:
+        return f"date_trunc(timestamp({timestamp_literal}), day)"
+
+    def sql_expr_timestamp_add_day(self, timestamp_literal: str) -> str:
+        return f"{timestamp_literal} + interval 1 day"
```
mivds, line 90: Why don't these functions also use AST expressions? I expect the current approach won't scale well. What if we want different units of time? Multiples of the unit, etc?

Something like `ADD(TIMESTAMP("ts"), INTERVAL(1, DAY))` seems more in line with the other syntax.

mivds (commented):
Still reviewing, but already sharing some preliminary feedback for the planned retro session.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: This was indeed leftover from debugging. Removed it!

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: As we discussed together, We just put a check in the SELECT expression that warns the user for potentially unexpected behavior.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: This is what's currently there, just adapted for BigQuery. 
Would currently consider that out of scope for this PR as it would require that we change and verify this for all other implemented datasources.

m1n0 (commented):
I can't approve a PR that I originally opened, but I checked the follow up after our sync and this all LGTM!

Niels-b:
As discussed in the meeting this morning; moved some files around to create better structure for the data sources.
Currently this is done for all data sources except Databricks (seems to fail tests locally and is not running in CI).

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
mivds: From what I saw this is testing-only functionality currently, so seems fair to put it out of scope. Downside is that it means you'll be implementing this function for all the data sources, and probably later we'll have to redo it the 'proper' way. Seems like an acceptable price in this case.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
```diff
@@ -0,0 +1,147 @@
...
+    def _create_connection(
+        self,
+        config: BigQueryConnectionProperties,
+    ):
+        self._load_project_id_and_credentials(config)
+        self._load_optional_impersonated_credentials(config)
+        self._apply_optional_params(config)
+
+        client_info = ClientInfo(
+            user_agent="soda-library",
```
mivds: Shouldn't this be set to `soda-core`?

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
```diff
@@ -0,0 +1,147 @@
...
+            client_info=client_info,
+            location=config.location,
+            client_options=self.client_options,
+        )
+
+        return dbapi.Connection(self.client)
+
+    def format_rows(self, rows: list[Row]) -> list[tuple]:
+        formatted_rows = [tuple(r.values()) for r in rows]
+        return formatted_rows
```
mivds, line 147: Why was this necessary? Can't the `Row` object be treated as if it is a `tuple`? From a quick look at the `Row` class it appears to support access by index the same way a tuple would.

I imagine this can in some cases be a pretty expensive operation to duplicate all the queried data

soda-bigquery/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+from soda_core.common.logging_configuration import configure_logging
+
+
+def pytest_sessionstart(session) -> None:
+    configure_logging(verbose=True)
```
mivds, line 5: Should this live inside the `soda-bigquery` tests or at a higher level?

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -1,16 +1,65 @@
...
 
 @dataclass
-class SELECT:
+class SELECT(BaseSqlExpression):
     fields: SqlExpression | str | list[SqlExpression | str]
 
+    def __init__(self, fields: SqlExpression | str | list[SqlExpression | str]):
+        super().__init__()
+        self.fields = fields
+        self.handle_parent_node_update(fields)
```
mivds: I think [`__post_init__`](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__) might be a cleaner solution for these update routines, as now you're having to manually initialize all the fields in every class.

mivds (approved):
LGTM, left a few smaller comments

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
Niels-b: The way BigQuery returns the rows is slightly different. We need to convert them. Just returning the rows (like the default implementation in `data_source_connection.py` crashes almost every test. The Row seems to contain some additional data for column/index info that the rest of the soda-core library is not expecting. So we need to extract the relevant information (`values()`) from it. See screenshot for the difference (using debugger).
image

soda-bigquery/tests/conftest.py
Niels-b: Copied over from other datasources. They all have this. Maybe something we can/should change? Will skip this for now as it seems easy/small task to change this later.

soda-core/src/soda_core/common/sql_ast.py
Niels-b: It is indeed a bit of a hassle to initialize all the fields for every class. 
Implemented it using `__post_init__` now.

PR #2338, opened 2025-07-13
Added data source method execute_query_one_by_one

tombaeyens:


tombaeyens:
the method was added in another PR

PR #2342, opened 2025-07-17
Support multiple datasources in cli+api

m1n0:


m1n0:
@tombaeyens @nielsn I am merging this, but you should be aware of this change

soda-core/src/soda_core/cli/handlers/contract.py
```diff
@@ -18,7 +18,7 @@
 def handle_verify_contract(
     contract_file_paths: Optional[list[str]],
     dataset_identifiers: Optional[list[str]],
-    data_source_file_path: Optional[str],
+    data_source_file_paths: Optional[str],
```
nielsn, line 21: List type?

PR #2344, opened 2025-07-22
Add SQL Server v4 support

Niels-b:
Also adds support for Synapse.

Note; still pending changes as discussed in [this](https://sodadata.slack.com/archives/C089R1R13N3/p1753712275999769) Slack Thread (probably will be done in separate branch/PR).

IMPORTANT: Merge https://github.com/sodadata/soda-core/pull/2352 first!


Niels-b:
NOTE: Merge https://github.com/sodadata/soda-core/pull/2352 first!

soda-sqlserver/src/soda_sqlserver/test_helpers/sqlserver_data_source_test_helper.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+
+    def drop_test_schema_if_exists(self) -> None:
+        # First find all the tables in the schema
+        table_names: list[str] = self.query_existing_test_table_names(return_fully_qualified_table_names=True)
+        for fully_qualified_table_name in table_names:
+            table_identifier = f"{fully_qualified_table_name.database_name}.{fully_qualified_table_name.schema_name}.{fully_qualified_table_name.table_name}"
+            self.data_source_impl.execute_update(f"DROP TABLE {table_identifier};")
+        # Drop the schema if we found any tables
+        if len(table_names) > 0:
```
mivds: Why the different approach here? I think we usually just drop the schema if it exists, regardless of any tables inside it.

Here we're dropping the tables and only dropping the schema if there were any tables in it, which seems a kind of weird change. Probably good to add the why as a comment in the code

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
```diff
@@ -0,0 +1,49 @@
...
+        values_sql: str = "\n" + "\nUNION ALL ".join(
+            [self._build_insert_into_values_row_sql(value) for value in insert_into.values]
+        )
+        return values_sql
+
+    def _build_insert_into_values_row_sql(self, values: VALUES_ROW) -> str:
+        values_sql: str = "SELECT " + ", ".join([self.literal(value) for value in values.values])
+        values_sql = values_sql.encode("unicode_escape").decode(
+            "utf-8"
+        )  # This escapes values that contain newlines correctly.
```
mivds: Would it make sense to make this a utility method somewhere? Think I saw this implemented twice (maybe more across the codebase). Not entirely sure in what circumstances it needs to be applied.

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source_connection.py
```diff
@@ -0,0 +1,75 @@
...
+        SynapsePasswordAuth,
+        SynapseActiveDirectoryInteractiveAuthentication,
+        SynapseActiveDirectoryPasswordAuthentication,
+        SynapseActiveDirectoryServicePrincipalAuthentication,
+    ] = Field(..., alias="connection", description="Synapse connection configuration")
+
+
+class SynapseDataSourceConnection(SQLServerDataSourceConnection):
+    def __init__(self, name: str, connection_properties: DataSourceConnectionProperties):
+        super().__init__(name, connection_properties)
```
mivds: Think this override isn't needed here? Suggest just using `pass` as the class body to avoid having to change it everywhere if we were to change the function signature.

soda-tests/src/helpers/data_source_test_helper.py
```diff
@@ -255,7 +268,7 @@ def end_test_session_drop_schema(self) -> None:
         if self.is_cicd:
             self.drop_test_schema_if_exists()
 
-    def query_existing_test_table_names(self):
+    def query_existing_test_table_names(self, return_fully_qualified_table_names: bool = False):
```
mivds: Think it's best to avoid this type of argument that changes what the function returns. It's not visible here because the return type is not hinted, but vscode shows it as

Image

IMO the return type should just be `list[str]`. If we need fully qualified names somewhere, I would introduce a new function `query_existing_test_tables` that returns `list[FullyQualifiedTableName]` and then call that function from this one.

Implementation for this function would then become a simple:
```python
return [
    fully_qualified_test_table_name.table_name
    for fully_qualified_test_table_name in self.query_existing_test_tables()
]
```

soda-synapse/setup.py
```diff
@@ -0,0 +1,25 @@
...
+
+from setuptools import setup
+
+package_name = "soda-synapse"
+package_version = "4.0.0b9"
+description = "Soda Synapse V4"
+
+requires = [
+    f"soda-core=={package_version}",
+    "pyodbc",
```
mivds: Suggest dropping `pyodbc` here. It's already included via `soda-sqlserver`, and it avoids conflicts if we were to add a version requirement for `pyodbc` in future (which we should)

mivds (approved):

soda-sqlserver/src/soda_sqlserver/test_helpers/sqlserver_data_source_test_helper.py
Niels-b: SQL Server does not support `IF EXISTS` statements. So we can either do it with unreadable queries that give trouble on Synapse because it doesn't support that type of query.
So I opted for the more manual approach of checking for existing tables, dropping them, then drop the schema.

Dropping the schema only if we found tables was to check if the schema exists. But for that we can probably use the metadata query. I'll update it!

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
Niels-b: This needs to be applied when we have newlines in the strings. This gives errors with some databases (Bigquery, at least). Can add this as a utility method indeed. Good catch!

soda-tests/src/helpers/data_source_test_helper.py
Niels-b: Very good suggestion!
This is much clearer code indeed.

PR #2346, opened 2025-07-23
Obsl 616 fix bigquery contract generation bug

paulteehan:


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -8,6 +8,23 @@
 from soda_core.common.sql_ast import *
 
 
+class DBDataType:
+    """
+    DBDataTypes contains data source-neutral constants for referring to the basic, common column data types.
+    """
+
+    __test__ = False
```
mivds: What's this thing for?

m1n0 (approved):
Awesome cleanup!

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -17,6 +34,33 @@ class SqlDialect:
...
+        else:
+            # if db engine doesn't support varchar length, probably no need to override this method
+            return self.get_contract_type_dict()[DBDataType.TEXT]  # e.g. VARCHAR
+
+    def get_sql_type_dict(self) -> dict[str, str]:
+        """Data type that is used in the create table statement.
+        This can include the length of the column, e.g. VARCHAR(255)"""
+        return self.get_contract_type_dict()
+
+    def get_contract_type_dict(self) -> dict[str, str]:
```
mivds, line 48: Don't need to act on it right now, but just wondering if there will ever be a difference between these two methods? If not we could probably merge them?

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -62,6 +63,22 @@ def create_metadata_columns_query(self) -> MetadataColumnsQuery:
 class BigQuerySqlDialect(SqlDialect):
     DEFAULT_QUOTE_CHAR = "`"
 
+    def text_col_type(self, length: Optional[int] = 255) -> str:
+        """No length in BigQuery"""
+        return "STRING"
```
mivds: Looking at the `SqlDialect` base class, I would think this function isn't needed. It's already handled in the base implementation using `supports_varchar_length()`

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
```diff
@@ -34,3 +34,8 @@ def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
     def create_schema_if_not_exists_sql(self, schema_name: str) -> str:
         quoted_schema_name: str = self.quote_default(schema_name)
         return f"CREATE SCHEMA IF NOT EXISTS {quoted_schema_name} AUTHORIZATION CURRENT_USER;"
+
+    def get_sql_type_dict(self) -> dict[str, str]:
+        base_dict = super().get_sql_type_dict()
+        base_dict[DBDataType.TEXT] = "character varying(255)"
```
mivds: This looks weird. We're specifying a character count here, but I would think that doesn't play well with the `text_col_type` function?

Implementation below for reference:
```python
    def text_col_type(self, length: Optional[int] = 255) -> str:
        """Get the column type specificier for a variable length text column of a specific length."""
        if self.supports_varchar_length():
            return self.get_contract_type_dict()[DBDataType.TEXT] + f"({length})"  # e.g. VARCHAR(255)
        else:
            # if db engine doesn't support varchar length, probably no need to override this method
            return self.get_contract_type_dict()[DBDataType.TEXT]  # e.g. VARCHAR
```

soda-tests/tests/features/test_schema_check.py
```diff
@@ -56,28 +76,41 @@ def test_schema_errors(data_source_test_helper: DataSourceTestHelper):
             columns:
               - name: id
                 data_type: {test_table.data_type('id')}
-                character_maximum_length: 512
+                {get_character_maximum_length_expression(data_source_test_helper, 512, overwrite_with_expected_length=True)}
```
mivds, line 79: Nit: it might make more sense to move this to a separate test, then you can just skip the test for databases that don't support string lengths. Not needed to address, but maybe a future thing

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Yes, there is a difference. The contract type is for contracts. The sql type is for table creation (see function docs).

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
Niels-b: As discussed per slack message; leave it like this for now. Can always come back later and change it.
Currently seems like the way to go while new data sources are still being added.

soda-tests/tests/features/test_schema_check.py
Niels-b: Fully agree! We played with the idea of skipping the test if the datasource does not support varchar lengths. But it was a very relevant test for other things as well. So decided just to make it work for now :)

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Removed this. That was for the old `TestDataType` class. `__test__` is used to tell pytest to ignore this class ;)

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: Indeed, good catch! Removed it.

PR #2349, opened 2025-07-25
Oracle support

paulteehan:
Oracle implementation in soda-core.  Tests passing locally, I think CI may not be configured yet.  Ready for review.

Still pending:
- Update soda-library to match connection YAML
- Add support and test in soda-extensions

### DIFF
--- a/soda-tests/setup.py
+++ b/soda-tests/setup.py
@@ -13,10 +13,7 @@ package_name = "soda-tests"
 package_version = "4.0.0b10"
 description = "Soda Core V4 Tests"
 
-requires = [
-    f"soda-core=={package_version}",
-    "pytz"
-]
+requires = [f"soda-core=={package_version}", "pytz"]
 
 setup(
     name=package_name,


soda-core/src/soda_core/common/statements/metadata_columns_query.py
```diff
@@ -64,8 +64,12 @@ def build_sql(self, dataset_prefix: Optional[list[str]], dataset_name: str) -> O
...
                 FROM(self.sql_dialect.table_columns()).IN(
                     [
-                        *prefixes,
-                        self.sql_dialect.schema_information_schema(),
+                        id
+                        for id in [  # prefixes can be None if information schema is top-level, e.g. Oracle
+                            *prefixes,
+                            self.sql_dialect.schema_information_schema(),
+                        ]
+                        if id
```
mivds: Hmm, wondering if we can't resolve this by passing `prefixes=[]` to `MetadataColumnsQuery.__init__`? Then we use the already available interface instead of adding hidden logic here

soda-core/src/soda_core/common/statements/metadata_tables_query.py
```diff
@@ -63,15 +63,23 @@ def build_sql_statement(
...
                 self.sql_dialect.table_tables(),
                 table_prefix=[
-                    *prefixes,
-                    self.sql_dialect.schema_information_schema(),
+                    prefix
+                    for prefix in [
+                        *prefixes,  # this could be None for some data sources e.g. Oracle
+                        self.sql_dialect.schema_information_schema(),
+                    ]
+                    if prefix
```
mivds: Same question as above for `MetadataColumnsQuery`

soda-core/src/soda_core/common/statements/metadata_tables_query.py
```diff
@@ -63,15 +63,23 @@ def build_sql_statement(
...
                 ],
             ),
             SELECT(
                 [
-                    self.sql_dialect.column_table_catalog(),
-                    self.sql_dialect.column_table_schema(),
-                    self.sql_dialect.column_table_name(),
+                    col
+                    for col in [
+                        self.sql_dialect.column_table_catalog(),  # this could be None for some data sources e.g. Oracle
```
mivds: Is it an option to have this function return `"NULL"`, i.e. `SELECT NULL, schema_name, table_name FROM ...`? That way we don't need to add conditionals here or in the result processing.

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -45,6 +50,7 @@ def open_connection(self) -> None:
             try:
                 logger.debug(f"'{self.name}' connection properties: {self.connection_properties}")
                 self.connection = self._create_connection(self.connection_properties)
+                self._post_connection_hook()
```
mivds: Do we need the `_post_connection_hook` here? I would think we can just add that functionality at the end of `_create_connection`?

soda-core/src/soda_core/common/datetime_conversions.py
```diff
@@ -5,6 +5,7 @@
 
 def convert_datetime_to_str(dt: datetime) -> Optional[str]:
     try:
+        # Attention, if dt does not have timezone info, Python will assume dt is in the local system timezone (e.g. UTC+1)
         return dt.astimezone(timezone.utc).isoformat(timespec="seconds")
```
mivds, line 9: Nice find with that comment. Out of scope for this PR, but I suspect this may not be the behavior we want.

Probably want to do something like this instead:
```python
if dt.tzinfo is None:
  dt = dt.replace(tzinfo=timezone.utc)
return dt.isoformat(timespec="seconds")
```

That would interpret the timestamp as UTC if no timezone is provided (which is more reproducible than using the local timezone). Then formats to string including timezone, which may be UTC or a previously set timezone

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -240,7 +240,7 @@ def __post_init__(self):
 
 @dataclass
 class COLUMN(SqlExpression):
-    name: str
+    name: SqlExpression | str  # Modified to allow SQL expressions, e.g. if you need to use a case statement to generate a column during SELECT
```
mivds: Nit: I would remove the `Modified to` prefix. That makes sense in context of PR reviewing, but once merged it should just be `Allow SQL expressions, e.g. if you need to use a case statement to generate a column during SELECT`

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -344,7 +344,9 @@ def build_expression_sql(self, expression: SqlExpression | str | Number) -> str:
 
     def _build_column_sql(self, column: COLUMN) -> str:
         table_alias_sql: str = f"{self.quote_default(column.table_alias)}." if column.table_alias else ""
-        column_sql: str = self.quote_default(column.name)
+        column_sql: str = self.build_expression_sql(
+            column.name
+        )  # Now supports any expression.  If string is passed, will default to old quote_default behavior
```
mivds: Same feedback as above, should be written in context of the long term. `Now` and `old` only make sense in context of the PR

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -618,6 +623,9 @@ def get_database_prefix_index(self) -> int | None:
     def get_schema_prefix_index(self) -> int | None:
         return 1
 
+    def sql_expr_timestamp_with_tz_literal(self, datetime_in_iso8601: str) -> str:
+        return self.sql_expr_timestamp_literal(datetime_in_iso8601)
```
mivds: Hmm, not sure what the best approach is here. I fear this function as-is will be hard to use because you will always need to be aware of whether the datetime you have contains a timezone or not, and call the appropriate function. We could have a single function that detects if timezone is present and then acts accordingly. On the other hand that may lead to bugs because it works differently than expected.

I guess let's keep it as you've implemented it for now. We'll find out later on if there are places where we don't know which function to call.

soda-tests/src/helpers/data_source_test_helper.py
```diff
@@ -66,6 +66,14 @@ def create(cls, test_datasource: str) -> DataSourceTestHelper:
             )
 
             return BigQueryDataSourceTestHelper()
+
+        elif test_datasource == "oracle":
+            from soda_oracle.test_helpers.oracle_data_source_test_helper import (
+                OracleDataSourceTestHelper,
+            )
+
+            return OracleDataSourceTestHelper()
```
mivds, line 75: Probably out of scope here, but think it would be good to also have a plugin mechanism for this. Having to add a helper here when we create a new extension data source isn't great

soda-tests/tests/features/test_schema_check.py
```diff
@@ -109,7 +109,8 @@ def test_schema_errors(data_source_test_helper: DataSourceTestHelper):
         assert varchar(512) == length_mismatch.get_expected()
 
     type_mismatch = schema_check_result.column_data_type_mismatches[index_of_type_mismatch]
-    assert data_type_map[DBDataType.DATE] == type_mismatch.get_actual()
+    expected_data_type = data_type_map[DBDataType.DATE]
+    assert expected_data_type == type_mismatch.get_actual()
```
mivds: Nit: the original seems more readable to me here, especially since we have another line below this using different syntax that also refers to an 'expected' data type

mivds (commented):
Looks pretty good! Noticed a few odd quirks though; I left a few questions/suggestions

soda-core/src/soda_core/common/datetime_conversions.py
Niels-b: Very big fan of this! Ran into issues with the "auto-conversion" that happens here.
With all the timezone stuff we're doing for this PR and for Oracle, I think we can definitely change it (if it doesn't break other things!!)

PR #2350, opened 2025-07-25
soda connect to trino using a self-signed certificate

sumit-gupta-sgt:
Hi everyone! I'm trying to connect soda to trino but using username and password and a self-signed certificate.

It does not work because there is no parameter "verify" in soda-core-trino [1] that can be passed to python-trino to address this situation [2].


sumit-gupta-sgt:
Hi @mivds 

Could you please review this as well when you get a chance?

Appreciate your time and support!

Thanks,
Sumit

PR #2351, opened 2025-07-28
Added query one by one on data source connection

tombaeyens:
This is a commit needed for an extensions feature: https://github.com/sodadata/soda-extensions/pull/16

### DIFF
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Optional, Callable
+from typing import Callable, Optional
 
 from soda_core.common.data_source_results import QueryResult, UpdateResult
 from soda_core.common.logging_constants import soda_logger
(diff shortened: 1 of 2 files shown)


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -102,6 +102,20 @@ def execute_update(self, sql: str) -> UpdateResult:
...
+        usage: execute_query_one_by_one("SELECT ...", lambda row, description: your_handle_row(row))
+        """
+        # noinspection PyUnresolvedReferences
+        cursor = self.connection.cursor()
+        try:
+            logger.debug(f"SQL query fetch one-by-one: \n{sql}")
+            cursor.execute(sql)
+            for row in cursor:
+                row_callback(row, cursor.description)
+        finally:
```
nielsn: You can get rid of the explicit error handling by using a context manager:

```python
with self.connection.cursor() as cursor:
   cursor.execute(sql)
   for row in cursor:
     row_callback(row, cursor.description)
 ```

tombaeyens: Super tip.  Tnx.  Updated.

PR #2352, opened 2025-07-29
Add new sql_ast statements: INSERT INTO, CREATE TABLE and DROP TABLE

Niels-b:
Co-authored by @paulteehan 

Added the above mentioned statements to provide a unified way of creating queries that contain these. Also allows for easier overloading in new data sources.

### DIFF
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -480,7 +480,7 @@ class CREATE_TABLE_COLUMN(BaseSqlExpression):
 class INSERT_INTO(BaseSqlExpression):
     fully_qualified_table_name: str
     values: list[VALUES]
-    columns: Optional[list[SqlExpression | str]] = None
+    columns: Optional[list[COLUMN]] = None
 
     def __post_init__(self):
         super().__post_init__()
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -466,10 +466,14 @@ class CREATE_TABLE_IF_NOT_EXISTS(CREATE_TABLE):
 @dataclass
 class CREATE_TABLE_COLUMN(BaseSqlExpression):
     name: str
-    type: DBDataType
+    type: DBDataType | str
     length: Optional[int] = None
     nullable: Optional[bool] = None
     default: Optional[SqlExpression | str] = None
+    # If True, the type will be looked up in the contract type dict.
+    # This allows for our custom DBDataTypes to be used in the CREATE TABLE statement :).
+    # We have to do this, because in soda-extensions we sometimes get data types from the metadata query. We don't want to convert them into DBDataTypes, instead we use them as is.
+    do_type_lookup: bool = False
 
     def __post_init__(self):
         super().__post_init__()
@@ -479,7 +483,7 @@ class CREATE_TABLE_COLUMN(BaseSqlExpression):
 @dataclass
 class INSERT_INTO(BaseSqlExpression):
     fully_qualified_table_name: str
-    values: list[VALUES]
+    values: list[VALUES_ROW]
     columns: Optional[list[COLUMN]] = None
 
     def __post_init__(self):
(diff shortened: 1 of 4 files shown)


soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -444,3 +445,81 @@ def __post_init__(self):
...
+class CREATE_TABLE_COLUMN(BaseSqlExpression):
+    name: str
+    type: DBDataType | str
+    length: Optional[int] = None
+    nullable: Optional[bool] = None
+    default: Optional[SqlExpression | str] = None
+    # If True, the type will be looked up in the contract type dict.
+    # This allows for our custom DBDataTypes to be used in the CREATE TABLE statement :).
+    # We have to do this, because in soda-extensions we sometimes get data types from the metadata query. We don't want to convert them into DBDataTypes, instead we use them as is.
+    do_type_lookup: bool = False
```
mivds: Can't we use `isinstance(type, DBDataType)` for this?

mivds: Ah.. Just looked at the definition of the `DBDataType`. It's just a glorified namespace, so that won't work.

I would suggest refactoring a bit further, to avoid the overhead of having to set this property correctly everywhere. If we convert the `DBDataType` to an enum, we should be able to use `isinstance`.

```python
class DBDataType:
```

becomes

```python
class DBDataType(enum.Enum):
```

I think that will be enough

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -444,3 +445,81 @@ def __post_init__(self):
...
+    fully_qualified_table_name: str
+    columns: list[CREATE_TABLE_COLUMN]
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.handle_parent_node_update(self.columns)
+
+
+@dataclass
+class CREATE_TABLE_IF_NOT_EXISTS(CREATE_TABLE):
```
mivds, line 461: No real opinion, just a question: should this be a separate expression or just a flag on the `CREATE_TABLE` expression?

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -143,6 +131,102 @@ def create_schema_if_not_exists_sql(self, schema_name: str) -> str:
         quoted_schema_name: str = self.quote_default(schema_name)
         return f"CREATE SCHEMA IF NOT EXISTS {quoted_schema_name};"
 
+    #########################################################
+    # CREATE TABLE
+    #########################################################
+    def build_create_table_sql(
+        self, create_table: CREATE_TABLE | CREATE_TABLE_IF_NOT_EXISTS, add_semicolon: bool = True
```
mivds, line 138: Are there cases where you don't want a semicolon terminator? I can't really think of any use cases. You can't do something like `CREATE TABLE (...) UNION ...` as far as I know

mivds (approved):
LGTM, just a few questions/suggestions

soda-core/src/soda_core/common/sql_ast.py
Niels-b: Needed to add a few more things to make it work. 
We tried to achieve this at first, but didn't think of using an `Enum` here. Much cleaner than the flag indeed! I like the idea :)

soda-core/src/soda_core/common/sql_ast.py
Niels-b: I think this should be a separate expression. That makes it more consistent with the rest of the AST.

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: According to @paulteehan , we need to be able to overwrite the semicolon for Oracle. So we added it just to be safe.

paulteehan: Indeed, I'm overriding this method and setting add_semicolon to False to make it work for Oracle

Niels-b:
Discussed with Tom and Paul: refactor of the entire AST inbound, to make it more like an AST to avoid surprises in the code. But we will continue with what we have for now. So we got the all-clear to merge.

PR #2355, opened 2025-07-31
Fixes for SQL AST to make it compatible for SQLServer and Synapse

Niels-b:
Noticed some issues while implementing PR feedback for soda-extensions. This PR fixes those issues.
This includes:
- INSERT AST now requires the columns to be specified! 
- Extra testing for the AST `CREATE`, `INSERT`, `DROP` statements. Now we verify all these statements on every datasource.

### DIFF
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -475,15 +475,17 @@ class CREATE_TABLE_COLUMN(BaseSqlExpression):
         super().__post_init__()
         self.handle_parent_node_update(self.default)
 
+    def convert_to_standard_column(
+        self, table_alias: Optional[str] = None, field_alias: Optional[str] = None
+    ) -> COLUMN:
+        return COLUMN(name=self.name, table_alias=table_alias, field_alias=field_alias)
+
 
 @dataclass
 class INSERT_INTO(BaseSqlExpression):
     fully_qualified_table_name: str
     values: list[VALUES_ROW]
-    columns: Optional[
-        list[COLUMN]
-    ] = None  # For SQLServer and Synapse, if the columns are not specified, the metadata query is used to determine the columns.
-    # The order of values that is inserted should be the ordinal position of the columns!
+    columns: list[COLUMN]  # The order of values that is inserted should be the ordinal position of the columns!
 
     def __post_init__(self):
         super().__post_init__()
@@ -495,10 +497,7 @@ class INSERT_INTO(BaseSqlExpression):
 class INSERT_INTO_VIA_SELECT(BaseSqlExpression):
     fully_qualified_table_name: str
     select_elements: list[SqlExpression | str]
-    columns: Optional[
-        list[COLUMN]
-    ] = None  # For SQLServer and Synapse, if the columns are not specified, the metadata query is used to determine the columns.
-    # The order of values that is inserted should be the ordinal position of the columns!
+    columns: list[COLUMN]  # The order of values that is inserted should be the ordinal position of the columns!
 
     def __post_init__(self):
         super().__post_init__()
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -470,10 +470,6 @@ class CREATE_TABLE_COLUMN(BaseSqlExpression):
     length: Optional[int] = None
     nullable: Optional[bool] = None
     default: Optional[SqlExpression | str] = None
-    # If True, the type will be looked up in the contract type dict.
-    # This allows for our custom DBDataTypes to be used in the CREATE TABLE statement :).
-    # We have to do this, because in soda-extensions we sometimes get data types from the metadata query. We don't want to convert them into DBDataTypes, instead we use them as is.
-    # do_type_lookup: bool = False
 
     def __post_init__(self):
         super().__post_init__()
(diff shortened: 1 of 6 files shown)


soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -484,7 +480,10 @@ def __post_init__(self):
 class INSERT_INTO(BaseSqlExpression):
     fully_qualified_table_name: str
     values: list[VALUES_ROW]
-    columns: Optional[list[COLUMN]] = None
+    columns: Optional[
+        list[COLUMN]
+    ] = None  # For SQLServer and Synapse, if the columns are not specified, the metadata query is used to determine the columns.
+    # The order of values that is inserted should be the ordinal position of the columns!
```
mivds: Is the comment necessary here? Seems weird if we're moving those databases to live in commercial extensions

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
```diff
@@ -70,6 +82,42 @@ def _build_create_table_statement_sql(self, create_table: CREATE_TABLE | CREATE_
...
 
+    def _build_insert_into_columns_sql(self, insert_into: INSERT_INTO) -> str:
+        columns = insert_into.columns
+        if columns is None:
+            columns = self.__get_columns_for_table(insert_into.fully_qualified_table_name)
+        columns_sql: str = " (" + ", ".join([self.build_expression_sql(column) for column in columns]) + ")"
+        return columns_sql
+
+    def __get_columns_for_table(self, table_name: str) -> list[COLUMN]:
+        # We do not have access to the DataSourceImpl here, so we cannot use the metadata query. (TODO: do we move this creation to the SQLDialect?)
```
mivds: Outdated comment? I see that the `DataSourceImpl` is now passed in as an argument (which seems like a bad idea)

Niels-b:
As discussed per meeting: I had implemented the "must always specify columns" path, just wanted to make it easier to use for other datasources.
But I agree that always having to specify the columns is better in the long run to avoid strange bugs. 
Still to discuss with Tom.

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -470,21 +470,22 @@ class CREATE_TABLE_COLUMN(BaseSqlExpression):
...
+    ) -> COLUMN:
+        return COLUMN(name=self.name, table_alias=table_alias, field_alias=field_alias)
+
 
 @dataclass
 class INSERT_INTO(BaseSqlExpression):
     fully_qualified_table_name: str
     values: list[VALUES_ROW]
-    columns: Optional[list[COLUMN]] = None
+    columns: list[COLUMN]  # The order of values that is inserted should be the ordinal position of the columns!
```
mivds: Is the comment still relevant here? I would think you can specify columns/values in whatever order you like, as long as they are both in the same order

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -496,10 +497,12 @@ def __post_init__(self):
 class INSERT_INTO_VIA_SELECT(BaseSqlExpression):
     fully_qualified_table_name: str
     select_elements: list[SqlExpression | str]
+    columns: list[COLUMN]  # The order of values that is inserted should be the ordinal position of the columns!
```
mivds: Same question here whether this comment is still relevant

soda-tests/tests/components/test_sql_generation.py
```diff
@@ -145,15 +129,45 @@ def test_sql_ast_insert_into_with_datetimes():
...
 
 
+def test_sql_ast_insert_into_via_select():
+    sql_dialect: SqlDialect = SqlDialect()
+    my_insert_into_statement = sql_dialect.build_insert_into_via_select_sql(
+        INSERT_INTO_VIA_SELECT(
+            fully_qualified_table_name='"customers"',
+            select_elements=[
+                SELECT(COLUMN("name").IN("C")),
+                FROM("customers").AS("C").IN(["sdb", "mschm"]),
```
mivds, line 149: Not really related to the scope of this PR, just sharing: I found the use of `.IN` quite confusing in these 2 lines. Wasn't clear to me what it does until I looked at the SQL output

soda-tests/tests/features/test_sql_ast_inserts.py
```diff
@@ -0,0 +1,121 @@
...
+    assert result.rows[2][2] is None
+
+    assert result.rows[0][3] == datetime.date(2021, 1, 1)
+    assert result.rows[1][3] == datetime.date(2021, 1, 2)
+    assert result.rows[2][3] is None
+
+    # Then drop the table to clean up
+    # We explicitly do not use the "if exists" variant, because the table should exist at this point.
+    drop_table_sql = sql_dialect.build_drop_table_sql(DROP_TABLE(fully_qualified_table_name=my_table_name))
+    data_source_impl.execute_update(drop_table_sql)
```
mivds: Is cleanup necessary? Would expect it's not a problem to just leave it there. If it would be a problem, we may need to a `try: ... finally:` to ensure it gets removed even when the test fails

soda-core/src/soda_core/common/sql_ast.py
Niels-b: Ah, right. Didn't catch that. Think it's still relevant to comment that the list of columns should be in the same order as the values.

soda-tests/tests/components/test_sql_generation.py
Niels-b: Yes, I had the same thing. This is just the way the AST is coded now and that might change in the future.

soda-tests/tests/features/test_sql_ast_inserts.py
Niels-b: Cleanup is preferred here I think. 
I've added the `finally:` .

PR #2358, opened 2025-08-01
Migrate data sources from core to extensions

paulteehan:


paulteehan:
I'm pretty sure we can delete this since we decided to keep sqlserver and synapse in core

PR #2365, opened 2025-08-06
Add MS Fabric support to soda-core

Niels-b:


soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source_connection.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+CONTEXT_AUTHENTICATION_DESCRIPTION = "Use context authentication"
+
+
+# All of these classes are just copies of the SQLServerConnectionProperties classes, but with the Synapse type
+class FabricConnectionProperties(SqlServerConnectionProperties, ABC):
+    autocommit: Optional[bool] = Field(
+        True, description="Whether to use autocommit"
+    )  # Synapse requires autocommit to be True.
```
mivds: Some copy pasta renaming left here: `Synapse` -> `Fabric`.

Also, should the `autocommit` setting be exposed here? It's defaulting to `True`, do we want to allow users to override that (assuming it's the right setting for Fabric)?

soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source_connection.py
```diff
@@ -0,0 +1,94 @@
...
+        tup[4],
+        tup[5],
+        tup[6] // 1000,
+        timezone.utc,  # Timezone is not supported by Fabric, so we assume that all timestamps are in UTC (i.e. we do no conversion!)
+    )
+
+
+class FabricDataSourceConnection(SqlServerDataSourceConnection):
+    def _get_autocommit_setting(self) -> bool:
+        return True  # Synapse requires autocommit to be True.
```
mivds: Synapse -> Fabric

PR #2368, opened 2025-08-08
Add insert script to upload bus dataset to test datasources

Niels-b:
Script that was used to upload the bus breakdown dataset to multiple datasources using the existing connections that we created. So it has done it job quite well already.
Seems to work okay for most datasets, but might require some work here and there if you run it on a specific datasource. Tried to mitigate that by introducing batch sizes based on the datasource.
Test is configured by default to be skipped (as we don't have the csv file).

### DIFF
--- a/scripts/upload_bus_dataset.sh
+++ b/scripts/upload_bus_dataset.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+# Run this from the root project dir with scripts/upload_bus_dataset.sh
+# Please look at the test file below and adjust the CSV_FILE_LOCATION and SCHEMA_NAME if needed.
+# Additionally, you will have to uncomment the @pytest.mark.skip decorator.
+
+# Also make sure you are on the correct venv before running this script.
+
+pytest soda-tests/tests/features/test_sql_upload_bus_breakdown.py::test_insert_bus_breakdown_dataset -o log_cli=true -o log_cli_level=DEBUG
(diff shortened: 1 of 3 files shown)


mivds (approved):
Assuming you want to merge this, maybe more suited as a script under `scripts`? Think we'll be more likely to find it there than as a feature test

Niels-b:
> Assuming you want to merge this, maybe more suited as a script under `scripts`? Think we'll be more likely to find it there than as a feature test

@mivds I added a `.sh` script that calls the test. This should provide more visibility, without having to move and change the code because it makes use of the testing framework in core. Mostly for internal use anyway to provide data for testing. And can be adopted for other datasets if needed.
Would like to merge this so we don't run into issues later again with rebasing/merging to make use of new datasources and provide everyone with access to this ;).
WDYT? Is this okay?

mivds:
> > Assuming you want to merge this, maybe more suited as a script under `scripts`? Think we'll be more likely to find it there than as a feature test
> 
> @mivds I added a `.sh` script that calls the test. This should provide more visibility, without having to move and change the code because it makes use of the testing framework in core. Mostly for internal use anyway to provide data for testing. And can be adopted for other datasets if needed. Would like to merge this so we don't run into issues later again with rebasing/merging to make use of new datasources and provide everyone with access to this ;). WDYT? Is this okay?

Yes for sure. Thanks for the update!

PR #2370, opened 2025-08-08
Obsl 592 core add redshift support

paulteehan:


soda-core/src/soda_core/common/aws_credentials.py
```diff
@@ -0,0 +1,65 @@
+from typing import Optional
+
+import boto3
+
+
+class AwsCredentials:
```
paulteehan, line 6: Lifted directly from soda-library

soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source_connection.py
```diff
@@ -0,0 +1,126 @@
...
+
+        client = boto3.client(
+            "redshift",
+            region_name=resolved_aws_credentials.region_name,
+            aws_access_key_id=resolved_aws_credentials.access_key_id,
+            aws_secret_access_key=resolved_aws_credentials.secret_access_key,
+            aws_session_token=resolved_aws_credentials.session_token,
+        )
+
+        cluster_name = self.host.split(".")[0]
```
mivds: This seems a little weird. What does a redshift hostname look like?

The description for the `host` field above shows both a hostname and IP address are allowed. I don't think this can work when an IP address is provided. Also not sure what would happen if there is a `https://` or similar prefix.
```python
host: Union[str, IPvAnyAddress] = Field(..., description="Database host (hostname or IP address)")
```

soda-redshift/tests/data_sources/test_redshift.py
```diff
@@ -0,0 +1,157 @@
...
+
+import pytest
+from helpers.test_connection import TestConnection
+
+# define environment variables used in test cases
+REDSHIFT_HOST = os.getenv("REDSHIFT_HOST", "")
+REDSHIFT_PORT = os.getenv("REDSHIFT_PORT", None)
+REDSHIFT_DATABASE = os.getenv("REDSHIFT_DATABASE", "soda_test")
+REDSHIFT_USERNAME = os.getenv("REDSHIFT_USERNAME", "")
+REDSHIFT_PASSWORD = os.getenv("REDSHIFT_PASSWORD", "")
```
mivds, line 11: Nit: would be nice if we can avoid defining this twice. It's currently both here and in `soda-redshift/src/soda_redshift/test_helpers/redshift_data_source_test_helper.py`. Not immediately sure how though, imports will likely be annoying. Feel free to ignore.

mivds (commented):
Generally looks good, though I think that `cluster_name` thing is likely to fail in some cases.

soda-redshift/tests/data_sources/test_redshift.py
paulteehan: Ignoring for now though I do agree with you

soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source_connection.py
paulteehan: Fixed

PR #2372, opened 2025-08-13
Athena support

Niels-b:
This PR includes Athena support for soda-core.

Additionally, I've changed the code that ensures the test tables to use the AST for `CREATE_TABLE` statements. This way we don't have to implement everything twice.


soda-core/src/soda_core/common/datetime_conversions.py
```diff
@@ -5,8 +5,19 @@
 
 def convert_datetime_to_str(dt: datetime) -> Optional[str]:
```
Niels-b, line 6: This function gave issues with other DBs as well. 
Changed it so the timezone is now set accordingly (see comments)

soda-tests/tests/features/test_freshness_check.py
```diff
@@ -156,7 +156,10 @@ def test_freshness_no_rows(data_source_test_helper: DataSourceTestHelper):
         check_result: FreshnessCheckResult = contract_verification_result_t1.check_results[0]
         assert check_result.max_timestamp is None
         assert check_result.max_timestamp_utc is None
-        assert str(check_result.data_timestamp) == "2025-01-04 10:00:00+00:00"
+        assert str(check_result.data_timestamp) in [
+            "2025-01-04 10:00:00+00:00",
+            "2025-01-04 10:00:00",
```
Niels-b, line 161: We're not checking for a timestamp here, so if it's not included it's okay. But someone might have a different opinion, so I'm mentioning/highlighting it ;)

PR #2374, opened 2025-08-14
Move request CLI from extensions to OSS

LaurenDebruyn:
Moving request resource to OSS to make it available to design partners.
This move has been aligned with leadership.


LaurenDebruyn:
> oh wait, I just realized - this is the complete implementation correct? AFAIK some of the commercial commands are part of OSS but just the commands themselves, the implementation does not necessarily work - is this 100% clear with the team that request cli will be fully OSS, including the whole implementation?

We want the functionality to be available to design partners without having to go through installing the extensions since it's not stable yet.

PR #2379, opened 2025-08-18
Disable identities if `blake2b` is not available

### LINKED ISSUE #2354: Soda Core fails in FIPS-enabled environments due to use of hashlib.blake2b
### Summary

Soda Core fails to run in **FIPS-enabled environments** due to the use of `hashlib.blake2b`, which is not FIPS 140-2 compliant and is therefore disabled in these environments.

### Error

When attempting to use Soda Core with FIPS mode enabled (e.g., on hardened Linux systems), the following error is thrown:

TypeError: 'digest_size' is an invalid keyword argument for this function

This happens because `blake2b` is unavailable in Python under FIPS mode.

### Affected Code

The use of `hashlib.blake2b` appears in multiple places in the codebase for purposes like hashing identifiers or computing fingerprints.

### Proposed Solution

Introduce a utility function such as `fips_safe_hash()` that:

- Uses `hashlib.blake2b()` when available.
- Falls back to `hashlib.sha256()` with truncation when in FIPS mode.

```python
import hashlib

def fips_safe_hash(data: bytes, digest_size=32) -> bytes:
    try:
        return hashlib.blake2b(data, digest_size=digest_size).digest()
    except (TypeError, ValueError, AttributeError):
        return hashlib.sha256(data).digest()[:digest_size]

Then replace all hashlib.blake2b(...) calls with this wrapper to ensure compatibility in both FIPS and non-FIPS environments.

### PR DESCRIPTION
**Pull request by mivds**
Resolves #2354

### DIFF
--- a/soda/core/soda/execution/identity.py
+++ b/soda/core/soda/execution/identity.py
@@ -81,9 +81,11 @@ class ConsistentHashBuilder:
         from soda.sodacl.schema_check_cfg import SchemaValidations
         from soda.sodacl.threshold_cfg import ThresholdCfg
 
-        blake2b = self.get_blake2b()
+        if value is None:
+            return
 
-        if value is None or blake2b is None:
+        blake2b = self.get_blake2b()
+        if blake2b is None:
             return
         elif isinstance(value, str):
             blake2b.update(value.encode("utf-8"))

### CODE REVIEW

**Comment by sumit-gupta-sgt** [comment]
Hi @mivds, Thank you for addressing the FIPS compatibility issue in soda-core by removing the digest_size argument from hashlib.blake2b(). This change is crucial for ensuring the library functions correctly in FIPS-enabled environments.

To further enhance compatibility and avoid potential issues with openssl_blake2b, I recommend implementing a fallback mechanism that attempts to use hashlib.blake2b() without the digest_size argument. If that fails, the system should gracefully fall back to hashlib.sha256() or return None if hashing isn't feasible. This approach ensures that the library remains functional across various environments, including those with strict cryptographic standards.

import hashlib
from typing import Optional

def get_identity(data: str, digest_size: int = 16) -> Optional[str]:
    """
    Generate a FIPS-safe identity hash for the given data.
    
    - Uses blake2b with digest_size if supported.
    - Falls back to default blake2b if digest_size not supported.
    - Falls back to SHA256 if blake2b is unavailable (e.g., FIPS mode).
    - Returns None if no hash can be generated.
    """
    encoded_data = data.encode("utf-8")

    # Try blake2b with digest_size
    try:
        return hashlib.blake2b(encoded_data, digest_size=digest_size).hexdigest()
    except (TypeError, ValueError):
        pass  # Likely FIPS/OpenSSL restriction

    # Try default blake2b
    try:
        return hashlib.blake2b(encoded_data).hexdigest()
    except (AttributeError, ValueError):
        pass  # Blake2b unavailable

    # Fallback to SHA256 (FIPS-safe)
    try:
        return hashlib.sha256(encoded_data).hexdigest()
    except Exception:
        pass  # Extremely unlikely

    # Return None if everything fails
    return None

Implementing this fallback mechanism will enhance the robustness of soda-core in diverse environments. Please let me know if you need assistance integrating this solution or if you have any questions.

File: soda/core/soda/execution/identity.py [hunk 1]
```diff
@@ -75,10 +83,14 @@ def add(self, value: Optional[str]):
 
         if value is None:
             return
+
+        blake2b = self.get_blake2b()
+        if blake2b is None:
+            return
```
mark-boyle-sp, line 89: What is the downside to none Identity ?

PR #2380, opened 2025-08-19
Fix bug in data type casting

paulteehan:
this was broken with text DBDataType because `isinstance(cast.to_type, str)` was unexpectedly evaluating as True

### DIFF
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -101,7 +101,7 @@ class SqlDialect:
         }
 
     def get_data_type_type_str(self, db_data_type: DBDataType) -> str:
-        return self.get_contract_type_dict()[db_data_type]
+        return self.get_sql_type_dict()[db_data_type]
 
     def quote_default(self, identifier: Optional[str]) -> Optional[str]:
         return (
@@ -651,7 +651,7 @@ class SqlDialect:
         return f"COALESCE({args})"
 
     def _build_cast_sql(self, cast: CAST) -> str:
-        to_type_text: str = cast.to_type if isinstance(cast.to_type, str) else self.get_data_type_type_str(cast.to_type)
+        to_type_text: str = self.get_data_type_type_str(cast.to_type) if isinstance(cast.to_type, DBDataType) else cast.to_type
         return f"CAST({self.build_expression_sql(cast.expression)} AS {to_type_text})"
 
     def _build_case_when_sql(self, case_when: CASE_WHEN) -> str:


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -101,7 +101,7 @@ def get_contract_type_dict(self) -> dict[str, str]:
         }
 
     def get_data_type_type_str(self, db_data_type: DBDataType) -> str:
-        return self.get_contract_type_dict()[db_data_type]
+        return self.get_sql_type_dict()[db_data_type]
```
tombaeyens, line 104: I have a hard time interpreting these data types and seeing if this is a good or a bad change.
So it's good for me to go.

But the usage here goes beyond the original meaning of contract data types and sql data types.
Not sure if now's the time.  But maybe we should consider splitting it up more.  
* data type used in casts?
* data type in DDL create table statements? (includes length)
* data type as it comes from the column metadata?
* maybe more use cases?

Niels-b: I think this is a good idea. But it's beyond the scope of this PR for me.

Just brainstorming (ideas/input always welcome on this!):
If we were to do this, we'd have to normalize the types **everywhere** in the code (metadata queries, casting, ddl, schema checks,...). 
Potentially with a new class (something similar to `ColumnMetadata`), which includes column name, datatype (something similar to DBDataType -> so already standardized!!) and optional length. So no more `strings` being passed around as arguments  for column types -> it always has to be converted the new `ColumnType` class. This class will have to include some form of `__eq__`, but that is open for discussion on how exactly to implement: specifically; how to deal with _lengths_?
That way, we always have a standardized type and the function that uses this type can determine to what it needs to be converted (DDL, schema type, casting,...). Each of these can have a separate dict/set of functions if required.  

Again, just a braindump. Not fully worked out idea.
It's very much the same as what we have now, but slightly more standardized to include `length` into the data type as a separate parameter.

Niels-b (approved):
I ran into the same issue with Athena and made the exact same changes to the casting function.
Added extra comment, but that is beyond the scope of this PR and will require much more work.

soda-core/src/soda_core/common/sql_dialect.py
tombaeyens: Agreed: probably good idea, but not scope of this pr.
It's a big refactoring that would be best to discuss and plan ahead for.
So definitely no blocker to get this PR merged.

tombaeyens: I created https://linear.app/sodadata/issue/R-169/refactor-data-types for this

PR #2388, opened 2025-08-22
Stabilize soda extensions

paulteehan:
Companion for related work in soda-extensions
https://github.com/sodadata/soda-extensions/pull/57

### DIFF
--- a/soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
+++ b/soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
@@ -4,7 +4,7 @@ from typing import Optional
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.dataset_identifier import DatasetIdentifier
 from soda_core.common.logging_constants import soda_logger
-from soda_core.common.sql_ast import COLUMN, INSERT_INTO, TUPLE, VALUES, VALUES_ROW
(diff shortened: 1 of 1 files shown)


soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
```diff
@@ -119,6 +121,15 @@ def _create_connection(
 
                     return connection
                 else:
+                    if config.database == ":memory:":
+                        # Re-use existing in-memory connection if it exists
+                        global _in_memory_connection
```
tombaeyens, line 126: Isn't this automatic recycling of the connection going to lead to unexpected behavior?  Isn't it an option to pass the connection or the original data source as a parameter to the new data source creation?

soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source_connection.py
```diff
@@ -133,5 +133,5 @@ def _create_connection(
             options=options,
             **self.keepalives_params,
         )
-
+        conn.autocommit = True
```
tombaeyens, line 136: `autocommit` looks good to me as a strategy that we should apply everywhere.  Not sure if it's worth it to create a task to verify and ensure we apply autocommit on all other data sources.

PR #2393, opened 2025-08-26
PLATL-128 - Refactor soda-core v4 CI to use AWS Secretsmanager for se…

dakue-soda:
…crets


.github/workflows/main.workflow.yaml
```diff
@@ -31,13 +48,26 @@ jobs:
...
       - name: Define modules
         id: modules
         run: |
-          if [ -z "${{ inputs.dataSource }}" ]; then
+          echo "INFO: DATA_SOURCE is set to ${DATA_SOURCE}"
+          if [ "${DATA_SOURCE}" = "all" ]; then
             echo modules=$(bash scripts/test_matrix.sh) >> "$GITHUB_OUTPUT"
           else
-            echo modules=["${{ inputs.dataSource }}"] >> "$GITHUB_OUTPUT"
+            echo 'modules=["__DATA_SOURCE__"]' | sed "s|__DATA_SOURCE__|${DATA_SOURCE}|g" >> "$GITHUB_OUTPUT"
```
mivds, line 70: Why do we need the `sed` pipe here? Seems fine, but would think we can just specify the variable directly, e.g. `echo modules=["${DATA_SOURCE}"] >> "$GITHUB_OUTPUT"`

mivds (approved):

.github/workflows/main.workflow.yaml
dakue-soda: The problem here is that the quoting seems to get lost and then the conversion from JSON to Github Actions Matrix fails so I'm doing this with an old trick of mine.

nielsn (approved):

PR #2395, opened 2025-08-27
Supporting unsupported types, mapping to varchar

tombaeyens:
(description by Paul)

This is a monster PR that incorporates several branches of work which were happening in parallel and then eventually wound up here.   I believe the Athena support and the Snowflake auth addition have already been reviewed from a prior branch.  The rest is new in this branch.  
(update -- Athena was already merged but for some reason it's showing up in the diff as if it was new, not sure why)

### Structural changes
- Use new concept of DataSourceNamespaces to replace prefixes
- Metadata columns query refactored, entrypoint moved from SQlDialect to DataSourceImpl, support for extracting precision & scale
- Refactored Soda data types
- Added contract interfaces*

### New feature implementations
- Athena support
- Support precision and scale in numeric and datetime types
- Added bidrectional data mappings for all data sources and listed supported/unsupported types
	- Mappings are a first pass, not final, we will review in a future PR.  Don't consider mapping correctness in this review
- Added snowflake oauth login path
	
### Fixes

- Fixed typo (supports_data_type_character_maximun_length) -> supports_data_type_character_maximum_length
- fixed an issue with sql server regexes
- Other small fixes


soda-core/src/soda_core/common/metadata_types.py
Niels-b, line 1: Remark in general (not to be fixed in this PR): I think the `get_sql_data_type_str_with_parameters(self)` function should be something data source specific (i.e. put it in the dialect) instead of being a global one. Now it requires "hacky" workarounds in some data sources to get it to work.
Not to be fixed immediately (certainly not in this PR). This doesn't seem like an easy fix.

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -1007,3 +1028,140 @@ def default_numeric_precision(self) -> Optional[int]:
 
     def default_numeric_scale(self) -> Optional[int]:
         return None
+
+    def build_columns_metadata_query_str(self, table_namespace: DataSourceNamespace, table_name: str) -> str:
```
Niels-b, line 1042: Seems like a large chunk of code to have in the dialect file.
I think the dialect is a nice place for this function, but maybe we can move the actual implementation to another file to keep it a bit cleaner?
No hard opinion, just a suggestion

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -135,5 +134,30 @@ def test_connection_error_message(self) -> Optional[str]:
...
     def parse_column_names_from_query_result(self, query_result: QueryResult) -> list[str]:
         return [self.connection._execute_query_get_result_row_column_name(column) for column in query_result.columns]
+
+    def get_columns_metadata(self, dataset_prefixes: list[str], dataset_name: str) -> list[ColumnMetadata]:
+        sql: str = self.build_columns_metadata_query_str(dataset_prefixes=dataset_prefixes, dataset_name=dataset_name)
+        query_result: QueryResult = self.execute_query(sql)
+        return self.sql_dialect.build_column_metadatas_from_query_result(query_result)
+
+    def build_columns_metadata_query_str(self, dataset_prefixes: list[str], dataset_name: str) -> str:
+        database_index: int | None = self.sql_dialect.get_database_prefix_index()
```
nielsn, line 150: Generic remark: I'm still not happy about the way we deal with the concepts of "database" and "schema", we still need to work on a proper terminology and/or abstraction IMHO.

This snippet still seems pretty brittle, but that's work for another time.

soda-core/src/soda_core/common/sql_dialect.py
nielsn: I think this approach would be valid in a couple of places: replace a single file by a directory with multiple modules, maybe even introduce packages? E.g. splitting up the current `soda_cloud.py` a bit as well.

No hard requirement for the time being.

Niels-b: FYI; we tried moving this, but we can only really move two functions (`build_columns_metadata_query_str` and `build_column_metadatas_from_query_result`). The other functions are overwritten in the inheriting classes, so they require different logic. Currently makes little sense to add the extra complexity of moving those two functions, but still pointing towards the dialect. That makes it so you're bouncing between files way too much.
"Nice to have" at the moment to move all of this: we might need an extra "dialect-like" class for each datasource impl just to do the metadata stuff. But then the argument can be made to just keep it in the dialect 🙃 .

PR #2400, opened 2025-09-01
Added new interfaces for Soda Cloud Output and Logging Output

Niels-b:
Merge this first before the corresponding PR in soda-extensions (https://github.com/sodadata/soda-extensions/pull/81)

### DIFF
--- a/soda-core/src/soda_core/contracts/contract_interfaces.py
+++ b/soda-core/src/soda_core/contracts/contract_interfaces.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from typing import Any
 
 
-class ISodaCloudOutput(ABC):
+class SodaCloudJsonable(ABC):
     """Interface for getting the Soda Cloud output.
     Implement this if you want to convert your object to something that can be sent to Soda Cloud.
     The output of this function depends on the interface that Soda Cloud expects. Can be a scalar value, can be a list, dict, etc.
@@ -16,7 +16,7 @@ class ISodaCloudOutput(ABC):
         raise NotImplementedError
 
 
-class ILoggingOutput(ABC):
+class Loggable(ABC):
     """Interface for getting the logging output.
     Implement this if you want to convert your object to something that can be logged.
     We currently expect the interface to be implemented in a class that has the values required to generate the output. But feel free to add a parameter to the function if you want to.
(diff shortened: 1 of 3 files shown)


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -297,25 +298,42 @@ def log_table_row_diagnostics(self, verbose: bool = True) -> str:
         diagnostics = []
 
         if self.diagnostic_metric_values:
-            for metric_name, value in self.diagnostic_metric_values.items():
-                formatted_value = value
-                if not verbose:
-                    formatted_value = self._log_console_format(value)
-                diagnostics.append(f"{metric_name}: {formatted_value}")
+            # If the diagnostic metric values implements the ISodaCloudOutput interface, use it
+            if isinstance(self.diagnostic_metric_values, ILoggingOutput):
```
m1n0: I wonder how this works for a list of ILoggingOutput instances, it would skip this part and go into `_log_console_format` where handling of that interface is not present, or am I mistaken? Wouldn't it make sense to handle ILoggingOutput in there alongside other types so that recursive calls are handled as well?

soda-core/src/soda_core/contracts/contract_interfaces.py
m1n0, line 1: this feels a bit too java-like, I am not sure about the `I...` prefixes to be honest. I am not strongly opinionated on this but imho we could go with a bit more generic naming, like `Jsonable` and `Loggable` which would feel a bit more pythonic and afaik are used in some libraries out there as well, wdyt?

soda-core/src/soda_core/contracts/contract_verification.py
Niels-b: I was thinking that if you have a list of `LoggingOutput`s, then this should be encapsulated into another class that contains that list. Then this can implement the interface and log it. An example of this is the Groupby class implemented in the extensions PR.
Also, the rest of the code (in the `else`) assumes that `diagnostic_metric_values` is a dict (we call `.items()`). So that's not always the case and would require a bit more refactoring. This seems like an okay solution to that imo.
WDYT?

soda-core/src/soda_core/contracts/contract_interfaces.py
Niels-b: Happy to change the name to something like that yes.
`Loggable` is perfect.
For `Jsonable` I would change that to `SodaCloudJsonable` to be more specific. WDYT?

soda-core/src/soda_core/contracts/contract_verification.py
m1n0: ah yes I have seen that in the other PR and forgot to circle back here. Seems fine!

soda-core/src/soda_core/contracts/contract_interfaces.py
m1n0: works for me

PR #2404, opened 2025-09-04
Fix logging for groupby (and future SodaCloudJsonable)

Niels-b:


soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source_connection.py
```diff
@@ -71,3 +76,12 @@ def _create_connection(
...
     ):
         return psycopg2.connect(**config.to_connection_kwargs())
+
+    def execute_query(self, sql: str) -> QueryResult:
+        try:
+            return super().execute_query(sql)
+        except psycopg2.errors.Error as e:  # Catch the error and roll back the transaction
+            logger.warning(f"SQL query failed: \n{sql}\n{e}")
+            logger.debug("Rolling back transaction")
+            self.rollback()
```
mivds, line 86: This suggests there is a transaction active. What happens if there is no transaction? There may be one automatically created for every statement, just checking because I don't know.

Looking a bit further, I see we're doing this in the base class:
```python
cursor = self.connection.cursor()
try:
    ...
finally:
    cursor.close()
```

Would consider changing that to:
```python
with self.connection.cursor() as cursor:
    ...
```

For postgres, that should automatically handle the rollback based on [the docs](https://www.psycopg.org/docs/connection.html#connection.rollback):
> Changed in version 2.5: if the connection is used in a with statement, the method is automatically called if an exception is raised in the with block.

I imagine other databases might have similar problems, but I don't know if all databases support the `with` statement. So would warrant some further research/testing.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source_connection.py
```diff
@@ -71,3 +76,12 @@ def _create_connection(
         config: PostgresConnectionProperties,
     ):
         return psycopg2.connect(**config.to_connection_kwargs())
+
+    def execute_query(self, sql: str) -> QueryResult:
+        try:
+            return super().execute_query(sql)
+        except psycopg2.errors.Error as e:  # Catch the error and roll back the transaction
+            logger.warning(f"SQL query failed: \n{sql}\n{e}")
```
mivds, line 84: Do we need the warning statement here? I'm guessing the same info is already logged in the base class. Not sure though, just going off of experience with how `soda-library` does things.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
```diff
@@ -247,3 +248,22 @@ def default_varchar_length(self) -> Optional[int]:
...
+        # SqlServer supports a max of 1000 values in an insert statement. If that's the case, split the insert into multiple statements and recursively call this function.
+        STEP_SIZE = 1000
+        if len(insert_into.values) > STEP_SIZE:
+            final_insert_sql = ""
+            for i in range(0, len(insert_into.values), STEP_SIZE):
+                temp_insert_into = INSERT_INTO(
+                    fully_qualified_table_name=insert_into.fully_qualified_table_name,
+                    columns=insert_into.columns,
+                    values=insert_into.values[i : i + STEP_SIZE],
+                )
```
mivds, line 262: Don't we need to consider the number of columns here? Say we're inserting 500 rows into a table with 3 columns. That's 1500 values. The first iteration would take the first 1000 values, but that number is not divisible by the number of columns. I would expect the database to error out on that

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1269,7 +1260,14 @@ def _build_v4_diagnostics_check_type_json_dict(check_result: CheckResult) -> Opt
             # We skip the checkRowsTested because it causes extra compute.
             # To be re-evaluated when we have users asking for it.
         }
-    raise SodaException(f"Invalid check type: {type(check_result.check).__name__}")
+    else:
+        # If we have a diagnostic metric values and it implements the ISodaCloudOutput interface, use it
+        if check_result.diagnostic_metric_values is None:
```
mivds, line 1265: Sanity check: is the `diagnostic_metric_values` attribute guaranteed to be available? Previous there was a `try: ... catch AttributeError: ...` around it, implying that the attribute may not be there.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
Niels-b: The SqlServer (and error) state `rows`. So just splitting based on the rows, like we do here, is okay. In testing, I create a table with ~3000 values (3 columns x 1000 rows). That seems to be working :).
Note: the `insert_into.values` is a `VALUE_ROW` object which contains all values for a single row.

soda-core/src/soda_core/common/soda_cloud.py
Niels-b: Yes, that is always available. It's created and set to `None` in the constructor of `CheckResult`.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source_connection.py
Niels-b: I like to explicitly add it. We still throw the exception and that may need to be handled by the programmer that uses this function. Now it's not required that they add their own code to print out the sql anymore (this is not always the case AFAIK).
So needed; no. But I think it's okay to have it here :)

mivds: In that case: definitely good to have it here 👍! The way it works in `soda-library` is query failures are immediately logged (as error), but fail silently. Thought it might be the same here, which would lead to always having double logging

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
mivds: Ah nice. Maybe good to rename the comment, it currently mentions 1000 values which is a little misleading. I guess the limit is inserting 1000 rows at once?

Niels-b: Yes, will do that! Good idea 👍

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source_connection.py
Niels-b: Using the `with` statement does **not** fix the error (even though the docs say so, verified with psycopg2 v2.9.10). 🤔 . We still need the try/except block for postgres to handle the invalid transaction error.
To avoid conflicts with other datasources, maybe best not to change it. WDYT?

Niels-b: During testing, I found that  no other datasets have this problem. It's specific to Postgres. 
Each datasource has to handle at least one query with a column that does not exists -> results in an error for that query that we catch and handle in the rest of the code.

mivds: Thanks for testing! Ignore my comment in that case, all good 👍

PR #2408, opened 2025-09-16
Fix data type mappings in pg and fabric

m1n0:
Duckdb: use fixture connection in db filename to avoid race conditions
Fabric: adjust data type mappings for text and timezone aware timestamp
Postgres: use decimal for decimal not double precision to match internal "decimal" vs "numeric" semantic types


Niels-b (approved):
LGTM!

PR #2410, opened 2025-09-18
Add more data type tests and mappings

Niels-b:
Still very much a WIP commit and expecting stuff to break for lots of data sources.

Found some gaps in the testing and want to address them with this PR:
- Not all data types are ever created with a test table
- We don't check the mappings for these data types once they're created

This PR addresses that by adding tests that create a test table with all possible SodaDataTypes and then read the columns back from the metadata and verifies that we can map it again to the same (or synonym) of that SodaDataType. 
If this is possible, that means that we should be able to read supported data types from a data source and insert them into the same data source again.

### DIFF
--- a/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
+++ b/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
@@ -140,7 +140,7 @@ class DuckDBSqlDialect(SqlDialect):
         return {
             SodaDataTypeName.CHAR: "varchar",  # DuckDB doesn't have fixed CHAR, maps to VARCHAR
             SodaDataTypeName.VARCHAR: "varchar",
-            SodaDataTypeName.TEXT: "text",  # TEXT is an alias for VARCHAR
+            SodaDataTypeName.TEXT: "text",  # TEXT is an alias for VARCHAR in DuckDB. So using text will result in a VARCHAR being created.
             SodaDataTypeName.SMALLINT: "smallint",
             SodaDataTypeName.INTEGER: "integer",
             SodaDataTypeName.BIGINT: "bigint",
--- a/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
+++ b/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
@@ -317,7 +317,7 @@ class SqlServerSqlDialect(SqlDialect):
     def is_same_soda_data_type(self, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
         found_synonym = False
         synonym_correct = False
-        if expected == SodaDataTypeName.TEXT or actual == SodaDataTypeName.VARCHAR:
+        if expected == SodaDataTypeName.TEXT or expected == SodaDataTypeName.VARCHAR:
             (found_synonym, synonym_correct) = (
                 True,
                 actual == SodaDataTypeName.VARCHAR or actual == SodaDataTypeName.TEXT,
--- a/soda-tests/tests/features/test_soda_data_types.py
+++ b/soda-tests/tests/features/test_soda_data_types.py
@@ -26,7 +26,9 @@ def test_all_data_types_explicitly(data_source_test_helper: DataSourceTestHelper
 
     # Make sure that each SodaDataType is mapped to a data type in the data source
     for soda_data_type in SodaDataTypeName:
-        assert soda_data_type in mappings_from_canonical_to_data_type
+        assert (
+            soda_data_type in mappings_from_canonical_to_data_type
+        ), f"SodaDataType {soda_data_type} not found in mappings_from_canonical_to_data_type"
 
 
 def test_mapping_canonical_to_data_type_to_canonical(data_source_test_helper: DataSourceTestHelper):
@@ -110,7 +112,7 @@ def test_mapping_canonical_to_data_type_to_canonical(data_source_test_helper: Da
         dataset_prefixes=test_table.dataset_prefix, dataset_name=test_table.unique_name
     )
 
-    assert len(actual_columns) == 23
+    assert len(actual_columns) == len(test_table.columns)
 
     def convert_metadata_to_soda_data_type(metadata: ColumnMetadata) -> SodaDataTypeName:
         sql_data_type: SqlDataType = metadata.sql_data_type
--- a/soda-tests/tests/features/test_sql_ast_inserts.py
+++ b/soda-tests/tests/features/test_sql_ast_inserts.py
@@ -229,7 +229,7 @@ def test_datetime_microsecond_precision_insert(data_source_test_helper: DataSour
         .build()
     )
 
-    test_table = data_source_test_helper.ensure_test_table(microsecond_test_table_specification)
+    test_table = data_source_test_helper.ensure_test_table(microsecond_test_table_specification, force_recreate=True)
     my_table_name = data_source_test_helper.get_qualified_name_from_test_table(test_table)
 
     sql_dialect: SqlDialect = data_source_test_helper.data_source_impl.sql_dialect
@@ -245,14 +245,10 @@ def test_datetime_microsecond_precision_insert(data_source_test_helper: DataSour
     assert len(result.rows) == NUMBER_OF_ROWS
     for i in range(NUMBER_OF_ROWS):
         if data_source_test_helper.data_source_impl.sql_dialect.supports_datetime_microseconds():
-            assert result.rows[i][3] == datetime.datetime(2021, 1, 1, i, 0, 0, 123) or result.rows[i][
-                3
-            ] == datetime.datetime(
-                2021, 1, 1, i, 0, 0, 123, tzinfo=datetime.timezone.utc
-            )  # We're not here to check the timezone returned, only that the microsecond is correct
+            # We're not here to check the timezone returned, only that the microsecond is correct
+            returned_datetime = result.rows[i][3].replace(tzinfo=None)
+            assert returned_datetime == datetime.datetime(2021, 1, 1, i, 0, 0, 123)
         else:
-            assert result.rows[i][3] == datetime.datetime(2021, 1, 1, i, 0, 0) or result.rows[i][
-                3
-            ] == datetime.datetime(
-                2021, 1, 1, i, 0, 0, tzinfo=datetime.timezone.utc
-            )  # We're not here to check the timezone returned, only that the microsecond is correct
+            # We're not here to check the timezone returned, only that the microsecond is correct.
+            returned_datetime = result.rows[i][3].replace(tzinfo=None)
+            assert returned_datetime == datetime.datetime(2021, 1, 1, i, 0, 0)


soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
```diff
@@ -156,12 +156,44 @@ def get_data_source_data_type_name_by_soda_data_type_names(self) -> dict:
             SodaDataTypeName.FLOAT: "float",
             SodaDataTypeName.DOUBLE: "double",
             SodaDataTypeName.DATE: "date",
-            SodaDataTypeName.TIME: "date",
+            SodaDataTypeName.TIME: "varchar",  # Athena does not support time data type, for time specifically, it should be formatted as a string!
```
mivds, line 159: If the data type isn't supported, shouldn't we just remove it from the `dict` in that case? I haven't looked at the full flow, but I imagine that might result in more comprehensive behavior. E.g. logging a warning and treating it as an unmapped data type rather than silently handling it as string.

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
```diff
@@ -156,12 +156,44 @@ def get_data_source_data_type_name_by_soda_data_type_names(self) -> dict:
...
             SodaDataTypeName.TIMESTAMP: "timestamp",
             SodaDataTypeName.TIMESTAMP_TZ: "timestamp",
             SodaDataTypeName.BOOLEAN: "boolean",
         }
 
+    # AI generated
+    def get_soda_data_type_name_by_data_source_data_type_names(self) -> dict[str, SodaDataTypeName]:
```
mivds, line 166: Out of scope for this PR, but not sure a `dict` actually covers all the use cases. I think a signature like `get_soda_data_type_name_from_data_source_data_type(self, str) -> SodaTypeName` might make more sense to offer flexibility for data sources that do funky stuff, e.g. `decimal(2,5)`.

Guessing we're already doing some string preprocessing to remove suffixes; just not sure that works for all databases.

mivds, line 166: Any distinction needed for DML & DDL types? if I remember correctly athena was annoying here and used different names for those. Not sure if that also impacts the data types we read from `INFORMATION_SCHEMA` tables

soda-core/src/soda_core/common/consistent_hash_builder.py
```diff
@@ -35,6 +35,8 @@ def add(self, value: Optional[object]) -> ConsistentHashBuilder:
                 self.add(str(value))
             elif isinstance(value, date):
                 self.add(str(value))
+            elif isinstance(value, time):
+                self.add(str(value))
```
mivds, line 39: Just out of curiosity: why are we including times in hashes? That seems to deviate from the original use case for consistent hashes (so does `datetime` and `date`, so nothing new blocking here)

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -199,3 +199,37 @@ def build_literal_with_alias(literal, alias: COLUMN | None) -> str:
...
+                    actual in list_of_synonyms and expected in list_of_synonyms,
+                )
+                break
+
+        if found_synonym and synonym_correct:
+            if expected != actual:
+                logger.debug(f"In is_same_soda_data_type, Expected {expected} and actual {actual} are the same")
+            return True
+        else:
+            return super().is_same_soda_data_type(expected, actual)
```
mivds, line 235: Think this can be simplified a bit


```suggestion
    def is_same_soda_data_type(self, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
        if expected == actual:
            return True

        list_of_text_synonyms = [SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR, SodaDataTypeName.CHAR]
        list_of_integer_synonyms = [SodaDataTypeName.INTEGER, SodaDataTypeName.BIGINT, SodaDataTypeName.SMALLINT]
        list_of_numeric_synonyms = [SodaDataTypeName.NUMERIC, SodaDataTypeName.DECIMAL]
        list_of_timestamp_synonyms = [
            SodaDataTypeName.TIMESTAMP,
            SodaDataTypeName.TIMESTAMP_TZ,
        ]  # Bigquery does not support timezones
        list_of_float_synonyms = [SodaDataTypeName.FLOAT, SodaDataTypeName.DOUBLE]

        for list_of_synonyms in [
            list_of_text_synonyms,
            list_of_integer_synonyms,
            list_of_numeric_synonyms,
            list_of_timestamp_synonyms,
            list_of_float_synonyms,
        ]:
            if expected in list_of_synonyms and actual in list_of_synonyms:
                logger.debug(f"In is_same_soda_data_type, Expected {expected} and actual {actual} are the same")
                return True

        return super().is_same_soda_data_type(expected, actual)
```

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -233,7 +239,9 @@ def qualify_dataset_name(self, dataset_prefix: list[str], dataset_name: str) ->
     def literal(self, o: object) -> str:
         if o is None:
             return "NULL"
-        elif isinstance(o, Number):
+        elif isinstance(o, Number) and not isinstance(
+            o, bool
+        ):  # We don't want to interpret booleans as numbers. In Python a boolean is a subclass of int, so we have to explicitly check it
```
mivds, line 244: In terms of perfomance, maybe better to move the boolean condition ahead of this one?

See below:
```
        elif isinstance(o, bool):
            return self.literal_boolean(o)
```

soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
```diff
@@ -140,7 +140,7 @@ def get_data_source_data_type_name_by_soda_data_type_names(self) -> dict:
         return {
             SodaDataTypeName.CHAR: "varchar",  # DuckDB doesn't have fixed CHAR, maps to VARCHAR
             SodaDataTypeName.VARCHAR: "varchar",
-            SodaDataTypeName.TEXT: "varchar",  # TEXT is an alias for VARCHAR
+            SodaDataTypeName.TEXT: "text",  # TEXT is an alias for VARCHAR
```
mivds: Comment seems confusing here, not sure if still applicable?

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
mivds: I see many of the data sources use a similar implementation. Maybe good to extract that as a utility function?

Something like `is_same_soda_data_type(aliases: Iterable[Iterable[SodaDataTypeName]], type1: SodaDataTypeName, type2: SodaDataTypeName) -> bool:`

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
```diff
@@ -304,3 +313,19 @@ def map_test_sql_data_type_to_data_source(self, source_data_type: SqlDataType) -
         if sql_data_type.name == "varchar":
             sql_data_type.character_maximum_length = self.default_varchar_length()
         return sql_data_type
+
+    def is_same_soda_data_type(self, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
+        found_synonym = False
+        synonym_correct = False
+        if expected == SodaDataTypeName.TEXT or actual == SodaDataTypeName.VARCHAR:
```
mivds: Think there is an error here?


```suggestion
        if expected == SodaDataTypeName.TEXT or expected == SodaDataTypeName.VARCHAR:
```

It's similar pattern as above, so might be replaced with utility method?

soda-tests/tests/features/test_soda_data_types.py
```diff
@@ -10,3 +17,137 @@ def test_soda_data_types(data_source_test_helper: DataSourceTestHelper):
...
+
+
+def test_all_data_types_explicitly(data_source_test_helper: DataSourceTestHelper):
+    mappings_from_canonical_to_data_type: dict = (
+        data_source_test_helper.data_source_impl.sql_dialect.get_data_source_data_type_name_by_soda_data_type_names()
+    )
+
+    # Make sure that each SodaDataType is mapped to a data type in the data source
+    for soda_data_type in SodaDataTypeName:
+        assert soda_data_type in mappings_from_canonical_to_data_type
```
mivds: Just to make life easier, I suggest adding add a message that indices which data type is missing.

```suggestion
        assert soda_data_type in mappings_from_canonical_to_data_type, f"{soda_data_type} does not have a mapping"
```

soda-tests/tests/features/test_soda_data_types.py
```diff
@@ -10,3 +17,137 @@ def test_soda_data_types(data_source_test_helper: DataSourceTestHelper):
...
+    # Building the test table already tests the mapping from canonical to data type
+    # Behind the scenes, we build this table using SodaDataTypeName.CHAR, SodaDataTypeName.SMALLINT, etc.
+    # So if the table gets created, then the mapping from canonical to data type is correct
+    test_table = data_source_test_helper.ensure_test_table(test_table_specification)
+
+    actual_columns: list[ColumnMetadata] = data_source_test_helper.data_source_impl.get_columns_metadata(
+        dataset_prefixes=test_table.dataset_prefix, dataset_name=test_table.unique_name
+    )
+
+    assert len(actual_columns) == 23
```
mivds: I'm sure we'll change the list of columns at some point 😅

```suggestion
    assert len(actual_columns) == len(test_table_specification.columns)
```

soda-tests/tests/features/test_soda_data_types.py
```diff
@@ -10,3 +17,137 @@ def test_soda_data_types(data_source_test_helper: DataSourceTestHelper):
...
+
+    def convert_metadata_to_soda_data_type(metadata: ColumnMetadata) -> SodaDataTypeName:
+        sql_data_type: SqlDataType = metadata.sql_data_type
+        data_type_name = sql_data_type.name
+        return data_source_test_helper.data_source_impl.sql_dialect.get_soda_data_type_name_by_data_source_data_type_names()[
+            data_type_name
+        ]
+
+    column_mappings = {
```
mivds, line 124: Nit: maybe the column names would have been a better choice as key here, rather than indices

soda-tests/tests/features/test_sql_ast_inserts.py
```diff
@@ -209,3 +209,50 @@ def test_large_insert_test_table(data_source_test_helper: DataSourceTestHelper):
...
+
+    test_table = data_source_test_helper.ensure_test_table(microsecond_test_table_specification)
+    my_table_name = data_source_test_helper.get_qualified_name_from_test_table(test_table)
+
+    sql_dialect: SqlDialect = data_source_test_helper.data_source_impl.sql_dialect
+
+    select_sql = sql_dialect.build_select_sql(
+        [
+            SELECT([COLUMN("id"), COLUMN("id2"), COLUMN("my_value"), COLUMN("my_timestamp")]),
+            FROM(my_table_name[1:-1] if sql_dialect.is_quoted(my_table_name) else my_table_name),
```
mivds, line 240: Oof, that `sql_dialect.is_quoted` trick feels hacky. Can't we define `data_source_test_helper.get_qualified_name_from_test_table(...)` in such a way that we know it will always be quoted (or not quoted)?

soda-tests/tests/features/test_sql_ast_inserts.py
```diff
@@ -209,3 +209,50 @@ def test_large_insert_test_table(data_source_test_helper: DataSourceTestHelper):
...
+    )
+    result: QueryResult = data_source_test_helper.data_source_impl.execute_query(select_sql)
+    assert len(result.rows) == NUMBER_OF_ROWS
+    for i in range(NUMBER_OF_ROWS):
+        if data_source_test_helper.data_source_impl.sql_dialect.supports_datetime_microseconds():
+            assert result.rows[i][3] == datetime.datetime(2021, 1, 1, i, 0, 0, 123) or result.rows[i][
+                3
+            ] == datetime.datetime(
+                2021, 1, 1, i, 0, 0, 123, tzinfo=datetime.timezone.utc
+            )  # We're not here to check the timezone returned, only that the microsecond is correct
```
mivds: Suggest a small simplification (should help readability; didn't immediately realize why there were 2 checks)

```suggestion
            # We're not here to check the timezone returned, only that the microsecond is correct
            returned_datetime = result.rows[i][3].replace(tzinfo=None)
            assert result.rows[i][3] == datetime.datetime(2021, 1, 1, i, 0, 0, 123)
```

mivds (approved):

mivds:
@Niels-b Left some comments. I think all of this is polish and nice-to-have. The only actual bug (I think) is https://github.com/sodadata/soda-core/pull/2410#discussion_r2362679579

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
Niels-b: We explicitly check that all the SodaDataTypes are in this dict. 
That means that if the SodaDataType is created, we assume that each data source has a mapping for it.
If we encounter a data type that is not in the SodaDataType, it's automatically mapped to `varchar`. So even if we'd remove it, it will still have the same mapping. 

Coincidentally, for Athena, the "goto" for TIME is to map it to varchar apparently 🙃 .

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
Niels-b: That is indeed something that needs to be handled. We have the `ColumnMetaData` class for that. 
So the support for `decimal(2,5)` is there. But that's not what the `SodaDataTypeName` is supposed to handle.
Makes it more complicated and definitely out of scope for this PR, but we should be covered on those things AFAIK.

Niels-b: No distinction implemented here. Athena is indeed the only data source that gives issues with this. More specifically with the `TIME` type. Don't think it's needed at this moment.
Types read from information schema are always the same, and these are mapped correctly. (At least that's what we have the tests for now :D )

soda-core/src/soda_core/common/consistent_hash_builder.py
Niels-b: I'm not fully aware of the deviation from the original use case. That's news to me.

Context for this: the `TIME` type was never tested, and the hash builder was giving issues on this. Explicitly adding the type (for which we already run `datetime` and `date` through it) fixed the issue and seemed to behave okay.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: Milan suggested the same thing. I got a few ideas to improve this too. Think it can be even shorter/simpler than this.
For me, thats OOS for this PR. But definitely writing this down to fix next week before it gets out of hand!

soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
Niels-b: Good catch! I'll remove that.

Niels-b: Ah, just remembered. DuckDB maps TEXT to VARCHAR behind the scenes. That's what this comment means. I'll try to explain more clearly.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
Niels-b: Oooh, good catch! I will update that.
I'll improve this with the refactor on this next week.

soda-core/src/soda_core/common/consistent_hash_builder.py
mivds: As far as I know the hash builder was in the past only used to generate check/metric identities. It would receive things like the column, filename, etc. I doubt there was ever date/datetime/time in there as that would logically make it not consistent (when using the concept of now), but maybe there were other use cases 🤷

soda-tests/tests/features/test_soda_data_types.py
Niels-b: I wish I thought of that. Many bugs during development because of incorrect mapping. I'll add it to the fixes for next week.

soda-tests/tests/features/test_sql_ast_inserts.py
Niels-b: Yes, feels hacky indeed.
I believe that the FROM will always quote things automatically. So if it's already quoted, we need to remove it.
Again; I'll add it to improvements for next week. This seems to be working for now and don't want to break stuff last minute.

soda-tests/tests/features/test_sql_ast_inserts.py
Niels-b: Definitely more readable! Good suggestion, I'll change it.

PR #2412, opened 2025-09-22
Improvements around SodaDataTypeName and testing suite 

Niels-b:
- Update data type synonyms 
- Changed mappings from numbers to column names
- Some small changes for Databricks compatibility

### DIFF
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
@@ -33,7 +33,7 @@ class DatabricksDataSourceImpl(DataSourceImpl, model_class=DatabricksDataSourceM
 class DatabricksSqlDialect(SqlDialect):
     DEFAULT_QUOTE_CHAR = "`"
 
-    SODA_DATA_TYPE_SYNONYM = (
+    SODA_DATA_TYPE_SYNONYMS = (
         (SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR, SodaDataTypeName.CHAR),
         (SodaDataTypeName.NUMERIC, SodaDataTypeName.DECIMAL),
         (SodaDataTypeName.TIMESTAMP_TZ, SodaDataTypeName.TIMESTAMP),


soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
```diff
@@ -147,42 +147,10 @@ def post_schema_create_sql(self, prefixes: list[str]) -> Optional[list[str]]:
...
-                logger.debug(f"In is_same_soda_data_type, Expected {expected} and actual {actual} are the same")
-            return True
-        else:
-            return super().is_same_soda_data_type(expected, actual)
+    def get_synonyms_for_soda_data_type(self) -> list[list[SodaDataTypeName]]:
+        return [
+            [SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR, SodaDataTypeName.CHAR],
+            [SodaDataTypeName.NUMERIC, SodaDataTypeName.DECIMAL],
+            [SodaDataTypeName.TIMESTAMP_TZ, SodaDataTypeName.TIMESTAMP],
+            [SodaDataTypeName.TIME, SodaDataTypeName.TEXT],
```
mivds: If I'm not mistaken this doesn't correctly reflect the earlier special case. Previously we had `TEXT` is allowed when `TIME` is expected (1-way mapping). Now it's a 2-way mapping where also `TIME` is allowed when `TEXT` is expected. Not sure if that combination of inputs is possible, but it seems risky.

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
```diff
@@ -360,25 +360,10 @@ def extract_numeric_scale(self, row: Tuple[Any, ...], columns: list[Tuple[Any, .
...
-        ):  # Athena does not support timezones, so we consider TIMESTAMP and TIMESTAMP_TZ to be the same
-            (found_synonym, synonym_correct) = (True, actual == SodaDataTypeName.TIMESTAMP)
-
-        if found_synonym and synonym_correct:
-            if expected != actual:
-                logger.debug(f"In is_same_soda_data_type, Expected {expected} and actual {actual} are the same")
-            return True
-        else:
-            return super().is_same_soda_data_type(expected, actual)
+    def get_synonyms_for_soda_data_type(self) -> list[list[SodaDataTypeName]]:
```
mivds: I would consider putting these things as class variables rather than functions. Since it's just a static definition for all implementations, I think the runtime cost as a function will be a lot higher.

Essentially a choice between creating new `dict` objects on every call (with memory allocation & garbage collection) or having a definition that is in-memory for the entire lifecycle of the program.

Something like:
```python
class AthenaSqlDialect:
    SODA_DATA_TYPE_SYNONYMS = (
        (SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR),
        (SodaDataTypeName.NUMERIC, SodaDataTypeName.DECIMAL),
        (SodaDataTypeName.TIMESTAMP_TZ, SodaDataTypeName.TIMESTAMP),
        (SodaDataTypeName.TIME, SodaDataTypeName.VARCHAR),
    )
```

Note: best to use tuples if we go down that path, to ensure immutability

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
Niels-b: Indeed, previously it was a one way mapping.
I think it's best to check for this unique case in an overloaded function in the Databricks data source. Then use `super()` to do the rest of the mappings. That will be cleaner.

It shouldn't give any issues with the mappings though. The other way around shouldn't be possible (that's why I thought of doing it this way). However, it's better to do it correctly.

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
Niels-b: Yes, great suggestion! Thanks for that 🙌

PR #2414, opened 2025-09-22
Thresholds: support warn level

m1n0:
Adds Warn level to checks using thresholds yaml. Defaults to fail.


tombaeyens:
@m1n0 I'ld only consider to work with Aldi to ensure there are guidelines in the docs on how to do stop the pipeline.  How should people translate the results and contract verification outcome to stopping the pipeline or not.

PR #2418, opened 2025-09-23
Add missing return statement

paulteehan:
FIxes issue in which failed row keys table was overwritten on each run


Niels-b (approved):
LGTM! Nice catch for the missing return statement

PR #2420, opened 2025-09-24
Update bus dataset upload script to use correct data type mappings

Niels-b:


m1n0:
@Niels-b  is this still relevant?

Niels-b:
> @Niels-b is this still relevant?

Yes, for when we need it, I would like to keep the script somewhat updated if there are small fixes like this 😄 .
This might move to soda-extensions eventually, as we have bulk-inserts there to speed things up. But we can definitely merge it here now. WDYT?

PR #2428, opened 2025-10-01
Performance optimizations for inserts using bulk inserts

Niels-b:
We have some performance concerns for DWH with the amount of time it takes to insert a large number of rows.
E.g. for BigQuery, we do inserts 500 rows at a time, but that takes to long.
This PR is the start of adding bulk insertion methods to datasources, using datasource specific approaches to optimize the insertion performance.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -48,8 +48,8 @@ class BigQueryDataSourceNamespace(DataSourceNamespace):
 
 
 class BigQueryDataSourceImpl(DataSourceImpl, model_class=BigQueryDataSourceModel):
-    def __init__(self, data_source_model: BigQueryDataSourceModel):
-        super().__init__(data_source_model=data_source_model)
+    def __init__(self, data_source_model: BigQueryDataSourceModel, connection: Optional[DataSourceConnection] = None):
+        super().__init__(data_source_model=data_source_model, connection=connection)
         self.cached_location = None
 
     def _create_sql_dialect(self) -> SqlDialect:
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -74,6 +74,10 @@ class DataSourceImpl(ABC):
         cls.__model_classes[type_name] = model_class
         cls.__implementation_classes[type_name] = cls
 
+    def create_copy_with_different_connection(self, connection: DataSourceConnection) -> DataSourceImpl:
+        "Use this method when you want to use the DataSourceImpl while having multiple connections open. Can be used for bulk inserts for example."
+        return self.__class__(data_source_model=self.data_source_model, connection=connection)
+
     def __str__(self) -> str:
         return self.name
 
--- a/soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source.py
+++ b/soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source.py
@@ -1,4 +1,5 @@
 from logging import Logger
+from typing import Optional
 
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_impl import DataSourceImpl
@@ -22,8 +23,8 @@ TIMESTAMP_WITH_LOCAL_TIME_ZONE = "timestamp with local time zone"
 
 
 class SnowflakeDataSourceImpl(DataSourceImpl, model_class=SnowflakeDataSourceModel):
-    def __init__(self, data_source_model: SnowflakeDataSourceModel):
-        super().__init__(data_source_model=data_source_model)
+    def __init__(self, data_source_model: SnowflakeDataSourceModel, connection: Optional[DataSourceConnection] = None):
+        super().__init__(data_source_model=data_source_model, connection=connection)
 
     def _create_sql_dialect(self) -> SqlDialect:
         return SnowflakeSqlDialect()
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
@@ -27,8 +27,8 @@ logger: logging.Logger = soda_logger
 
 
 class AthenaDataSourceImpl(DataSourceImpl, model_class=AthenaDataSourceModel):
-    def __init__(self, data_source_model: AthenaDataSourceModel):
-        super().__init__(data_source_model=data_source_model)
+    def __init__(self, data_source_model: AthenaDataSourceModel, connection: Optional[DataSourceConnection] = None):
+        super().__init__(data_source_model=data_source_model, connection=connection)
 
     def _create_sql_dialect(self) -> SqlDialect:
         return AthenaSqlDialect(self)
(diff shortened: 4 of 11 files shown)


soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
paulteehan, line 1: Nice, looks very clean

soda-tests/tests/features/test_sql_bulk_inserts.py
```diff
@@ -0,0 +1,454 @@
...
+                FROM(my_table_name[1:-1] if sql_dialect.is_quoted(my_table_name) else my_table_name),
+            ]
+        )
+        result: QueryResult = data_source_impl.execute_query(select_sql)
+        assert result.rows[0][0] == NUMBER_OF_ROWS + 1  # + 1 because we insert a single row
+    finally:
+        # Then drop the table to clean up
+        # We explicitly do not use the "if exists" variant, because the table should exist at this point.
+        drop_table_sql = sql_dialect.build_drop_table_sql(
+            DROP_TABLE_IF_EXISTS(fully_qualified_table_name=my_table_name)
```
paulteehan: Is that meant to be `DROP TABLE` instead of `DROP TABLE IF EXISTS` per the comment?

soda-tests/tests/features/test_sql_bulk_inserts.py
```diff
@@ -0,0 +1,454 @@
...
+
+        # Here we can define upper limits for the bulk insert time for each data source.
+        # Define them with some margin of error. For example: BigQuery takes ~10 seconds locally, so give it about 60 seconds in CI.
+        # Account for local vs online testing (data transfer speeds)!
+        if test_datasource == "bigquery":
+            assert bulk_insert_time < 60, "Bulk insert time taken is too long"
+        elif test_datasource == "postgres":
+            assert bulk_insert_time < 40, "Bulk insert time taken is too long"
+        elif test_datasource == "snowflake":
+            assert bulk_insert_time < 30, "Bulk insert time taken is too long"
```
paulteehan: I'm concerned about this taking a long time in CI, as we add more and more tests - not urgent for now but we may want to have a pytest marker for more extensive tests that we only run some of the time, and this would be a good candidate

soda-tests/tests/features/test_sql_bulk_inserts.py
```diff
@@ -0,0 +1,454 @@
...
+                    FROM(my_table_name[1:-1] if sql_dialect.is_quoted(my_table_name) else my_table_name),
+                ]
+            )
+            result: QueryResult = data_source_impl.execute_query(select_sql)
+            assert (
+                result.rows[0][0] == NUMBER_OF_ROWS * 2
+            )  # Note: this is multiplied by 2 because we insert the same rows twice!
+
+            # Testing locally, so we should report that the bulk insert is faster
+            assert bulk_insert_time < insert_into_time, (
```
paulteehan: I like this :)

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -90,6 +92,49 @@ def build_columns_metadata_query_str(self, dataset_prefixes: list[str], dataset_
...
+            raise ValueError(f"Missing required columns in the dataframe: {missing_required_columns}")
+
+        # Keep only the columns that are in the dataframe
+        schema_to_use = [schema for schema in schema_to_use if schema.name in list(df.columns)]
+
+        # Make sure that we don't use columns in the dataframe that are not in the schema
+        unknown_columns = [column for column in df.columns if column not in [schema.name for schema in schema_to_use]]
+        if unknown_columns:
+            logger.error(f"Unknown columns in the dataframe: {unknown_columns}")
+            raise ValueError(f"Unknown columns in the dataframe: {unknown_columns}")
```
paulteehan: I'm curious why the schema consistency check is necessary for BigQuery, but not for the other data sources?
Also you might consider factoring this part out into a distinct method to improve modularity

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -161,3 +168,100 @@ def build_columns_metadata_query_str(self, dataset_prefixes: list[str], dataset_
...
+    def _build_values_to_insert_dataframe(
+        self, insert_into: INSERT_INTO, convert_dtypes: Optional[str] = None
+    ) -> pd.DataFrame:
+        values_to_insert = self._build_values_to_insert_dict(insert_into)
+        start_time = time.time()
+        df = pd.DataFrame(values_to_insert)
+        if convert_dtypes:
+            df = df.convert_dtypes(
+                dtype_backend=convert_dtypes
+            )  # For BigQuery, the behind the scenes uses pyarrow for the data types. So let's convert everything using that.
```
paulteehan: We should probably have end-to-end testing to see how data types propagate through our pipeline and confirm it's working as we think it is.  There are a lot of conversions between source and DWH.  (not for now)

soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source.py
```diff
@@ -33,6 +34,22 @@ def _create_data_source_connection(self) -> DataSourceConnection:
...
+
+        connection = self.connection.connection
+
+        # Extract the unique parts of the fully qualified table name
+        database, schema, table = insert_into.fully_qualified_table_name.split(".")
+        database = database.replace('"', "")
+        schema = schema.replace('"', "")
+        table = table.replace('"', "")
+
+        result = write_pandas(connection, df, table, database=database, schema=schema, use_logical_type=True)
```
paulteehan: I love how simple this is 
Again wondering why BQ required consistency checks and this doesn't
Also I'm certain there is a pandas connector for BigQuery, did you consider using that instead of the official client?

soda-tests/tests/features/test_sql_bulk_inserts.py
Niels-b: Comment may be outdated indeed. Think there was some issue on some datasource that did some specific weird thing 😅 .
Nice catch!

soda-tests/tests/features/test_sql_bulk_inserts.py
Niels-b: Yes, I agree. 
I consider that to be part of the cleanup task.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: BigQuery is very picky about the schema and "REQUIRED" fields etc. And this is relatively specific to BigQuery.
If it's needed for other data sources we can definitely do that. For now, the methods for other datasources seem to handle it well.

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: We do have testing in extensions that validate the data that we query from the DWH. That uses the bulk-insert method, so we verify the results that way. We can try to do something similar in core, but that's indeed "not for now".

soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source.py
Niels-b: Snowflake does the matching itself 🙃 . So much simpler.
The official client for BigQuery had more options (was also exploring other ways of doing it besides pandas). So I just stuck with that tbh. We may be moving away from pandas later, so it will do for now.

Niels-b:
Thanks for reviewing!
> Do we need bulk inserts in core? 

This is a product question. I think it's fine to have it here. Don't think we should paywall performance ;).

> I am not super fond of adding pandas as core dependency, and also psycopg2+3 for postgres

@nielsn and I just had a discussion on that; see this [Slack thread](https://sodadata.slack.com/archives/C08KXG6PGP2/p1759745110559889)

PR #2429, opened 2025-10-01
Add debug print limits for query results

paulteehan:


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -69,25 +74,52 @@ def execute_query(self, sql: str) -> QueryResult:
...
         finally:
             cursor.close()
 
     def format_rows(self, rows: list[tuple]) -> list[tuple]:
         return rows
 
+    def truncate_sql(self, sql: str) -> str:
+        """Truncate large strings in sql to a reasonable length."""
+        max_chars = SODA_DEBUG_PRINT_SQL_MAX_CHARS
+        if len(sql) > max_chars:
```
Niels-b: NIT: do we need the extra check for this? We can always cut it off using `sql[:max_chars]`. Would then only be to determine if we need the `...` at the end. Which for me, is not needed as we already say in the debug statement that it's limited.
Although this will make it clearer when a cut-off has happened 😅 .
No strong opinion on this! I'm good to leave it like this as well. Just wanted to write out my thoughts 😉

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -69,25 +74,52 @@ def execute_query(self, sql: str) -> QueryResult:
...
+            return sql[:max_chars] + "..."
+        return sql
+
+    def truncate_rows(self, rows: list[tuple]) -> list[tuple]:
+        """Truncate large strings in rows to a reasonable length, and return only the first n rows."""
+        max_chars = SODA_DEBUG_PRINT_VALUE_MAX_CHARS
+        max_rows = SODA_DEBUG_PRINT_RESULT_MAX_ROWS
+
+        def truncate_cell(cell: Any) -> Any:
```
Niels-b, line 110: This can move outside the function.
We don't want to define this function every time we call the truncate_rows function. Seems like unnecessary overhead.

Niels-b (approved):
Pre-approving already, once the `truncate_cell` comment has been addressed. Minor change for that ;).
Also have one "braindump", feel free to do with that whatever you want.
Thanks for this Paul! Good work 🙌 !

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -69,25 +74,52 @@ def execute_query(self, sql: str) -> QueryResult:
...
+            if isinstance(cell, str) and len(cell) > max_chars:
+                return cell[:max_chars] + "..."
+            return cell
+
+        rows = [tuple(truncate_cell(cell) for cell in row) for row in rows]
```
milanaleksic, line 117: shouldn't we first truncate the count of rows, and then go through the rows we intend on returning to truncate the cells in them? Seems like a cheaper thing to do

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -69,25 +74,52 @@ def execute_query(self, sql: str) -> QueryResult:
...
             cursor.close()
 
     def format_rows(self, rows: list[tuple]) -> list[tuple]:
         return rows
 
+    def truncate_sql(self, sql: str) -> str:
+        """Truncate large strings in sql to a reasonable length."""
+        max_chars = SODA_DEBUG_PRINT_SQL_MAX_CHARS
+        if len(sql) > max_chars:
+            return sql[:max_chars] + "..."
```
milanaleksic: nit: I wouldn't check for max_chars but for max_chars-3 since you plan on adding 3 more characters, so the edge case is that this can end up being max 1027 chars long

soda-core/src/soda_core/common/data_source_connection.py
paulteehan: Yeah good question, I think I prefer to leave the '...' in, just to be extra clear

paulteehan:
> Wouldn't be bad to have some tests for at least the truncate_* methods

Good idea, I've added some tests

PR #2432, opened 2025-10-08
Refactor invalidity check to support failed rows

paulteehan:
Companion to https://github.com/sodadata/soda-extensions/pull/121 

Refactors the `InvalidReferenceCountQuery` to support failed rows extraction

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from enum import Enum
+
 from soda_core.common.data_source_impl import DataSourceImpl
 from soda_core.common.data_source_results import QueryResult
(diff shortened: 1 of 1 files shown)


soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -192,79 +192,88 @@ def __init__(
         data_source_impl: Optional[DataSourceImpl],
     ):
         super().__init__(data_source_impl=data_source_impl, metrics=[metric_impl])
+        self.metric_impl = metric_impl
+        self.dataset_filter = dataset_filter
+        self.check_filter = check_filter
+        # C stands for the 'C'ontract dataset
+        self.referencing_alias: str = "C"
```
Niels-b: I know this was there previously, but since we're touching it anyway we could maybe create an enum with these values? I think that's a bit cleaner.

paulteehan: Fixed thanks :)

PR #2435, opened 2025-10-11
Supporting changes for DWH failed row queries

Niels-b:
Goed hand-in-hand with a PR from soda-extensions (will post here):
Changes include:
- Minor updates to Failed Rows Check
- More type mappings for Postgres
- Updated synonym functions to support new cursor mappings (since not all mappings are exactly the same in cursor)
- Added data test helper function to create a table that includes ALL data types currently used by Soda

### DIFF
--- a/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
+++ b/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
@@ -149,6 +149,7 @@ class PostgresSqlDialect(SqlDialect):
             return "text"
         return super()._build_create_table_column_type(create_table_column=create_table_column)
 
+    @classmethod
     def is_same_soda_data_type_with_synonyms(cls, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
         # Postgres cursor can return bpchar, which is an unbounded synonym for char. We convert that to TEXT as that is the best fit. So we could expect CHAR, but actual is TEXT.
         if expected == SodaDataTypeName.CHAR and actual == SodaDataTypeName.TEXT:
--- a/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
+++ b/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
@@ -318,6 +318,7 @@ class SqlServerSqlDialect(SqlDialect):
             sql_data_type.character_maximum_length = self.default_varchar_length()
         return sql_data_type
 
+    @classmethod
     def is_same_soda_data_type_with_synonyms(cls, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
         if expected == SodaDataTypeName.CHAR and actual == SodaDataTypeName.VARCHAR:
             logger.debug(


soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
```diff
@@ -141,6 +141,15 @@ def evaluate(self, measurement_values: MeasurementValues) -> CheckResult:
             diagnostic_metric_values=diagnostic_metric_values,
         )
 
+    def get_threshold_metric_impl(self) -> Optional[MetricImpl]:
+        return self.failed_rows_count_metric_impl
+
+    def is_expression_check(self) -> bool:
```
paulteehan, line 147: Oh I like these, adds clarity

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
```diff
@@ -90,14 +90,21 @@ def get_soda_data_type_name_by_data_source_data_type_names(self) -> dict[str, So
...
             "char": SodaDataTypeName.CHAR,
             "text": SodaDataTypeName.TEXT,
+            "bpchar": SodaDataTypeName.TEXT,
             "smallint": SodaDataTypeName.SMALLINT,
+            "int2": SodaDataTypeName.SMALLINT,
             "integer": SodaDataTypeName.INTEGER,
+            "int": SodaDataTypeName.INTEGER,
+            "int4": SodaDataTypeName.INTEGER,
             "bigint": SodaDataTypeName.BIGINT,
+            "int8": SodaDataTypeName.BIGINT,
```
paulteehan, line 100: I'm fine with this, however I thought it wasn't necessary to define `int4`, `int8` etc here because they're defined in `_get_data_type_name_synonyms`, what's the reason for adding them here?   Do I need to do something similar when I work on other data sources?

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
```diff
@@ -317,3 +317,22 @@ def map_test_sql_data_type_to_data_source(self, source_data_type: SqlDataType) -
         if sql_data_type.name == "varchar" and sql_data_type.character_maximum_length is None:
             sql_data_type.character_maximum_length = self.default_varchar_length()
         return sql_data_type
+
+    @classmethod
+    def is_same_soda_data_type_with_synonyms(cls, expected: SodaDataTypeName, actual: SodaDataTypeName) -> bool:
+        if expected == SodaDataTypeName.CHAR and actual == SodaDataTypeName.VARCHAR:
```
paulteehan, line 323: Again I'm fine with this, I'm wondering about the SODA_DATA_TYPE_SYNONYMS structure and if we could leverage that to do one generic check in the base DataSourceImpl, rather than having specific manual checks like this?

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
Niels-b: This is specifically for the data type mapping from Postgres cursor -> Soda. The cursor returns the code for `int4`. We could just as easily change the cursor mapping. But I think this covers more different cases when we catch it here.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
Niels-b: These specific checks are required for data source specific **one-way** mappings. Two way mappings are already covered by the `SODA_DATA_TYPE_SYNONYMS` structure and the default implementation for that.

PR #2439, opened 2025-10-15
remove distutils in order to support python 3.12

### LINKED ISSUE #2091: distutils is deprecated. distutils.util.strtobool needs replacement.

vebjorre:
Finishing the work of @akx in #2144, removing any use of distutils in order to support python>=3.12.

Fixes #2091

### DIFF
--- a/soda/core/tests/helpers/data_source_fixture.py
+++ b/soda/core/tests/helpers/data_source_fixture.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import logging
 import os
 import re
@@ -10,7 +11,6 @@ from helpers.test_column import TestColumn
 from helpers.test_scan import TestScan
 from helpers.test_table import TestTable
 from soda.common.lazy import Lazy
-from soda.common.random_helper import generate_random_alpha_num_str
 from soda.common.yaml_helper import YamlHelper
 from soda.execution.data_source import DataSource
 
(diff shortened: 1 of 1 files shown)


soda/core/soda/common/env_helper.py
```diff
@@ -1,12 +1,15 @@
...
-from distutils.util import strtobool
 
 from dotenv import load_dotenv
 from soda.common.logs import Logs
 
 
+def strtobool(val: str) -> bool:
+    return val.lower() in ("yes", "y", "true", "t", "1")
+
+
```
paulteehan, line 21: Hello @vebjorre , thanks very much for this contribution!  We'd like to request two small edits after which we'd be happy to merge and release this.  

1. It looks like `"on"` should be in the tuple
2. The original `strtobool` implementation throws a ValueError if you pass in an invalid value, we'd like to keep that behavior

If you could make those edits then we'll move this forward, thanks again!

paulteehan: Thank you, that's perfect! 

I'm seeing a failure in our CI, I think it's probably spurious but I'll have to investigate that before we can merge.  I'll investigate next week and update you.

paulteehan: I'll need to make a small change to our workflow file to update some credentials, should be able to get to it this week, thanks for your patience.

paulteehan: I had to move some other things ahead of this in my to-do list, I will come back to this as soon as I can, and I'll keep you updated.  Thank you

PR #2440, opened 2025-10-15
R-288: send list of post-processing tasks to backend

milanaleksic:
This is a preliminary change in soda core that would allow us to run the plugins as DWH and report them as post-processing stages of the main scan logic. This is needed to drive further the DWH logs improvement epic - we need to know that the post-processing stages exist so that we can handle scan in a way that can allow for further follow ups with logs, post-processing stage updates etc.

This particular PR is the first in chain: I still need to handle the `sodaCorePostProcessingUpdate` calls, but I am already sharing since it has an important change: it allows registration of the plugins into the verification handlers, and avoids import cycle, causing much nicer API (I think). More will be built on top of this.

### DIFF
--- a/soda-core/src/soda_core/contracts/contract_verification.py
+++ b/soda-core/src/soda_core/contracts/contract_verification.py
@@ -406,8 +406,8 @@ class ContractVerificationStatus(Enum):
 
 
 class PostProcessingStageState(Enum):
-    ONGOING = ("ongoing",)
-    COMPLETED = ("completed",)
+    ONGOING = "ongoing"
+    COMPLETED = "completed"
     FAILED = "failed"
 
 


soda-tests/tests/conftest.py
```diff
@@ -41,3 +44,12 @@ def now(tz=None):
 ]
 freezegun.configure(extend_ignore_list=my_ignore_list)
 freeze_time = freezegun.freeze_time
+
+
+@pytest.fixture(autouse=True)
```
nielsn, line 49: Nice one, didn't know about this.

milanaleksic:
> a handler provides many stages) but in reversing you're basically saying that a stage is "unique" and will have a single handler dealing with it.

I think that I haven't add enough thought about the future in this case here and wanted to be a bit more open on the interface side: I assumed that each plugin can have multiple (post-processing) stages, but that only one can be used (so if they conflict that's a problem - but I am not fixing the problem, just reporting it as a warning for now since I don't know a better solution). If you think it would make more sense to have only 1 stage per plugin for me that's fine, I am not sure either about returning an array, it was maybe a step too far, WDYT @nielsn?

milanaleksic:
> * The problem is that DWH runs after scans as its own process and we have no way of interacting with it from the platform, e.g logging and monitoring
> * This PR proposes a new mechanism to define a list of "PostProcessingStages" attached to a scan, of which DWH can be one
> * All that's required is the name, and the state (ongoing, completed, failed).  I guess the state will be pushed to soda cloud upon change

Yes, that's the other PR in which I am handling the possible outcomes of the stage: it can succeed or fail. 
In core, however, we need to keep the state of the scan updated with the fact that _we do have some postprocessing stages that might have work or not (we don't know yet)_. This will allow more logs to come in, specifically for those post-processing stages, independent from the main scan logs.

> 
> This seems like a very useful and important conceptual change, in effect we begin to consider our modeling as a pipeline rather than a single process. Is this a kind of DAG? Are there assumptions built-in about the stages being dependent or independent from one another? It's maybe not a question for right now, I'm not suggesting any changes, just more to understand the vision

There are no assumptions being made at this time. We are just fitting a well-known concept of stages into our scan workflow on the backend, but not committing on having (maybe ever) additional stages. Or we can decide to add for example backend notification sending as a post-processing stage (handled completely on backend, just different components) or maybe updating our public API as another post-processing stage, etc. The idea is indeed that we open up a bit more to the customers what backend and core and other processes actually do.

thanks for the review @paulteehan

PR #2442, opened 2025-10-16
R-288: implement post-processing stage update even in the case of failures in post-processing

milanaleksic:
In this PR I am adding support for failure detection & forwarding to the cloud in case a post-processing stage fails.
The same code in soda cloud client will be used by the extension to actually inform of success as well (incoming PR)

### DIFF
--- a/soda-tests/tests/components/test_soda_cloud.py
+++ b/soda-tests/tests/components/test_soda_cloud.py
@@ -334,7 +334,7 @@ def test_soda_cloud_results_with_post_processing_with_failure(
 
     assert (
         len(data_source_test_helper.soda_cloud.requests) == 3
-    ), f"Expected 2 cloud requests, got more: {data_source_test_helper.soda_cloud.requests}"
+    ), f"Expected 3 cloud requests, got more: {data_source_test_helper.soda_cloud.requests}"
 
 
 def test_execute_over_agent(data_source_test_helper: DataSourceTestHelper):


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1057,13 +1058,34 @@ def migration_upload_contracts(self, migration_id: str, contracts: list["Contrac
...
         response = self._execute_command(request, request_log_name="upload_migration_contracts")
-        response_dict = response.json()
+        response.json()  # verify response is in JSON format
 
         if response.status_code != 200:
             raise SodaCloudException(f"Failed to upload migration contracts: {response}")
 
         logger.info(f"Uploaded {len(contracts)} contracts for migration {migration_id}")
 
+    def post_processing_update(
```
paulteehan, line 1231: I imagined there would be a mechanism for displaying DWH logs in the product while the process is ongoing, but so far the only mechanism I see to update state in the product is by setting 'state'.    What if we want to send checkpoint messages to the user from within the DWH process, like 'processing xyz', '50% complete', etc, do you imagine just using logs for this?  Is there some way of distinguishing customer-facing logs from engineer-facing logs?

soda-tests/tests/components/test_soda_cloud.py
```diff
@@ -251,39 +241,76 @@ def provides_post_processing_stages(self) -> list[PostProcessingStage]:
...
+            raise RuntimeError("Intentional failure for testing")
+
+        def provides_post_processing_stages(self) -> list[PostProcessingStage]:
+            return [PostProcessingStage("testStage", PostProcessingStageState.ONGOING)]
+
+    ContractVerificationHandlerRegistry.register(DummyHandler())
+
+    env_vars["SODA_SCAN_ID"] = "env_var_scan_id"
+
+    data_source_test_helper.enable_soda_cloud_mock(
```
paulteehan, line 280: FYI there is a dedicated test helper for DWH, `DWHTestSetup`, it has some new methods that can reduce boilerplate in tests including the soda cloud mock.  Usage example https://github.com/sodadata/soda-extensions/blob/main/soda/tests/all_data_sources/test_failed_rows_duplicate.py

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -620,17 +622,18 @@ def verify(self) -> ContractVerificationResult:
...
-            scan_id: Optional[str] = soda_cloud_response_json.get("scanId") if soda_cloud_response_json else None
+            scan_id = soda_cloud_response_json.get("scanId") if soda_cloud_response_json else None
             if not scan_id:
                 contract_verification_result.sending_results_to_soda_cloud_failed = True
         else:
             logger.debug(f"Not sending results to Soda Cloud {Emoticons.CROSS_MARK}")
 
-        try:
-            for contract_verification_handler in ContractVerificationHandlerRegistry.contract_verification_handlers:
+        for contract_verification_handler in ContractVerificationHandlerRegistry.contract_verification_handlers:
```
paulteehan, line 635: what happens if handler #2 depends on #1, and #1 failed?  It seems here is a baked-in assumption that the stages can run independently

soda-tests/tests/components/test_soda_cloud.py
```diff
@@ -292,6 +319,23 @@ def provides_post_processing_stages(self) -> list[PostProcessingStage]:
...
+            "scanId": "env_var_scan_id",
+            "name": "testStage",
+            "state": "failed",
+            "error": matcher_string_contains("RuntimeError: Intentional failure for testing"),
+        },
+    )
+
+    assert (
+        len(data_source_test_helper.soda_cloud.requests) == 3
+    ), f"Expected 2 cloud requests, got more: {data_source_test_helper.soda_cloud.requests}"
```
nielsn:
```suggestion
    ), f"Expected 3 cloud requests, got more: {data_source_test_helper.soda_cloud.requests}"
```

soda-core/src/soda_core/common/soda_cloud.py
milanaleksic: This is correct observation indeed. We need the state to be handled first, that's R-288, and then there's R-294 which will handle actual sending of logs. We currently can't send logs since the logs are completely ingested into object storage by backend when scan completes, so we need first this work that signals to backend "hey, don't package and send, wait a bit longer until post-processing stages are complete".

> What if we want to send checkpoint messages to the user from within the DWH process, like 'processing xyz', '50% complete', etc, do you imagine just using logs for this?

They way what I planned on doing is to implement the same algorithm that Milan Lukacz implemented for batched log sending as part of the specification for batched ingestion, in which in core every time  we log something we verify that either:
1. enough time has passed since the last time we sent the log batch
2. log batch is full
3. scan is complete (no more logs to send)
and if any of those things is true we will send a batch of logs. This means that we should from time to time get logs show up on the backend/frontend and the users will be able to get more of a real-time feeling than currently with "big bang log upload". By the way, if this experiment works I will integrate the same logic for main log processing in soda core v4, but we'll see how much time is still left (if any) for that work.

soda-tests/tests/components/test_soda_cloud.py
milanaleksic: Thanks, I will look into it and I will update it with more features (e.g. making it possible to have DWH _disabled_), add the state update for the post-processing and similar. I am working on PR no 4 to add more tests and it will be incorporated there

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
milanaleksic: This is a completely valid remark and my answer is: I haven't considered this to be a graph and haven't taken any steps into implementing support for it, but if it makes sense to add it I will. Since we currently only have 1 possible stage it doesn't immediately demand a change, but I like how this is going, it is definitely a good point to keep in mind for the next work that might add new stages, thanks

PR #2443, opened 2025-10-16
catch schema drop errors

paulteehan:
Fixes errors in CI due to race conditions when dropping schemas


paulteehan:
Yes we should address it but it takes some effort, and extensions CI is blocked by this, I think more important to unblock now and give this a proper look in cooldown

paulteehan:
Yes we should address it but it takes some effort, and extensions CI is blocked by this, I think more important to unblock now and give this a proper look in cooldown

paulteehan:
Yes something has changed, because some tests in extensions are now more strict and will fail if there are unexpected errors in logs, the schema race condition was causing tests to fail, I advised Milan to disable tests so he could get his PR through, this fix bypasses the problem for now

PR #2449, opened 2025-10-20
R-294: implement batch log cloud push

milanaleksic:
This is mostly work done by Jarek while he was helping out with python backlog in soda library.
I have adapted it to use `LogRecord` instead of `Log` because it fits better to the way how Soda Core v4 code is written.

At this time no effort was made to move soda core v4 to use the LogsQueue, the intention is to build the capability and to use that capability in the DWH extension where logs during DWH stage logs will be sent to the cloud.

The feature, thus, is not directly used for now inside soda-core, but only in soda-extensions project (PR incoming)

### DIFF
--- a/soda-tests/tests/features/test_logs.py
+++ b/soda-tests/tests/features/test_logs.py
@@ -1,4 +1,5 @@
 import logging
+
 from soda_core.common.logs import Logs
 
 
@@ -10,10 +11,10 @@ def test_catch_args_in_logs_messages():
     l.remove_from_root_logger()
     logs = l.get_logs()
     assert len(logs) == 3, f"Expected 3 error log, got {len(logs)}"
-    assert 'This is a test message 1' in logs
+    assert "This is a test message 1" in logs
     assert "This is a test message 2" in logs
     assert "This is a test message 2" in logs
 
     error_logs = l.get_errors()
     assert len(error_logs) == 1, f"Expected 1 error log, got {len(error_logs)}"
-    assert 'This is a test message 3' in error_logs, f"Expected error log message not found in {error_logs}"
+    assert "This is a test message 3" in error_logs, f"Expected error log message not found in {error_logs}"


soda-core/src/soda_core/common/logs.py
```diff
@@ -67,13 +79,16 @@ def get_errors_str(self) -> str:
         return "\n".join(self.get_errors())
 
     def get_errors(self) -> list[str]:
-        return [r.msg for r in self.records if r.levelno >= ERROR]
+        return [r.msg for r in self.gatherer.get_error_logs()]
```
paulteehan: With the Thrift backend error today, the logs that were captured were less useful because `msg` contained a `%s` parameter with a value that was not expanded.  The expansion was in `args`, or there was a fully expanded message in `message`.  I don't know enough about Python logging to say if can/should reliably use `message` instead of `msg`, but it would be best if the logs we capture include any additional info.  For reference here are the printed contents one of the error log records:

```
>           raise Exception("Unexpected error: " + error_msg)
E           Exception: Unexpected error: name: ['databricks.sql.backend.thrift_backend']
E           msg: ['ThriftBackend.attempt_request: Exception: %s']
E           args: [(ValueError('I/O operation on closed file.'),)]
E           levelname: ['ERROR']
E           levelno: [40]
E           pathname: ['/home/runner/work/soda-extensions/soda-extensions/.venv/lib/python3.10/site-packages/databricks/sql/backend/thrift_backend.py']
E           filename: ['thrift_backend.py']
E           module: ['thrift_backend']
E           exc_info: [None]
E           exc_text: [None]
E           stack_info: [None]
E           lineno: [477]
E           funcName: ['attempt_request']
E           created: [1761036706.7389688]
E           msecs: [738.0]
E           relativeCreated: [289637.84527778625]
E           thread: [140131613236096]
E           threadName: ['MainThread']
E           processName: ['MainProcess']
E           process: [8940]
E           message: ['ThriftBackend.attempt_request: Exception: I/O operation on closed file.']
```

milanaleksic: While developing this feature I got runtime errors if I tried to use `message` instead of `msg` and I assumed that's how things are. But, based on your comment I went down the rabbit hole a bit deeper and noticed that there is indeed a method `getMessage()` which apparently does catch this syntax of error reporting.
Thanks for catching this!

I have also changed on other places as well the fact that we are using `.msg` instead of `.getMessage()`.

Here is a simplified example 

```
import logging
from logging import Handler


class LogCapturer(Handler):
    def __init__(self):
        super().__init__()
        logging.root.addHandler(self)

    def emit(self, record: logging.LogRecord):
        print(record.msg)
        print(record.getMessage())

LogCapturer()
logging.warning("This is a test %s", "message")
```

PR #2453, opened 2025-10-22
Add Spark Dataframe compatibility

Niels-b:


soda-sparkdf/tests/data_sources/test_sparkdf.py
```diff
@@ -0,0 +1,121 @@
...
+
+def test_more_complex_setup_with_existing_session(data_source_test_helper: DataSourceTestHelper):
+    my_spark_session = (
+        SparkSession.builder.master("local")
+        .appName("test_sparkdf")
+        .config("spark.sql.warehouse.dir", data_source_test_helper.test_dir)
+        .getOrCreate()
+    )
+
+    my_data_source_impl = SparkDataFrameDataSourceImpl.from_existing_session(
```
m1n0: nitpick: can we use SparkDataFrameDataSource alias to make the test an even nicer example for docs/users, but also to cover the exported alias code path?

soda-sparkdf/src/soda_sparkdf/test_helpers/sparkdf_data_source_test_helper.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Called in _create_data_source_impl to initialized self.data_source_impl
+        self.database_name and self.schema_name are available if appropriate for the data source type
+        """
+        self.test_dir = tempfile.mkdtemp(prefix=f"soda_test_sparkdf_{self.name}_")
+        return f"""
+            type: sparkdf
+            name: {self.name}
+            connection:
+                new_session: true
```
m1n0, line 20: at first I wasn't sure the `new_session` config would be that useful, but I like your thinking here - this way there is a generic session-creation logic in the datasource, and helper does not need to do anything special to set this up 👍

soda-sparkdf/src/soda_sparkdf/common/data_sources/sparkdf_data_source.py
```diff
@@ -0,0 +1,227 @@
...
+            session = config.spark_session
+        elif isinstance(config, SparkDataFrameNewSessionProperties):
+            session = (
+                SparkSession.builder.master("local")
+                .appName(self.name)
+                .config("spark.sql.warehouse.dir", config.test_dir)
+                .getOrCreate()
+            )
+            session.sql("SET spark.sql.session.timeZone = +00:00;")
+            session.sql("SET TIME ZONE 'UTC';")
```
m1n0, line 190: does this session timezone do anything now that we decided to go with the freeze gun approach? Or are we keeping this just in case, as it makes sense to run utc sessions?

m1n0 (approved):

soda-sparkdf/src/soda_sparkdf/test_helpers/sparkdf_data_source_test_helper.py
Niels-b: Figured it was easier to do it here, as it plugs in nicely to the rest of the testing flow :).

soda-sparkdf/src/soda_sparkdf/common/data_sources/sparkdf_data_source.py
Niels-b: I would keep them just in case. It is tested and verified with this config now, so let's keep that 😄

PR #2458, opened 2025-10-23
Dtl 1248 subset checks

m1n0:
Adds the ability to run subset of checks.

- Multiple check paths are accepted e.g. `... --check-paths checks.freshness checks.row_count`
- Invalid check paths are ignored
- Scans are marked as `partial = True` if any of the checks is Excluded
- All checks are still sent to Cloud in new `excluded` state
- api, counters, scan summary is adjusted with the new state, counters etc
- diagnostics are still sent to Cloud since they are required, v4 diagnostics are set to `None`. 
- a small refactor to separate check impl creation from setting up metrics was needed in order to control this behaviour in one spot

Notes for docs:
- check paths are accepted as an optional multi argument e.g. `... --check-paths checks.freshness checks.row_count`
- qualifier can be used as well. The potentially non-intuitive part is that the path still has the original check in it, qualifier is added on top e.g. `... --check-paths checks.freshness.freshness_qualifier`
- if empty check paths `... --check-paths --verbose ...`is used then it's ignored, i.e. all checks are executed
- if a non-matching check path is present it gets ignored e.g. with `... --check-paths does.not.exist` no checks will be executed at all. Another example: `... --check-paths checks.freshness does.not.exist` - only the freshness check will be executed.
- the shortcut argument name is `-cp`, so e.g. `... -cp checks.freshness`
- Impact on UI: ui always shows the latest results, the skipped/excluded ones using this feature are not deleted, but there is no "excluded" state shown in UI. We just don't show them.

### DIFF
--- a/soda-tests/tests/features/test_check_paths_filter.py
+++ b/soda-tests/tests/features/test_check_paths_filter.py
@@ -89,13 +89,12 @@ def test_check_paths_qualifier(data_source_test_helper: DataSourceTestHelper):
         contract_yaml_str="""
             columns:
                 - name: id
-                    checks:
+                  checks:
                     - invalid:
                         valid_values: [1, 2, 3]
                     - invalid:
                         qualifier: invalid_min
                         valid_min: 1
-            checks:
         """,
     )
     assert len(contract_verification_result.check_results) == 2


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1439,6 +1448,10 @@ def _build_diagnostics_json_dict(check_result: CheckResult) -> Optional[dict]:
 
 
 def _build_v4_diagnostics_check_type_json_dict(check_result: CheckResult) -> Optional[dict]:
+    if check_result.outcome == CheckOutcome.EXCLUDED:
+        # TODO: can we come up with any diagnostics for excluded checks?
```
Niels-b, line 1452: Still open TODO here.
For me personally, I wouldn't expect diagnostics for any excluded checks.

soda-tests/tests/features/test_check_paths_filter.py
```diff
@@ -0,0 +1,106 @@
...
+def test_check_paths_qualifier(data_source_test_helper: DataSourceTestHelper):
+    test_table = data_source_test_helper.ensure_test_table(test_table_specification)
+
+    data_source_test_helper.enable_soda_cloud_mock(
+        [
+            MockResponse(status_code=200, json_object={"fileId": "a81bc81b-dead-4e5d-abff-90865d1e13b1"}),
+        ]
+    )
+
+    with freeze_time(datetime(year=2025, month=1, day=3, hour=10, minute=0, second=0, tzinfo=timezone.utc)):
```
Niels-b: Do we need `freeze_time` here?

soda-core/src/soda_core/common/soda_cloud.py
m1n0: I agree, but it's not possible - v4 diagnostics can be ommited, but the whole diagnostics block cannot as it's required by the api. Keeping this in for now because of that

soda-tests/tests/features/test_check_paths_filter.py
m1n0: ah good catch indeed,  no need here as I didn't include freshness checks here.

PR #2460, opened 2025-10-27
Add ALTER TABLE to AST

Niels-b:
Includes:
- ADD COLUMN
- DROP COLUMN (for supported data sources)

Important other changes:
- Use `DESCRIBE TABLE` for Databricks instead of the information schema. The latter is delayed for some tables, so we don't get accurate column metadata.

### DIFF
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
@@ -239,17 +239,10 @@ class DatabricksSqlDialect(SqlDialect):
             return True
         return super().is_same_soda_data_type_with_synonyms(expected, actual)
 
-    def _build_alter_table_add_column_sql(self, alter_table: ALTER_TABLE_ADD_COLUMN, add_semicolon: bool = True) -> str:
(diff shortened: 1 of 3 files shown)


soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
```diff
@@ -178,31 +189,14 @@ def _get_data_type_name_synonyms(self) -> list[list[str]]:
...
-        if expected == SodaDataTypeName.TIME and actual == SodaDataTypeName.TEXT:
-            logger.debug(
-                f"In is_same_soda_data_type_with_synonyms, Expected {expected} and actual {actual} are the same"
-            )
-            return True
-        return super().is_same_soda_data_type_with_synonyms(expected, actual)
-
-
-class DatabricksHiveSqlDialect(DatabricksSqlDialect):
+    # We move to DESCRIBE TABLE, that is a more up-to-date way to get the columns metadata. (information_schema is lagging behind sometimes, and does not always return the correct columns)
```
m1n0, line 199: oh no, so it was futile to have these two separated... @mivds aren't there any performance consequences to this though?

m1n0 (approved):

PR #2461, opened 2025-10-28
Add `dataset_id` and `check_id` to `ContractVerificationResult`

Niels-b:
As per request, add `dataset_id` and `check_id` to the `ContractVerificationSessionResult`. 
Note; the `ContractVerificationSessionResult` consists of multiple `ContractVerificationResult`. The latter is the class that contains the dataset and check ids.

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
@@ -626,6 +626,7 @@ class ContractImpl:
                     local_file_path=self.contract_yaml.contract_yaml_source.file_path,
                     soda_cloud_file_id=soda_cloud_file_id,
                 ),
+                dataset_id=None,  # Set this to None for now (default). Will be filled later when we get the dataset id from Soda Cloud
             ),
             data_source=data_source,
             data_timestamp=self.data_timestamp,
@@ -648,7 +649,8 @@ class ContractImpl:
                 contract_verification_result.sending_results_to_soda_cloud_failed = True
             else:
                 contract_verification_result.scan_id = scan_id
-                contract_verification_result.dataset_id = self.__get_dataset_id(
+                # Put the dataset id in the contract object
+                contract_verification_result.contract.dataset_id = self.__get_dataset_id(
                     soda_cloud_response_json, self.soda_qualified_dataset_name
                 )
         else:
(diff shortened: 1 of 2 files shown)


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -646,6 +646,9 @@ def verify(self) -> ContractVerificationResult:
             scan_id = soda_cloud_response_json.get("scanId") if soda_cloud_response_json else None
             if not scan_id:
                 contract_verification_result.sending_results_to_soda_cloud_failed = True
+            else:
+                contract_verification_result.scan_id = scan_id
+                contract_verification_result.dataset_ids = self.__get_dataset_ids(soda_cloud_response_json)
```
milanaleksic: Alternative would be to match the contract.dataset name in the response and return only the dataset ID of the dataset that corresponds to the dataset of the contract. That way is maybe more user friendly: the user would have contract.{dataset_name, dataset_id} beside each other in the same object. And we can then hide the complexity of the multiple dataset IDs

Niels-b: Thanks for the suggestion, it has been updated 👍

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -646,6 +646,11 @@ def verify(self) -> ContractVerificationResult:
             scan_id = soda_cloud_response_json.get("scanId") if soda_cloud_response_json else None
             if not scan_id:
                 contract_verification_result.sending_results_to_soda_cloud_failed = True
+            else:
+                contract_verification_result.scan_id = scan_id
+                contract_verification_result.dataset_id = self.__get_dataset_id(
```
milanaleksic: I actually meant contract_verification_result.contract.dataset_id, not contract_verification_result.dataset_id. Why? Because in contract_verification_result.contract we already have dataset-related stuff on one place:

```
class Contract:
    data_source_name: Optional[str]
    dataset_prefix: list[str]
    dataset_name: str
    soda_qualified_dataset_name: str
    source: YamlFileContentInfo
```

Niels-b: Ah okay, now I got it. Thanks for the feedback! It has been implemented that way.

PR #2467, opened 2025-10-31
Fixes for SparkDF schema check

Niels-b:


Niels-b:
For future reference; got all clear from Tyler ("Code looks good and test confirms it works").

PR #2469, opened 2025-11-05
Add schema query and function to verify if schema exists

Niels-b:
Needed to support feature in DWH. We need to be able to check if a schema has already been created.
This PR adds that functionality for all the datasources.

### DIFF
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -1009,7 +1009,7 @@ class SqlDialect:
         """
         return self.default_casify("schemata")
 
-    def column_catalog_name(self) -> str:
+    def column_schemata_catalog_name(self) -> str:
         """
         Name of the column that has the database/catalog information in the schemata metadata table.
         Purpose of this method is to allow specific data source to override.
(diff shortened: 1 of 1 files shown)


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -1002,6 +1002,27 @@ def table_columns(self) -> str:
...
         return self.default_casify("columns")
 
+    def table_schemata(self) -> str:
+        """
+        Name of the table that has the schema information in the metadata.
+        Purpose of this method is to allow specific data source to override.
+        """
+        return self.default_casify("schemata")
+
+    def column_catalog_name(self) -> str:
```
mivds: Not a huge fan of the naming here. This function is called `column_catalog_name` and below we have `column_table_catalog`. They seem pretty close and I wouldn't immediately know which one to use.

Not really sure how to fix it without large refactor, but maybe we can name this `column_schemata_catalog_name` to indicate it applies to the `schemata` table

Niels-b: Rename is a good suggestion, I will do that 👍

PR #2474, opened 2025-11-13
Updates to CLI and programatic API

Niels-b:
Updated contract API to allow only a single contract file. Includes preparatory changes to publish API.

Tried to cover as much from the CLI and API wrt to contract yaml files as possible. If this is too much for now, or not fully complete. Please let me know!

### DIFF
--- a/soda-core/src/soda_core/cli/cli.py
+++ b/soda-core/src/soda_core/cli/cli.py
@@ -110,7 +110,7 @@ def _setup_contract_verify_command(contract_parsers) -> None:
         "-c",
         "--contract",
         type=str,
-        help="One contract file path. Use this to work with local contracts.",
+        help="Contract file path to verify. Use this to work with local contracts.",
     )
     verify_parser.add_argument(
         "-d",
@@ -234,7 +234,7 @@ def _parse_variables(variables: Optional[List[str]]) -> Optional[Dict[str, str]]
 
 def _setup_contract_publish_command(contract_parsers) -> None:
     publish_parser = contract_parsers.add_parser("publish", help="Publish a contract")
-    publish_parser.add_argument("-c", "--contract", type=str, help="One contract file path.")
+    publish_parser.add_argument("-c", "--contract", type=str, help="Contract file path to publish.")
 
     publish_parser.add_argument(
         "-sc",
@@ -264,7 +264,7 @@ def _setup_contract_publish_command(contract_parsers) -> None:
 
 def _setup_contract_test_command(contract_parsers) -> None:
     test_contract_parser = contract_parsers.add_parser(name="test", help="Test a contract syntax without executing it")
-    test_contract_parser.add_argument("-c", "--contract", type=str, help="One contract file path.")
+    test_contract_parser.add_argument("-c", "--contract", type=str, help="Contract file path to test.")
 
     test_contract_parser.add_argument(
         "-v",
--- a/soda-core/src/soda_core/contracts/api/verify_api.py
+++ b/soda-core/src/soda_core/contracts/api/verify_api.py
@@ -33,7 +33,7 @@ def verify_contracts_locally(
     check_paths: Optional[list[str]] = None,
     dwh_data_source_file_path: Optional[str] = None,
 ) -> ContractVerificationSessionResult:
-    if not isinstance(contract_file_paths, list):
+    if isinstance(contract_file_paths, str):
         contract_file_paths = [contract_file_paths]
 
     if len(contract_file_paths) > 1:
@@ -104,7 +104,7 @@ def verify_contracts_on_agent(
     verbose: bool = False,
     blocking_timeout_in_minutes: int = 60,
 ) -> ContractVerificationSessionResult:
-    if not isinstance(contract_file_paths, list):
+    if isinstance(contract_file_paths, str):
         contract_file_paths = [contract_file_paths]
 
     if len(contract_file_paths) > 1:
@@ -183,17 +183,17 @@ def verify_contract(
     else:
         dataset_identifiers = None
 
+    # TODO: change this when we fully deprecate a list of contract file paths
+    # This is only here to make sure the rest of the code still works.
+    if isinstance(contract_file_path, str):
+        contract_file_paths = [contract_file_path]
+    else:
+        contract_file_paths = None
+
     soda_cloud_client: Optional[SodaCloud] = None
     try:
         if soda_cloud_file_path:
             soda_cloud_client = SodaCloud.from_config(soda_cloud_file_path, variables)
-
-        # TODO: change this when we fully deprecate a list of contract file paths
-        # This is only here to make sure the rest of the code still works.
-        if isinstance(contract_file_path, str):
-            contract_file_paths = [contract_file_path]
-        else:
-            contract_file_paths = None
 
         # TODO: verify the path where connection is provided
         validate_verify_arguments(


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -470,6 +470,7 @@ def publish_contract(self, contract_yaml: Optional[ContractYaml]) -> ContractPub
                     soda_cloud_file_id=response_json.get("fileId", None),
                 ),
                 soda_qualified_dataset_name=contract_yaml.dataset,
+                dataset_id=response_json.get("datasetId", None),
```
LaurenDebruyn: Could you do the same for the verify response? Or was there a specific reason why you want to get this from the checks?

soda-core/src/soda_core/contracts/api/publish_api.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
+
+from soda_core.contracts.contract_publication import (
+    ContractPublication,
+    ContractPublicationResultList,
+)
+
+
+def publish_contract(
+    contract_file_path: Optional[str], soda_cloud_file_path: Optional[str]
```
LaurenDebruyn: How can `contract_file_path` be optional?

soda-core/src/soda_core/contracts/api/test_api.py
```diff
@@ -10,11 +11,14 @@
 soda_telemetry = SodaTelemetry()
 
 
-def test_contracts(
+def test_contract(
     contract_file_paths: Union[str, list[str]],
     variables: Optional[Dict[str, str]] = None,
     verbose: bool = False,
+    contract_file_path: Optional[str] = None,
```
LaurenDebruyn: Shouldn't this replace `contract_file_paths` and not be optional?

soda-core/src/soda_core/contracts/api/verify_api.py
LaurenDebruyn, line 28: We should also have a new argument `dataset_identifier`, you either pass a contract file path or dataset identifier.

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -18,7 +18,7 @@
 soda_telemetry = SodaTelemetry()
 
 
-def verify_contracts_locally(
+def verify_contract_locally(
     data_source_file_path: Optional[str] = None,
     data_source_file_paths: list[str] = [],
```
LaurenDebruyn: You can remove this because we only support a single datasource per contract. And make `data_source_file_path` not optional.

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -81,8 +86,33 @@ def verify_contracts_on_agent(
...
+            contract_file_path = contract_file_paths[0]
+        else:
+            contract_file_path = contract_file_paths
+
+    return contract_file_path
+
+
+def verify_contract(
+    contract_file_path: Optional[str],
     dataset_identifiers: Optional[list[str]],
```
LaurenDebruyn: You either pass one contract file path or one dataset identifier. We should probably add a check for that as well in the method.
```suggestion
    dataset_identifiers: Optional[str],
```

soda-core/src/soda_core/contracts/api/verify_api.py
LaurenDebruyn: Remove

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -18,7 +18,7 @@
 soda_telemetry = SodaTelemetry()
 
 
-def verify_contracts_locally(
+def verify_contract_locally(
```
LaurenDebruyn: Either we are backwards compatible, or we are not. But renaming the function makes it per definition not backwards compatible IMO.
My proposal would be that you keep the existing function as is and mark it as deprecated (can you easily do this in Python?). We can still raise an exception if you try to verify more than one contract at once. I would create an entirely new method for the singular flow though.

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -81,8 +86,33 @@ def verify_contracts_on_agent(
...
+    if contract_file_paths and not contract_file_path:
+        if isinstance(contract_file_paths, list):
+            if len(contract_file_paths) > 1:
+                raise InvalidArgumentException("Only one contract is allowed at a time")
+            # Only 1 contract found, use it.
+            contract_file_path = contract_file_paths[0]
+        else:
+            contract_file_path = contract_file_paths
+
+    return contract_file_path
```
LaurenDebruyn: A few suggestions here at once:
* Adding a check to make sure at least one of the two arguments is not `None`
* Simplifying the flow by re-arranging the if-statements and early returns

```suggestion
    if not contract_file_path and not contract_file_paths:
        raise InvalidArgumentException(
            "..."
        )
        
    if contract_file_path and contract_file_paths:
        raise InvalidArgumentException(
            "Do not provide contracts in both `contract_file_paths` and `contract_file_path`"
        )
        
    if contract_file_paths:
        if isinstance(contract_file_paths, list):
            if len(contract_file_paths) > 1:
                raise InvalidArgumentException("Only one contract is allowed at a time")
            return contract_file_paths[0]
        else:
            return contract_file_paths
            
    return contract_file_path
```

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -109,8 +139,12 @@ def verify_contracts(
         if soda_cloud_file_path:
             soda_cloud_client = SodaCloud.from_config(soda_cloud_file_path, variables)
 
-        if isinstance(contract_file_paths, str):
-            contract_file_paths = [contract_file_paths]
+        # TODO: change this when we fully deprecate a list of contract file paths
+        # This is only here to make sure the rest of the code still works.
```
LaurenDebruyn: Are you planning a follow-up PR for this? I wouldn't keep this code around for too long because we don't use multiple contracts anymore.

soda-core/src/soda_core/contracts/contract_publication.py
```diff
@@ -52,7 +52,7 @@ class Contract:
     dataset_name: str
     soda_qualified_dataset_name: str
     source: YamlFileContentInfo
-    id: str
+    dataset_id: Optional[str] = None
```
LaurenDebruyn: Is this optional? We always get it back from Cloud.

soda-core/src/soda_core/contracts/api/verify_api.py
nielsn: I tend to agree here. Split into two methods, keeping the original signature in one, introducing new signature in the other. 

The old method would then perform the additional checks (length of the contract_file_paths array should be one).

soda-core/src/soda_core/contracts/api/verify_api.py
nielsn: Correct, again to introduce in new method.
Same array length validation applies as with `contract_file_paths`

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -30,15 +30,18 @@ def verify_contracts_locally(
...
 ) -> ContractVerificationSessionResult:
     """
     Verifies the contract locally.
     """
     if not isinstance(data_sources, list):
         data_sources = [data_sources]
 
-    return verify_contracts(
-        contract_file_paths=contract_file_paths,
+    contract_file_path = _verify_contract_file_paths(contract_file_paths, contract_file_path)
```
nielsn: Unfortunate naming, `verify` has a bit of an additional meaning in this context.

Maybe something like `_ensure_single_contract_path`?

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -81,8 +86,33 @@ def verify_contracts_on_agent(
     )
 
 
-def verify_contracts(
-    contract_file_paths: Optional[Union[str, list[str]]],
+def _verify_contract_file_paths(
```
nielsn: Splitting the behavior (old and new) into multiple methods will take away a lot of the validation complexity, as the new method already has a signature aligned with the intended use case.

E.g., no need to see if there is one and only one of the arguments provided etc.

Note: I would also include the dataset identifiers here

soda-core/src/soda_core/contracts/api/verify_api.py
nielsn: Indeed, technically we don't support lists anymore. You could do the inverse here: 

```
        if isinstance(contract_file_path, list):
            contract_file_path = contract_file_path[0]
        else:
            contract_file_path = None
```

Doesn't have to part of this PR, but ideally soon-ish, as we will forget about this :-)

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -53,7 +56,7 @@ def verify_contracts_locally(
     )
 
 
-def verify_contracts_on_agent(
+def verify_contract_on_agent(
     soda_cloud_file_path: str,
     contract_file_paths: Optional[Union[str, list[str]]] = None,
```
nielsn, line 98: Same remark about splitting as above.

soda-core/src/soda_core/contracts/api/test_api.py
```diff
@@ -10,11 +11,14 @@
 soda_telemetry = SodaTelemetry()
 
 
-def test_contracts(
+def test_contract(
```
nielsn: Same remark about splitting this up

soda-core/src/soda_core/contracts/api/publish_api.py
nielsn: Both of these args should not be optional I think.

soda-core/src/soda_core/contracts/api/verify_api.py
Niels-b: Not in scope for now. But I agree that this should definitely be done!

soda-core/src/soda_core/contracts/api/verify_api.py
Niels-b: Yes, splitting the methods makes that a lot simpler and we can even completely remove the `_verify_contract_file_paths` method.

soda-core/src/soda_core/contracts/api/verify_api.py
Niels-b: Method was removed, but thanks for the feedback. Indeed; naming is confusing now that you mention it.

soda-core/src/soda_core/common/soda_cloud.py
Niels-b: Good suggestion, but out of scope for this PR. We will tackle that later

Niels-b:
Note: before merging still check the publish flow with new BE changes!

Niels-b:
Update: testing revealed that the `dataset_id` was not parsed correctly from the response. The latest commit fixes that. 
PR should be ready now.

soda-core/src/soda_core/cli/cli.py
```diff
@@ -110,15 +110,13 @@ def _setup_contract_verify_command(contract_parsers) -> None:
         "-c",
         "--contract",
         type=str,
-        nargs="+",
-        help="One or more contract file paths. Use this to work with local contracts.",
+        help="One contract file path. Use this to work with local contracts.",
```
LaurenDebruyn: NIT: I wouldn't emphasize the `one`
```suggestion
        help="Contract file path to verify. Use this to work with local contracts.",
```

soda-core/src/soda_core/cli/cli.py
```diff
@@ -236,7 +234,7 @@ def _parse_variables(variables: Optional[List[str]]) -> Optional[Dict[str, str]]
 
 def _setup_contract_publish_command(contract_parsers) -> None:
     publish_parser = contract_parsers.add_parser("publish", help="Publish a contract")
-    publish_parser.add_argument("-c", "--contract", type=str, nargs="+", help="One or more contract file paths.")
+    publish_parser.add_argument("-c", "--contract", type=str, help="One contract file path.")
```
LaurenDebruyn: Same here
```suggestion
    publish_parser.add_argument("-c", "--contract", type=str, help="Contract file path to verify.")
```

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -30,16 +32,53 @@ def verify_contracts_locally(
     publish: bool = False,
     check_paths: Optional[list[str]] = None,
     dwh_data_source_file_path: Optional[str] = None,
+) -> ContractVerificationSessionResult:
+    if not isinstance(contract_file_paths, list):
```
LaurenDebruyn: NIT: I would check for `str` instead of `list` to make it easier to read:

```suggestion
    if isinstance(contract_file_paths, str):
```

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -63,13 +103,46 @@ def verify_contracts_on_agent(
     publish: bool = False,
     verbose: bool = False,
     blocking_timeout_in_minutes: int = 60,
+) -> ContractVerificationSessionResult:
+    if not isinstance(contract_file_paths, list):
```
LaurenDebruyn: Same remark here:
```suggestion
    if isinstance(contract_file_paths, str):
```

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -104,13 +177,23 @@ def verify_contracts(
...
             soda_cloud_client = SodaCloud.from_config(soda_cloud_file_path, variables)
 
-        if isinstance(contract_file_paths, str):
-            contract_file_paths = [contract_file_paths]
+        # TODO: change this when we fully deprecate a list of contract file paths
+        # This is only here to make sure the rest of the code still works.
+        if isinstance(contract_file_path, str):
+            contract_file_paths = [contract_file_path]
+        else:
+            contract_file_paths = None
```
LaurenDebruyn: NIT: move this up next to the dataset_identifier backwards compatibility code block so that we have these workarounds nicely grouped together.

PR #2477, opened 2025-11-14
Dataset defining CTE: filtering and sampling, aggregate query support

m1n0:
- First iteration, only applied to Aggregate query
- CTE and the name of the CTE are constructed and kept in contract impl for reuse in other checks/queries/functionality
- support in other checks/queries AND tests will be added in subsequent PR

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
@@ -440,6 +440,7 @@ class ContractImpl:
             if (
                 self.dataset_configuration.test_row_sampler_configuration
                 and self.dataset_configuration.test_row_sampler_configuration.enabled
+                and self.dataset_configuration.test_row_sampler_configuration.test_row_sampler is not None
             ):
                 logger.info(
                     f"Row sampling is enabled for dataset {self.dataset_identifier.to_string()} "
@@ -1677,7 +1678,6 @@ class AggregationQuery(Query):
         cte_dataset_name: str,
         dataset_prefix: list[str],
         dataset_name: str,
-        # filter: Optional[str],
         data_source_impl: Optional[DataSourceImpl],
         logs: Logs,
     ):
--- a/soda-databricks/setup.py
+++ b/soda-databricks/setup.py
@@ -6,7 +6,10 @@ package_name = "soda-databricks"
 package_version = "4.0.0b12"
 description = "Soda Databricks V4"
 
-requires = [f"soda-core=={package_version}", "databricks-sql-connector"]
+requires = [
+    f"soda-core=={package_version}",
+    "databricks-sql-connector<4.2.0",  # Breaking change/bug with transactions in 4.2.0
+]
 
 setup(
     name=package_name,


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -396,6 +397,7 @@ def __init__(
         self.dataset_identifier = DatasetIdentifier.parse(contract_yaml.dataset)
         self.dataset_prefix: list[str] = self.dataset_identifier.prefixes
         self.dataset_name = self.dataset_identifier.dataset_name
+        self.cte_dataset_name: str = "filtered_dataset"
```
Niels-b, line 400: Note entirely convinced by this name. Maybe something like `_soda_source_dataset_cte`?
No hard opinion though.

Niels-b (commented):
Looks very good to me!
CTE part is very simple and expandable, I like it 👍 .

One minor comment about the name of the CTE, but no strong opinion.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -422,6 +424,33 @@ def __init__(
         )
         self.dataset_rows_tested: Optional[int] = None
 
+        self.cte = CTE(self.cte_dataset_name).AS(
```
paulteehan, line 427: When I thought about how I would implement this, I had exactly the same idea as you here

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -422,6 +424,33 @@ def __init__(
...
+            if (
+                self.dataset_configuration.test_row_sampler_configuration
+                and self.dataset_configuration.test_row_sampler_configuration.enabled
+                and self.dataset_configuration.test_row_sampler_configuration.test_row_sampler is not None
+            ):
+                logger.info(
+                    f"Row sampling is enabled for dataset {self.dataset_identifier.to_string()} "
+                    f"with sampler {self.dataset_configuration.test_row_sampler_configuration.test_row_sampler.type}"
+                )
+                self.cte.cte_query[1] = self.cte.cte_query[1].SAMPLE(
```
paulteehan, line 449: Ah you change FROM() into FROM.SAMPLE() right?   The array index of [1] confused me at first, I never get used to the fact we represent the top-level query as a list

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -1670,9 +1702,11 @@ def append_aggregation_metric(self, aggregation_metric_impl: AggregationMetricIm
 
     def build_sql(self) -> str:
         field_expressions: list[SqlExpression] = self.build_field_expressions()
-        select = [SELECT(field_expressions), FROM(self.dataset_name, self.dataset_prefix)]
-        if self.filter:
-            select.append(WHERE(SqlExpressionStr(self.filter)))
+        select = [
+            WITH([self.cte]),
+            SELECT(field_expressions),
+            FROM(self.cte_dataset_name),
```
paulteehan, line 1708: I got momentarily confused by `cte_dataset_name` as I thought it was linked to a soda dataset.  Now I think it's just the name of the CTE in SQL right?
like `WITH t0 as SELECT ...`, it's just referring to the `t0` part?
in that case, I petition for a more descriptive name like `cte_name`,  `cte_sql_name`, `sql_table_name` or something like that?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: yes indeed, it's a bit clumsy to access the list using hardcoded indexes like that but I haven't thought of any clever way to do this better (yet? :) )

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: yes indeed, it's just the name of  the cte in sql, I will go with `cte_name`, that's indeed more descriptive and clearly linked to the cte, thanks!

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: I will change it to cte_name, I am also not sure whether this should be public api of contract verification or "internal" with an underscore prefix. I will leave it public for now and we will see how this evolves

Niels-b: I think we may be speaking about different things here; I'm talking about `filtered_dataset`.
From the your comment, I think you're talking about `cte_dataset_name` (which is fine for me 😄 ).

m1n0: oh ok! You mean the actual name in the sql I see! Is your suggestion to make sure there are no clashes?

Niels-b: Ok scratch that, just read the rest of the comments ;). `cte_name` is fine.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
Niels-b: Facing the same issue with in-source moving of data in DWH. Parsing the AST is a bit strange as with the current AST.
Maybe we can introduce some helper methods like `find_all_FROM` or something like that?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
Niels-b: > oh ok! You mean the actual name in the sql I see! Is your suggestion to make sure there are no clashes?

Yes indeed, `filtered_dataset` might occur as the source table name. So that has to change imo.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: or we expedite the custom ast -> library switch 😅

paulteehan: oh yes please let's do that 
christmas present for ourselfs :D
https://ibis-project.org/

PR #2482, opened 2025-11-17
Dataset CTE: support all check types including custom sql

m1n0:
This PR broadens the original POC/MVP into a full blown feature:
- all check types supported, including custom sql ones
- added tests
- matured code for nicer flow and maintainability

### DIFF
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -5,7 +5,7 @@ from abc import abstractmethod
 from datetime import date, datetime, time
 from numbers import Number
 from textwrap import indent
-from typing import Any, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Optional, Tuple
 
 from soda_core.common.data_source_results import QueryResult
 from soda_core.common.dataset_identifier import DatasetIdentifier
@@ -84,6 +84,9 @@ from soda_core.common.sql_ast import (
 )
 from soda_core.common.sql_utils import apply_sampling_to_sql
 
+if TYPE_CHECKING:
+    from soda_core.common.data_source_impl import DataSourceImpl
+
 logger: logging.Logger = soda_logger
 
 
@@ -100,10 +103,8 @@ class SqlDialect:
 
     def __init__(
         self,
-        data_source_impl: "DataSourceImpl",
+        data_source_impl: DataSourceImpl,
     ):
-        from soda_core.common.data_source_impl import DataSourceImpl
-
         self.data_source_impl: DataSourceImpl = data_source_impl
 
         self._data_type_name_synonym_mappings: dict[str, str] = self._build_data_type_name_synonym_mappings(
@@ -1200,7 +1201,7 @@ class SqlDialect:
     def supports_case_sensitive_column_names(self) -> bool:
         return True
 
-    def apply_sampling_to_sql(
+    def apply_sampling(
         self,
         sql: str,
         sampler_limit: Number,
--- a/soda-core/src/soda_core/common/sql_utils.py
+++ b/soda-core/src/soda_core/common/sql_utils.py
@@ -7,6 +7,36 @@ from soda_core.common.soda_cloud_dto import SamplerType
 from sqlglot import exp
 
 
+def build_sample_clause(sampler_limit: Number, sampler_type: SamplerType) -> exp.TableSample:
+    if sampler_limit <= 0:
+        raise ValueError("sampler_limit must be positive")
+
+    size = exp.Literal.number(sampler_limit)
+    sample = exp.TableSample()
+
+    if sampler_type == SamplerType.ABSOLUTE_LIMIT:
+        sample.set("size", size)
+    else:
+        raise ValueError(f"Unsupported sample type: {sampler_type}")
+
+    return sample
+
+
+def attach_sample_to_relation(rel: exp.Expression, sampler_limit: Number, sampler_type: SamplerType) -> None:
+    """
+    Attach a TableSample clause to a relation (Table or Subquery),
+    unless it already has one.
+    """
+    if rel is None:
+        return
+
+    if rel.args.get("sample"):
+        return
+
+    if isinstance(rel, (exp.Table, exp.Subquery)):
+        rel.set("sample", build_sample_clause(sampler_limit, sampler_type))
+
+
 def apply_sampling_to_sql(
     sql: str,
     sampler_limit: Number,
@@ -20,42 +50,19 @@ def apply_sampling_to_sql(
 
     Exact rendering is dialect-specific.
     """
-    if sampler_limit <= 0:
-        raise ValueError("sampler_limit must be positive")
-
     tree = sqlglot.parse_one(sql, read=read_dialect) if read_dialect else sqlglot.parse_one(sql)
 
-    def build_sample_clause() -> exp.TableSample:
-        size = exp.Literal.number(sampler_limit)
-        sample = exp.TableSample()
-
-        if sampler_type == SamplerType.ABSOLUTE_LIMIT:
-            sample.set("size", size)
-        else:
-            raise ValueError(f"Unsupported sample type: {sampler_type}")
-
-        return sample
-
-    def attach_sample_to_relation(rel: exp.Expression) -> None:
-        """
-        Attach a TableSample clause to a relation (Table or Subquery),
-        unless it already has one.
-        """
-        if rel is None:
-            return
-
-        if rel.args.get("sample"):
-            return
-
-        if isinstance(rel, (exp.Table, exp.Subquery)):
-            rel.set("sample", build_sample_clause())
-
     # FROM sources (top-level, CTE bodies, nested subqueries)
+    # Keep track of CTEs and skip them as they are already sampled at their definition
+    ctes = {cte.alias_or_name for cte in tree.find_all(exp.CTE)}
     for from_ in tree.find_all(exp.From):
-        attach_sample_to_relation(from_.this)
+        if isinstance(from_.this, exp.Table) and from_.this.alias_or_name in ctes:
+            continue
+
+        attach_sample_to_relation(from_.this, sampler_limit, sampler_type)
 
     # JOIN targets
     for join in tree.find_all(exp.Join):
-        attach_sample_to_relation(join.this)
+        attach_sample_to_relation(join.this, sampler_limit, sampler_type)
 
     return tree.sql(dialect=write_dialect) if write_dialect else tree.sql()
--- a/soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
@@ -90,7 +90,6 @@ class InvalidCheckImpl(MissingAndValidityCheckImpl):
             # this is used in the check extension to extract failed keys and rows
             self.ref_query = InvalidReferenceCountQuery(
                 cte=contract_impl.cte,
-                cte_name=contract_impl.cte_name,
                 sampler_type=contract_impl.sampler_type,
                 sampler_limit=contract_impl.sampler_limit,
                 metric_impl=self.invalid_count_metric_impl,
@@ -196,7 +195,6 @@ class InvalidReferenceCountQuery(Query):
     def __init__(
         self,
         cte: CTE,
-        cte_name: str,
         sampler_type: Optional[str],
         sampler_limit: Optional[Number],
         metric_impl: InvalidReferenceCountMetricImpl,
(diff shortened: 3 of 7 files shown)


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -96,7 +98,14 @@ class SqlDialect:
 
     SODA_DATA_TYPE_SYNONYMS: tuple[tuple[SodaDataTypeName, ...]] = ()
 
-    def __init__(self):
+    def __init__(
+        self,
+        data_source_impl: "DataSourceImpl",
+    ):
+        from soda_core.common.data_source_impl import DataSourceImpl
```
mivds: Can we make this a top level import? I think that would be preferable for type hinting in code editor. Guessing it's here because of circular dependencies. Maybe below would work?

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  from soda_core.common.data_source_impl import DataSourceImpl
```

soda-core/src/soda_core/common/sql_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        if isinstance(rel, (exp.Table, exp.Subquery)):
+            rel.set("sample", build_sample_clause())
+
+    # FROM sources (top-level, CTE bodies, nested subqueries)
+    for from_ in tree.find_all(exp.From):
+        attach_sample_to_relation(from_.this)
+
+    # JOIN targets
+    for join in tree.find_all(exp.Join):
+        attach_sample_to_relation(join.this)
```
mivds: Are we sure we want to add sampling to joins? Generally I would expect a join to pick records related to a source table using some condition (1-to-1 or 1-to-N mapping). If we're already sampling the source table, it doesn't seem necessary to also sample the joined table

Also, I'm not even sure what SQL a join with sampling would generate to be honest

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -191,62 +195,64 @@ class DatasetAlias(Enum):
 class InvalidReferenceCountQuery(Query):
     def __init__(
         self,
+        cte: CTE,
+        cte_name: str,
```
mivds: Shouldn't the CTE name be a property of the CTE? There seems to be `cte.alias` which is not-nullable

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -191,62 +195,64 @@ class DatasetAlias(Enum):
...
+            [
+                SELECT(STAR()),
+                FROM(referenced_dataset_name).IN(referenced_dataset_prefix),
+            ]
+        )
 
-            if dataset_filter_expr and check_filter_expr:
-                combined_filter_expr = AND([dataset_filter_expr, check_filter_expr])
+        if self.sampler_type and self.sampler_limit:
+            cte.cte_query[1] = cte.cte_query[1].SAMPLE(self.sampler_type, self.sampler_limit)
```
mivds, line 244: Sanity check: does it make sense to sample the referenced table? I'm not entirely sure about the use case for this check, but I imagine the referenced table would contain a list of allowed values (which could be many). If that's correct sampling the list of allowed values seems like it would lead to incorrect behavior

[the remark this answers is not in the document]
mivds: I kind of agree with ChatGPT here. I don't like the way our AST is constructed currently with just having a list. I would prefer something like `Select(Star()).from(referenced_datset)`, i.e. a `Select` expression that we operate on.

Anyway, that's beyond the scope of this PR. If we want to be less brittle using current API we could do:
```python
from = FROM(referenced_dataset_name).IN(referenced_dataset_prefix)
if self.sampler_type and self.sampler_limit:
    from.SAMPLE(self.sampler_type, self.sampler_limit)
cte = CTE(self.referenced_cte_name).AS([SELECT(STAR()), from])
```

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -484,6 +496,18 @@ def __init__(
...
+    @property
+    def is_test_verification_on_agent(self) -> bool:
+        return self.soda_config.is_running_on_agent and not self.publish_results
+
+    @property
+    def is_sampling_enabled(self) -> bool:
+        return self.sampler_type is not None and self.sampler_limit is not None
+
+    @property
+    def do_apply_sampling(self) -> bool:
```
mivds: I don't like the name for this property; the `do_` prefix suggests it has some kind of side effect. It doesn't. It's just a getter

Maybe `should_apply_sampling` is a better name? Although I'm not sure what the difference in usage is with the `is_sampling_enabled` property above

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Nice prevention of the circular import ;)

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -1191,6 +1200,20 @@ def get_sql_data_type_class(self) -> type:
     def supports_case_sensitive_column_names(self) -> bool:
         return True
 
+    def apply_sampling_to_sql(
+        self,
+        sql: str,
+        sampler_limit: Number,
+        sampler_type: SamplerType,
+    ) -> str:
+        return apply_sampling_to_sql(
```
Niels-b, line 1210: Naming here is confusing. It's the same name as the function we're calling it from. I thought it was recursive at first.

Niels-b: Maybe rename the function defined in `SqlDialect` to `apply_sampling`?

soda-core/src/soda_core/common/sql_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    including inside CTEs and subqueries.
+
+    Exact rendering is dialect-specific.
+    """
+    if sampler_limit <= 0:
+        raise ValueError("sampler_limit must be positive")
+
+    tree = sqlglot.parse_one(sql, read=read_dialect) if read_dialect else sqlglot.parse_one(sql)
+
+    def build_sample_clause() -> exp.TableSample:
```
Niels-b: Personally not a fan of having function definitions in another function. Is there a specific reason for that (I don't see one directly)?

soda-core/src/soda_core/common/sql_utils.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
+
+from numbers import Number
+
+import sqlglot
+from soda_core.common.soda_cloud_dto import SamplerType
+from sqlglot import exp
```
Niels-b, line 7: Seems to work well for this. Code is not that complicated just reading it (figuring it out may be a bit more difficult 😅 )

soda-tests/tests/components/test_sql_utils.py
```diff
@@ -0,0 +1,217 @@
...
+    assert isinstance(size1, exp.Literal)
+    assert size1.this == "10"
+
+    # cte as FROM source in outer query is represented as a Table
+    cte_table = _get_first_table(tree, "cte")
+    sample_cte = cte_table.args.get("sample")
+    assert isinstance(sample_cte, exp.TableSample)
+    size_cte = sample_cte.args.get("size")
+    assert isinstance(size_cte, exp.Literal)
+    assert size_cte.this == "10"
```
mivds: Not sure sampling makes sense here to be honest. Intuitively I would only apply sampling to the CTE source table (`t1` in this case)

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -191,62 +195,64 @@ class DatasetAlias(Enum):
...
         data_source_impl: Optional[DataSourceImpl],
     ):
         super().__init__(data_source_impl=data_source_impl, metrics=[metric_impl])
         self.metric_impl = metric_impl
-        self.dataset_filter = dataset_filter
         self.check_filter = check_filter
 
         self.referencing_alias: str = DatasetAlias.CONTRACT.value
         self.referenced_alias: str = DatasetAlias.REFERENCE.value
+        self.referenced_cte_name: str = "filtered_referenced_dataset"
```
Niels-b: Careful with the name here `filtered_referenced_dataset` may occur during a POC (for example) as an actual dataset name. This will break this CTE.

soda-core/src/soda_core/common/sql_dialect.py
m1n0: yes indeed it was because of the circular imports, I forgot about this trick, added, thanks!

soda-core/src/soda_core/common/sql_utils.py
m1n0: yes it is very aggressive, but this feature (at least in this form) is 100% geared for cost savings, correctness of results is not the goal here at all. We discussed the same concern in the team and chose to do simple/naive application of sampling everywhere without too much effort and ship the most cost-saving solution and adjust if the meaningless results become an issue

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
Niels-b: I second this.

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
m1n0: yes indeed, "join thinning" will happen here (i.e. if you get 10% of base table and then 10% of the joined one statistically you will get 1% of matching and 9% of non-matching rows from the original amount) - but we are more concerned about cost savings than good behaviour for now.

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
m1n0: oh I missed that one, yes indeed I can reuse it from within the CTE object itself!

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
m1n0: good point, I prefixed all names with `_soda_` to make them unique enough, wdyt?

soda-tests/tests/components/test_sql_utils.py
Niels-b: I think the case for this test is that the sampling gets applied to all FROM clauses, even the ones in a CTE. Which is specific to the use case for this code (as far as I know. @m1n0 correct me if I'm wrong). Doing this in normal queries is indeed strange 😅 .

However, even then I agree with @mivds here. If there is a way we can determine that the table has already been sampled in the CTE, we don't have to sample that CTE output anymore in the "main" query (or follow-up CTEs if that may occur). So we only sample each source-table exactly once 🤔 .

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: indeed, `should_apply_sampling` it is! The difference between `should_apply_sampling` and  `is_sampling_enabled` is:
- `is_sampling_enabled` - whether we have config for sampling
- `should_apply_sampling` - whether we have config AND we are running in the correct context (Test Contract in UI, i.e. is agent and publish is false)

soda-core/src/soda_core/common/sql_utils.py
m1n0: indeed, good enough, although the abstraction does not fully cover everything, more info on that on DE sync!

soda-core/src/soda_core/common/sql_utils.py
m1n0: yeah, I didn't want to make them fully separate and testabe, and write unit tests for them so I kept them in for a more compact interface. But they ended up smaller, so valid point - I refactored them out as separate functions, although no extra tests :)

soda-tests/tests/components/test_sql_utils.py
m1n0: Yes indeed! I changed the logic to skip FROM if it's a CTE name since those are already sampled, great catch!

soda-core/src/soda_core/common/sql_dialect.py
m1n0: ah ok, valid point, renamed as suggested

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
Niels-b: Perfect 👌

PR #2484, opened 2025-11-19
Add CREATE TABLE AS support for Postgres

Niels-b:
Also includes a change to verify if a table already exists or not

### DIFF
--- a/soda-tests/tests/components/test_sql_generation.py
+++ b/soda-tests/tests/components/test_sql_generation.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from unittest import mock
 
 from soda_core.common.sql_dialect import *
+from soda_postgres.common.data_sources.postgres_data_source import PostgresSqlDialect
 
 mock_data_source_impl = mock.MagicMock()
 
@@ -250,7 +251,9 @@ def test_sql_ast_alter_table_drop_column():
 
 
 def test_sql_ast_create_table_as_select():
-    sql_dialect: SqlDialect = SqlDialect(mock_data_source_impl)
+    sql_dialect: SqlDialect = PostgresSqlDialect(
+        mock_data_source_impl
+    )  # Temporarily use the postgres method specifically. It's not implemented in the base class yet.
 
     my_create_table_as_select_statement = sql_dialect.build_create_table_as_select_sql(
         CREATE_TABLE_AS_SELECT(


soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -15,7 +15,10 @@
     SchemaDataSourceNamespace,
 )
 from soda_core.common.sql_dialect import SqlDialect
-from soda_core.common.statements.metadata_tables_query import MetadataTablesQuery
+from soda_core.common.statements.metadata_tables_query import (
+    FullyQualifiedTableName,
```
m1n0, line 19: if `FullyQualifiedTableName` is used outside of that query should we maybe extract it into a separate file?

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -424,6 +425,15 @@ def _quote_column_for_create_table(self, column_name: str) -> str:
     def _is_not_null_ddl_supported(self) -> bool:
         return True
 
+    def build_create_table_as_select_sql(
```
m1n0, line 438: should this perhaps go into postgres dialect and raise NotImplemented in sql dialect?

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -214,3 +217,21 @@ def test_schema_exists(self, prefixes: list[str]) -> bool:
                 schema_exists = True
                 break
         return schema_exists
+
+    def verify_if_table_exists(self, prefixes: list[str], table_name: str) -> bool:
```
m1n0, line 221: would it be possible to use this in tests? Don't we check if table exists there as well?

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: I don't think so; we're using it to type the return value of the `metadata_tables_query.execute()` only. It never gets exposed outside of the `verify_if_table_exists` function. 
That's still "contained enough" for me. But no strong opinion; if you want me to change it, I can. We can also do this later

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Yes! For now that will be the way to go.
While usage is limited to only Postgres for now (we use it in extensions), the idea is to start implementing this function for other data sources ASAP. 
But I fully agree that it is the better approach and will result in a better intermediate state while working on other data sources. 
Good catch 🙌 !

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: Yes, should definitely be possible! However, that's out of scope for this PR imo.

PR #2486, opened 2025-11-20
Add support for BigQuery `storage_project_id`

mivds:
Adds support for the `storage_project_id` parameter to the BigQuery connection.

Note this parameter is currently not used, because the project ID where the dataset is located is already specified as part of the DQN in the data contract. It's more extension for future when we do discovery using the `soda-core` codebase.


mivds:
Replaced by #2490 (the branch name was incorrect, linking to the wrong linear issue)

PR #2500, opened 2025-11-25
Dremio contracts support

paulteehan:
To support some quirks with Dremio

Dremio development happening in extensions: 
https://github.com/sodadata/soda-extensions/pull/182

### DIFF
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -144,8 +144,7 @@ class DataSourceConnection(ABC):
         updates = cursor.execute(sql)
         self.commit()
         return updates
- 
-    
+
     def execute_query_one_by_one(
         self, sql: str, row_callback: Callable[[tuple, tuple[tuple]], None], log_query: bool = True
     ) -> tuple[tuple]:
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -9,11 +9,11 @@ from typing import TYPE_CHECKING, Any, Optional, Tuple
 
 from soda_core.common.data_source_results import QueryResult
 from soda_core.common.dataset_identifier import DatasetIdentifier
-from soda_core.common.logging_constants import soda_logger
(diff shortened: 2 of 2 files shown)


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -135,12 +135,16 @@ def execute_update(self, sql: str, log_query: bool = True) -> UpdateResult:
...
-            self.commit()
-            return updates
+            return self._cursor_execute_update_and_commit(cursor, sql)
+
         finally:
             cursor.close()
 
+    def _cursor_execute_update_and_commit(self, cursor: Any, sql: str):
+        updates = cursor.execute(sql)
+        self.commit()
```
m1n0, line 145: no autocommit option for dremio connections?

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -79,9 +83,16 @@ def __init__(
         self.execution_timestamp: datetime = datetime.now(timezone.utc)
         self.data_timestamp: datetime = self._get_data_timestamp(data_timestamp, self.execution_timestamp)
 
+        # Some dialects (Dremio) don't use ISO format, we need to override
```
m1n0, line 86: is this safe to override for all datasources?

soda-core/src/soda_core/common/data_source_connection.py
paulteehan: There is, that's the generic implementation, the Dremio implementation doesn't call `commit()`

soda-core/src/soda_core/contracts/impl/contract_yaml.py
paulteehan: I think so, the default behavior is to just call the original function again, and Dremio has the only override, so for everyone aside from Dremio the behavior should be unchanged.  And tests are passing for all data sources

paulteehan: I'm confident enough that I'm going to merge it

PR #2501, opened 2025-11-25
Add full `CREATE_TABLE_AS` AST support for all data sources

Niels-b:
Support the `CREATE_TABLE_AS` AST for all data sources:
- Centralize the function instead of putting it in all data sources individually
- Updated ast test to reflect the new changes
- Added `INTO` in the AST because SqlServer needs that

Also fixed for Oracle, but that is in the other repository


Niels-b:
> LGTM, but perhaps we should add at least some basic test coverage for generating the sql?

That was already added in a previous commit: https://github.com/sodadata/soda-core/blob/1cc621a69ad90da2a837753ef4bef7a4d5f14e34/soda-tests/tests/components/test_sql_generation.py#L253

PR #2502, opened 2025-11-26
Snowflake: WH switch logging tweaks

m1n0:


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -627,6 +627,11 @@ def verify(self) -> ContractVerificationResult:
                     f"Switching warehouse from '{self.datasource_warehouse}' to '{self.compute_warehouse}' for Contract verification of dataset '{self.dataset_identifier.to_string()}'"
                 )
             self.data_source_impl.switch_warehouse(self.compute_warehouse)
+        else:
+            if self.data_source_impl:
+                logger.info(
+                    f"Using warehouse '{self.datasource_warehouse}' for Contract verification of dataset '{self.dataset_identifier.to_string()}'"
```
mivds: Won't this log `None` for practically all data source types other than snowflake?

m1n0: oh wow what an overlook. Thanks for catching that. I refactored this a bit so that all of this is in snowflake and general implementation just has a method call

mivds (approved):

PR #2506, opened 2025-12-03
Add hash to dataset identifier. Lock databricks-sql-connector to `<4.2.2`

Niels-b:
Needed for work in soda-extensions.
Adds a `to_hash()` method to the `DatasetIdentifier`.

### DIFF
--- a/soda-databricks/setup.py
+++ b/soda-databricks/setup.py
@@ -8,7 +8,7 @@ description = "Soda Databricks V4"
 
 requires = [
     f"soda-core=={package_version}",
-    "databricks-sql-connector<4.2.2",  # Temporary fix to avoid issues with new version (4.2.2)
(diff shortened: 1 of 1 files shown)


soda-databricks/setup.py
```diff
@@ -8,7 +8,7 @@
 
 requires = [
     f"soda-core=={package_version}",
-    "databricks-sql-connector",
+    "databricks-sql-connector<4.2.2",  # Temporary fix to avoid issues with new version (4.2.2)
```
mivds: Might want to go with `!=4.2.2` here. I suspect the package will be yanked once they discover it broke things. Version `4.2.0` of this package was also yanked because of issues.

Alternatively we need a reminder to change the restriction whenever there is a new release

PR #2509, opened 2025-12-08
Add more support for limits

Niels-b:
- Add support for LIMIT in SqlServer
  - Retained backwards compatibility with the `OFFSET` limit of the queries (this requires different syntax)
  - The "normal LIMIT" is done by `SELECT TOP N...` in SqlServer
- Add limit in `execute_one_by_one` function so we can stop processing after a certain number of rows.

### DIFF
--- a/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
+++ b/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
@@ -79,19 +79,25 @@ class SqlServerSqlDialect(SqlDialect):
 
         limit_line = self._build_limit_line(select_elements)
         if limit_line:
-            if not limit_line.startswith("__SODA_LIMIT_TOP__"):
-                statement_lines.append(limit_line)
-            else:  # We should add the TOP statement to the SELECT statement.
-                # Get the last SELECT line and add the TOP statement to it.
-                last_select_line: str = [line for line in statement_lines if line.startswith("SELECT")][-1]
-                last_select_line_with_top = last_select_line.replace(
-                    "SELECT ", f"SELECT TOP {limit_line.split('__SODA_LIMIT_TOP__')[1]} "
-                )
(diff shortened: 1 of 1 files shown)


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -161,8 +165,11 @@ def execute_query_one_by_one(
 
             description: tuple[tuple] = cursor.description
 
+            rows_processed: int = 0
+
             row = cursor.fetchone()
-            while row:
+            while row and (row_limit is None or rows_processed < row_limit):
```
paulteehan, line 171: Oh great, we really need this!!

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source.py
```diff
@@ -79,9 +79,37 @@ def build_select_sql(self, select_elements: list, add_semicolon: bool = True) ->
 
         limit_line = self._build_limit_line(select_elements)
         if limit_line:
-            statement_lines.append(limit_line)
+            if not limit_line.startswith("__SODA_LIMIT_TOP__"):
```
paulteehan: Oh geez this turned out to be tricky huh. The select_elements list architecture strikes again 

I'm not a huge fan of passing string values back and parsing them to handle control flow, it just makes it hard to me to follow. 

Could you do something like this maybe?

```
def _requires_select_top(self, select_elements): 
   return self._limit_statement_present(select_elements) and not self.uses_offset(select_elements)
   
def _build_select_sql_lines(select_elements):
   (... as before)
   if self._requires_select_top(select_elements): 
      # logic to insert 'TOP' where needed 
 ```

Niels-b: Fantastic suggestion! Thanks a lot for this Paul.
That is indeed a much better flow, more readable and easier to understand.

PR #2510, opened 2025-12-09
Variables: add support in lists and dicts values as well

m1n0:
context: https://linear.app/sodadata/issue/DTL-1397/variables-are-not-passed-correctly-when-used-in-attributes

use case why we have the "only string" limitation - https://github.com/sodadata/soda-core/pull/2437/files#diff-170af9abbde7e05f89d12fa1d813f2f70e754ccecc38011f0e75affad7f1d8d8

I ran extensions tests on postgres before opening the PR

### DIFF
--- a/soda-core/src/soda_core/common/yaml.py
+++ b/soda-core/src/soda_core/common/yaml.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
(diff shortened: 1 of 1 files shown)


Niels-b (approved):
LGTM!

PR #2512, opened 2025-12-10
Fail queries gracefully

m1n0:
This catches contract verification queries and continue the verification to still run checks that can continue. Caveat: since multiple checks types are combined into one query, this query can fail and cause a non-intuitive number of checks to be in non-evaluated state since the aggregate combined query failed. 

This changes the previous approach of "fail the whole contract immediately in case of a query issue" to "fail gracefully, execute everything that is possible, including sending data to Cloud and running extensions"

extensions tests passed locally


m1n0:
Closing, this is too narrow and will need to be addressed with more nuance

PR #2513, opened 2025-12-10
Variables: retain value numeric type when possible

m1n0:
Small followup on previous PR that uncovered a limitation of variables - numeric type was not retained after resolving. This addresses that limitation by first resolving with no type casting if the whole value being resolved is the variable itself. If that is not the case we do what we did originally - resolve as a string so that both string and numeric variables can be injected into a string.
This PR also adds a new column in the logs verification overview - the threshold.

### DIFF
--- a/soda-core/src/soda_core/contracts/contract_verification.py
+++ b/soda-core/src/soda_core/contracts/contract_verification.py
@@ -255,19 +255,19 @@ class Threshold:
     def __str__(self) -> str:
         parts = []
         if self.must_be is not None:
-            parts.append(f"= {self.must_be}")
+            parts.append(f"must be: {self.must_be}")
         if self.must_not_be is not None:
-            parts.append(f"!= {self.must_not_be}")
+            parts.append(f"must not be: {self.must_not_be}")
         if self.must_be_greater_than is not None:
-            parts.append(f"> {self.must_be_greater_than}")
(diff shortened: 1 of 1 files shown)


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -249,6 +249,25 @@ class Threshold:
...
+            parts.append(f"must not be: {self.must_not_be}")
+        if self.must_be_greater_than is not None:
+            parts.append(f"must be greater than: {self.must_be_greater_than}")
+        if self.must_be_greater_than_or_equal is not None:
+            parts.append(f"must be greater than or equal: {self.must_be_greater_than_or_equal}")
+        if self.must_be_less_than is not None:
+            parts.append(f"must be less than: {self.must_be_less_than}")
+        if self.must_be_less_than_or_equal is not None:
+            parts.append(f"must be less than or equal: {self.must_be_less_than_or_equal}")
+        parts_str = "\n".join(parts)
```
Niels-b, line 269: Are there specific rules that you cannot config a threshold with "silly configs"?
For example: "must be x" and "must not be x" at the same time?

m1n0: I am not sure if we block those from the top of my head, but if we do then this is not the place - this is just rendering already existing data. Blocking those would happen in the parser phase

PR #2514, opened 2025-12-11
Fix min/max/avg length aggregate checks

paulteehan:
Explanation from this thread: https://sodadata.slack.com/archives/CAXRR8SS3/p1765470255686749

> Aggregate functions seem to be pretty broken in the product?  Are these known issues?
> Select any column in any data source
> Add an aggregate check and select 'min Length', 'max Length', or 'avg Length'
> The scan will fail with something like Aggregate function 'maxLength' is not supported on 'postgres'
> Benjamin Pirotte surfaced this in Dremio testing, it actually applies to all data sources.  It looks like min/max/avg length were never implemented in soda-core and have always been broken
> Also generated contracts will write 'avgLength', 'minLength', 'maxLength' but core expects 'avg_length', 'min_length', 'max_length'.
> I'm adding support for either construction in core

### DIFF
--- a/soda-tests/tests/features/test_aggregate_check.py
+++ b/soda-tests/tests/features/test_aggregate_check.py
@@ -1,5 +1,4 @@
 import pytest
-
 from helpers.data_source_test_helper import DataSourceTestHelper
 from helpers.mock_soda_cloud import MockResponse
 from helpers.test_functions import get_diagnostic_value


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -978,7 +990,35 @@ def _build_offset_sql(self, offset_element: OFFSET) -> str:
...
+            "max",
+            "min",
+            "max_length",
+            "min_length",
+            "sum",
+            "avglength",
+            "maxlength",
+            "minlength",
+        ]
+
```
paulteehan, line 1005: These two methods are coupled, supports_function and get_function_expression.  There should probably be more structure here but I just put them side-by-side for now.

Initial implementation expected "avg_length" but generated contracts in the platform will write "avgLength" so I've added support for either form

Niels-b: Thanks for the explanation, makes it clear.

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -978,7 +990,35 @@ def _build_offset_sql(self, offset_element: OFFSET) -> str:
         return f"OFFSET {offset_element.offset}"
 
     def supports_function(self, function: str) -> bool:
```
Niels-b, line 992: Do we expect certain databases to not support specific functions? (Not for this PR, just a question, I see no overrides for this function atm)

soda-tests/tests/features/test_aggregate_check.py
```diff
@@ -56,6 +58,109 @@ def test_aggregate_function_avg(data_source_test_helper: DataSourceTestHelper):
...
+                checks:
+                  - aggregate:
+                      function: avg_length
+                      threshold:
+                        must_be_less_than: 2.3
+        """,
+    )
+    check_result: CheckResult = contract_verification_result.check_results[0]
+    result_value = get_diagnostic_value(check_result, "avg_length")
+    assert (math.isclose(result_value, 2.2)) or (result_value == 2)  # some DBs will return ints
```
Niels-b: You can also use `pytest.approx` for this:
For example (copied straight from some other code: `assert pytest.approx(query_result.rows[0][6], 0.000001) == 6`

soda-core/src/soda_core/common/sql_dialect.py
paulteehan: Honestly don't know what the original intent was, I think maybe the intent was to leave openings for future functions?

soda-tests/tests/features/test_aggregate_check.py
paulteehan: Oh that's handy thank you!

Issue #2295, opened 2025-06-03
DuckDB 1.0.0 vulnerability


sammyzord:
### Description
Hi all, 

DuckDB 1.0.0 is subject to a vulnerability, namely https://github.com/advisories/GHSA-w2gf-jxc9-pf2q. 

The problem seems to come from the `soda-cora-duckdb` package, specifically the following line:

```
requires = [f"soda-core=={package_version}", "duckdb<1.1.0"]
```

This restricts it to duckdb 1.0.0 only, apparently.

The patched version is 1.1.0, I saw in some previous discussions that allowing for 1.2.0+ would require substantial rework, but is it the same for version==1.1.0?

Thanks in advance!

tools-soda:
CLOUD-9190

AntoineGiraud:
It'd be neat to go up to date on DuckDB 😍 
`v.1.4.3`

jckinjo:
Do we have any updates for this issue?

We cannot upgrade python to 3.13+ unless we upgrade duckdb to 1.1.0+

PR #2527, opened 2025-12-24
chore: allow new versions of duckdb

elefeint:
The duckdb version was pinned to below 1.1.0 because the dataframe behavior changed then. This PR restores the old behavior of scanning all registered dataframes, but because this option was only added in DuckDB 1.1.0, this version now becomes the minimum.

This PR also standardizes the duckdb configuration used in the in-memory and path-based connections and, because `:default:` in-memory singleton cannot accept configuration, switches from `:default:` to `:memory:`. The in-memory duckdb instances are so lightweight though that it won't make any measurable difference.

I've added `<2.0` as the dependency ceiling, but DuckDB treats its twice-a-yearly minor releases as major, so if you'd like to be conservative, the ceiling should be `<1.5`, which [will be released in February](https://duckdb.org/release_calendar).

`custom_user_agent` is a DuckDB property. It was added in DuckDB 1.0, so no additional constraints are needed.

### DIFF
--- a/soda/duckdb/setup.py
+++ b/soda/duckdb/setup.py
@@ -6,7 +6,7 @@ package_name = "soda-core-duckdb"
 package_version = "3.5.6"
 description = "Soda Core Duckdb Package"
 
-requires = [f"soda-core=={package_version}", "duckdb<1.1.0"]
+requires = [f"soda-core=={package_version}", "duckdb>=1.1.0, <2.0"]
 
 setup(
     name=package_name,
--- a/soda/duckdb/soda/data_sources/duckdb_data_source.py
+++ b/soda/duckdb/soda/data_sources/duckdb_data_source.py
@@ -12,6 +12,7 @@ import logging
 from pathlib import Path
 from typing import List, Optional
 
+from soda.__version__ import SODA_CORE_VERSION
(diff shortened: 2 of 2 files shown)


bartoszkosiorek:
What else needs to be done to merge this PR?
It will resolve High Severity security issue: https://github.com/sodadata/soda-core/issues/2295 (https://github.com/advisories/GHSA-w2gf-jxc9-pf2q)

elefeint:
It needs maintainer approval. @m1n0 perhaps?

renewfresho:
or @Niels-b ? though it looks like all the effort is understandably on v4 not v3 so this might not ever be prioritized. also it seems v4 is the license that doesn't allow third parties to host the software and sell it as a service for profit unlike v3 which is apache. so I very much doubt this PR will ever be merged in as it appears against sodadata's interests to maintain it. probably best to move on to v4

Niels-b:
Hi @renewfresho 

Thank you for the ping. We've looked at the PR and this looks good. Thank you for addressing this.

We will merge this so it's on the `v3` branch. But indeed, we recommend moving to v4, there we have DuckDB >= 1.2.0

Niels-b (approved):

PR #2657, opened 2026-04-10
fix(athena): quote schema/database identifiers in table prefix to handle hyphenated names

### LINKED ISSUE #2483: Missing quotation marks around the schema / database in athena query strings
Database names in AWS Athena can contain hyphen (-) characters that will be misinterpreted by the sql query engine without quotation marks. 

Currently, the query formatting for the athena data source does not put those quotation marks, I think. maybe, an overwrite of the table_prefix or the _create_table_prefix function in the athena data source might fix the issue.

hirenkumar-n-dholariya:
## Problem

AWS Athena database and schema names can contain hyphen (-) characters.
The current code returns self.schema unquoted in _create_table_prefix(),
causing SQL query failures when hyphens are present:

  SELECT * FROM my-schema.my_table   ← SQL engine misreads the hyphen

## Root Cause

In data_source.py, the _create_table_prefix() method returns:

  return self.schema

This bare, unquoted value is then used directly in qualified_table_name() to build SQL strings, with no protection against special characters.

## Fix
Wrap the schema in double quotes:

  if self.schema:
      return f'"{self.schema}"'
  return None

This produces valid SQL:  SELECT * FROM "my-schema".my_table

## Impact
- Fixes query failures for Athena databases/schemas with hyphens
- No breaking change for normal names without special characters
- Double quoting is standard SQL and supported by Athena

## References
- Fixes #2483

### DIFF
--- a/soda/core/soda/execution/data_source.py
+++ b/soda/core/soda/execution/data_source.py
@@ -1119,7 +1119,9 @@ class DataSource:
             return self.database
             return f'"{self.database}"."{self.schema}"'
         """
-        return self.schema
+        if self.schema:
+            return f'"{self.schema}"'
+        return None
 
     def update_schema(self, schema_name):
         self.schema = schema_name


hirenkumar-n-dholariya:
Hi @Niels-b @bmarinovic @tomassatka 👋

I've submitted this fix for issue #2483 where AWS Athena database/schema names containing hyphens (-) cause SQL query failures because they are not quoted in the table prefix.

The fix wraps the schema in double quotes in _create_table_prefix() so the generated SQL becomes "my-schema".my_table instead of my-schema.my_table.

All checks are passing!

Could someone please review when you get a chance? Happy to make any changes based on feedback. Thank you!

hirenkumar-n-dholariya:
Hi @bmarinovic @Niels-b

I hope you're doing well! I wanted to gently follow up on this PR which addresses the Athena query failure reported in issue #2483.

The fix is minimal - just wrapping the schema in double quotes in _create_table_prefix() so hyphenated database/schema names like "my-schema" are handled correctly in SQL queries.

I know your time is valuable and there are many things competing for your attention. Whenever this bubbles up in your queue, even a brief comment on whether the approach looks right with PR approval would mean a lot to me as a first-time contributor on this repo.

Thank you so much for your time!

hirenkumar-n-dholariya:
Hi @bmarinovic @tomassatka @Niels-b

Hope you are doing well!
Could you please review PR https://github.com/sodadata/soda-core/pull/2657 and give your approval/feedback to proceed.

PR #2659, opened 2026-04-13
fix: cast boolean threshold values to int before sending to Soda Cloud

Lepo00:
## Summary

- When a metric check uses a SQL query returning a boolean (e.g. a comparison expression on Databricks), the raw Python `bool` flows through to `diagnostics.value` and gets serialized as JSON `true`/`false` instead of `1`/`0`
- The backend expects a numeric value and rejects the payload with **400 "Could not parse request from json body"**
- Cast `bool` to `int` in `_build_diagnostics_json_dict` so the serialized value is always numeric

Fixes SAS-11624

## Test plan

- [x] Added parametrized unit test covering `True`, `False`, numeric, zero, and `None` threshold values
- [x] All 6 parametrized cases pass
- [ ] Verify on a Databricks data source with a boolean-returning metric query


Niels-b (approved):
LGTM! Thanks for taking care of this one 🙌 !

PR #2660, opened 2026-04-14
feat: add tokenUsage and recordsWritten to scan results (DES-279, DES-283)

AidaPaul:
## Summary

- Add `ScanTokenUsage` dataclass and `token_usage` field on `ContractVerificationResult` for scan-level AI token tracking
- Serialize `tokenUsage` list in the scan results JSON payload (`promptTokens`, `completionTokens`, `totalTokens`, `model`, `operation`)
- Add `records_written` to `PostProcessingStage` and send it in the `sodaCorePostProcessingUpdate` command
- Add unit tests for token usage serialization

Part of the [Usage Tracking](https://linear.app/sodadata/project/usage-tracking-d8a79499f9af) project.

**Companion PRs:**
- BE: sodadata/soda#12020
- Agent: sodadata/soda-extensions (usage-tracking branch)

## Test plan

- [x] `test_build_token_usage_dicts_serialization` — verifies JSON field mapping
- [x] `test_build_token_usage_dicts_empty_when_no_usage` — verifies empty/null handling


.claude/settings.local.json
Lepo00, line 1: I think you shouldn't commit the local claude settings file

Lepo00 (approved):
lgtm

.claude/settings.local.json
AidaPaul: Hah yeah, i relied on trusting gitignore to be in place, sniped :D

AidaPaul:
Added: `execute_update` now returns `cursor.rowcount` (int) instead of the unused `UpdateResult` wrapper.

`cursor.rowcount` is part of DB-API 2.0 — every driver already sets it after DML. This is zero-cost (no extra queries). Updated across all connection implementations:
- Base `DataSourceConnection` — captures `rowcount` before commit
- Trino — captures `rowcount` before `fetchall()` drains results
- Postgres, Athena — updated return type annotations
- Snapshot connection — replay mode returns `0`

This enables sodadata/soda-extensions#345 to track rows written by in-source-transfer paths (CTAS / INSERT...SELECT) without performance penalty.

soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
```diff
@@ -13,7 +13,6 @@
 )
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_impl import DataSourceImpl, FullyQualifiedTableName
-from soda_core.common.data_source_results import UpdateResult
```
paulteehan, line 16: From what I can tell UpdateResult was never used, and is now completely dead code.   I would remove it in that case to prevent future confusion

paulteehan (approved):
Recommend removing now un-used UpdateResult, otherwise okay to merge!

PR #2661, opened 2026-04-14
fix: Pre-flight probe for Databricks connection (SCS-884)

paulteehan:
## Human-authored summary

Databricks connections can take a long time to establish while the cluster warms up.  For that reason default timeout is 15 minutes.   A few months back we greatly shortened timeout because customers were waiting 15 minutes to discover their connection params were wrong (e.g. incorrect URL).  But now we're timing out before clusters come online. 

This PR adds a fast probe to see if the host is reachable and fail quickly if not, before attempting to connect with the full default timeout.   

See also: https://github.com/sodadata/soda-library/pull/726 

## Summary

- Adds a TCP+TLS pre-flight probe before `sql.connect()` for Databricks connections in soda-core v4. Fast-fails connect-phase errors (DNS, TCP refused, TLS handshake) in ~10s instead of retrying with exponential backoff for up to ~15 min.
- v4 never had PR #563's retry caps, so this is a proactive improvement — same category of hang was possible on misconfigured hostnames, just not yet reported.
- Companion to sodadata/soda-library#726 which fixes the active SCS-884 regression.

## Context

- **SCS-884:** Nubank — Databricks connection failures after agent upgrade. Root cause (in soda-library) was overly aggressive retry caps. This PR brings the same probe defense to soda-core v4 for consistency.
- **OBSL-445:** Original wrong-URL 15-min hang issue. Confirmed by Michaël as connect-phase (incorrect URL), which the probe addresses.

## Test plan

- [x] 6 new unit tests: probe success, DNS failure, TCP timeout, TCP refused, network unreachable, TLS handshake error
- [x] 15/15 full soda-databricks test suite passing (including 4 real-Databricks integration tests — probe is transparent for healthy hosts)
- [ ] End-to-end: wrong hostname confirms fast-fail (~10s instead of ~15 min)

### DIFF
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
@@ -33,9 +33,10 @@ def _probe_databricks_reachability(host: str, port: int = 443, timeout: float = 
     try:
         with socket.create_connection((host, port), timeout=timeout) as sock:
             ctx = ssl.create_default_context()
+            ctx.minimum_version = ssl.TLSVersion.TLSv1_2
             with ctx.wrap_socket(sock, server_hostname=host):
                 return
-    except (socket.gaierror, socket.timeout, ConnectionError, ssl.SSLError, OSError) as e:
+    except OSError as e:
         raise ConnectionError(f"Databricks host {host}:{port} is not reachable: {e}") from e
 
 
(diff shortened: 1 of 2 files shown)


paulteehan:
Closing in favor of the simpler approach: just reverting retry overrides in soda-library (sodadata/soda-library#726). v4 never had the retry caps, so no change needed here.

PR #2664, opened 2026-04-14
Add adapter conformance test suite

paulteehan:
## Summary

AI-written.

Adapter conformance test suite — **65 integration tests** across 3 new files that catch field bugs at the adapter contract layer. Plus two adapter source fixes for real bugs the suite found.

- `test_conformance_identifiers.py` (24 tests) — reserved SQL words, hyphens, mixed-case as column names; end-to-end DDL+DML+contract checks; parametrized quoting consistency.
- `test_conformance_discovery.py` (9 tests) — internal object filtering, view/MV contract checks, type round-trip for all SodaDataTypeNames, synonym bidirectionality, column type parameters (precision/scale/length).
- `test_conformance_types_dialect.py` (32 tests) — missing check per data type, aggregate per numeric type, schema check, sampling SQL execution per sampler, regex via invalid check, RANDOM() value range, forward/reverse type mapping, synonym consistency.

Designed to catch the top 3 categories of field bugs: identifier quoting (~30%), metadata discovery (~20%), and type mapping/dialect (~18%).

## Bugs found and fixed

While running this suite across all adapters, the synonym bidirectionality test found two real bugs in the synonym definitions. Both are fixed in this PR:

- **Redshift** (`soda-redshift/.../redshift_data_source.py`): `float` was declared a synonym of `real`/`float4`, but its reverse mapping correctly resolves it to `DOUBLE` (it is an alias for `float8`/`double precision`). Fix: move `float` out of `["real", "float4"]` and into `[REDSHIFT_DOUBLE_PRECISION, "float8", "float"]`.
- **DuckDB** (`soda-duckdb/.../duckdb_data_source.py`): two synonym groups lumped together types that map to different SodaDataTypes — all integer types were grouped with `decimal`/`numeric`, and single-precision floats were grouped with double-precision floats. Fix: split into nine correct synonym groups; drop names that aren't DuckDB types (`number`, `byteint`, `timestamp_ntz`, `timestamp_ltz`).

## Other changes from review

- Relaxed `datetime_precision == N` to `>= N` in `test_column_type_parameters_preserved`. Trino connectors (e.g. iceberg) normalize datetime precision to a connector-specific default (often 6) regardless of DDL — the relaxed check still catches "no precision returned at all" while accepting normalization.
- Case-insensitive identifier preservation check in `test_ddl_and_dml_quoting_both_preserve_identifier` — some dialects fold case during quoting.
- Dropped two redundant private-API tests (`test_sampling_sql_generation`, `test_regex_sql_generation`) — coverage is preserved by the end-to-end tests in the same file. Remaining private-API uses (`_get_data_type_name_synonyms`, `_data_type_name_synonym_mappings`) are kept with explanatory comments because no public accessor exists for the synonym table itself.
- Removed gratuitous always-true assertions (`assert result is not None` on a typed return, `assert len(rows) >= 0`).

## Results per adapter (latest CI on this branch)

To be updated once CI completes. Prior run was 620 passed / 46 skipped / 1 failed (Redshift float synonym, now fixed). DuckDB and Trino-S3 also failed and are now fixed.

## Test plan
- [x] All conformance tests pass on DuckDB locally (61/64, 3 capability skips)
- [ ] CI green on duckdb, redshift, trino-s3 after this push
- [ ] CI green on the rest of the matrix (postgres, sqlserver, snowflake, bigquery, fabric, synapse, athena, databricks, sparkdf, trino-postgres)
- [ ] SonarCloud quality gate — gratuitous assertions removed; will know on next analysis

### DIFF
--- a/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
+++ b/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
@@ -128,14 +128,19 @@ class DuckDBSqlDialect(SqlDialect, sqlglot_dialect="duckdb"):
 
     def _get_data_type_name_synonyms(self) -> list[list[str]]:
         # Implements data type synonyms
-        # Each list should represent a list of synonyms
+        # Each list should represent a list of synonyms — every member of a
+        # group must reverse-map to the same SodaDataTypeName (or to a Soda
+        # type pair that is_same_soda_data_type_with_synonyms treats as equal).
         return [
             ["varchar", "text", "string"],
-            ["number", "decimal", "numeric", "int", "integer", "bigint", "smallint", "tinyint", "byteint"],
-            ["float", "float4", "float8", "double", "double precision", "real"],
-            ["timestamp", "datetime", "timestamp_ntz", "timestamp without time zone"],
-            ["timestamp_ltz", "timestamp with local time zone"],
-            ["timestamp_tz", "timestamp with time zone"],
+            ["decimal", "numeric"],
+            ["smallint", "int2"],
+            ["integer", "int", "int4"],
+            ["bigint", "int8"],
+            ["real", "float4", "float"],
+            ["float8", "double", "double precision"],
+            ["timestamp", "datetime", "timestamp without time zone"],
+            ["timestamptz", "timestamp with time zone"],
         ]
 
     def get_data_source_data_type_name_by_soda_data_type_names(self) -> dict:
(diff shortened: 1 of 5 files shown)


m1n0 (approved):
NIT: `conformance` without context isn't clear, I had to skim through tests to understand the motivation of this PR. Would it make sense to a) rename or b) put these tests in a descriptive folder?

m1n0 (commented):
One more thought @paulteehan  - we are adding a whole bunch if e2e tests here, I am a bit worried about bloating the test suite. A lot of these e2e tests are very similar to check type tests (e.g. missing etc) just with the variation in table/column name using a special character or something like that. Wouldn't it make sense to either a) consolidate these into less tests overall or b) unit test generated queries to make sure identifiers are handled correctly?

PR #2665, opened 2026-04-15
DES-291: Support numeric precision/scale and datetime precision in schema checks

m1n0:
## Summary
- Wire `numeric_precision`, `numeric_scale`, and `datetime_precision` through from contract YAML parsing → schema check expected columns → mismatch reporting
- Add `include_type_parameters` param to `IContractGenerator` interface (default `True`)
- Extend `ColumnDataTypeMismatch` to report precision/scale/datetime mismatches with proper formatting
- Integration tests verify precision match/mismatch across all data sources
- Unit tests verify YAML parsing populates precision fields on expected columns

## Context
Schema checks silently ignored precision parameters because the YAML parser and expected-column builder only wired through `character_maximum_length`. The comparison engine (`is_same_data_type_for_schema_check`) already supported all five parameters. HelloFresh flagged this as a blocker — their generated contracts included precision but schema checks ignored it, causing false passes.

## Test plan
- [x] Unit tests: `test_schema_expected_columns_with_numeric_precision`, `test_schema_expected_columns_with_datetime_precision`
- [x] Integration tests: `test_schema_precision_pass`, `test_schema_precision_mismatch` (dialect-aware)
- [x] Manual testing against Postgres

### DIFF
--- a/soda-core/src/soda_core/contracts/contract_generator.py
+++ b/soda-core/src/soda_core/contracts/contract_generator.py
@@ -33,5 +33,6 @@ class IContractGenerator:
         soda_cloud: Optional[SodaCloud],
         use_agent: bool,
         generate_checks: bool = True,
+        include_type_parameters: bool = True,
     ):
         raise NotImplementedError("No implementation for create_skeleton() yet. Please install an appropriate plugin.")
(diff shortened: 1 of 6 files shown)


Niels-b (approved):
LGTM

PR #2669, opened 2026-04-16
Feat: Add internal streaming support

Niels-b:
## Summary
Adds a generic streaming pipeline (`StreamingOrchestrator` / `StreamingProducer` / `StreamingConsumer`) so check implementations can push data to extension consumers as it's produced, decoupling check logic from any DWH-side persistence. Foundation for the reconciliation rows_diff DWH flow built in [soda-extensions #353](https://github.com/sodadata/soda-extensions/pull/353); this PR contains the core primitives, plus a Synapse pagination fix the recon flow needs.

### Streaming infrastructure
New package `soda_core/common/streaming/`:
- `StreamingOrchestrator` — global registry; producers initialize an action and route data to all registered consumers without knowing about them.
- `StreamingProducer` / `StreamingActionHandle` — producer-side API. `initialize_action(...)` returns a context-manager handle whose `__exit__` always calls `finish()` (idempotent, never suppresses producer exceptions).
- `StreamingConsumer` — consumer interface (`initialize` / `process_data` / `finish`).
- Conftest hook in `soda-tests` resets the orchestrator between tests so consumers don't leak.

### Synapse pagination
Synapse Dedicated SQL Pool doesn't support `OFFSET ... FETCH NEXT`, so a custom paginator is required.
- `SynapseSqlDialect.select_all_paginated_sql` is now a `ROW_NUMBER()`-in-CTE paginator that **enumerates the column list inside the CTE** and selects only those columns from the outer query, so the synthetic `rn` column never leaks into `cursor.description` (which the rows_diff comparator inspects). Earlier iterations used a JOIN-back on key columns; that strategy was abandoned because `NULL = NULL` is false (drops rows with NULL keys silently) and non-unique keys amplify rows past the page size — both diverge from native OFFSET/LIMIT semantics on other dialects.
- When the caller doesn't pass an explicit column list (e.g. rows_diff with no `source_columns`/`target_columns` in YAML), Synapse resolves columns once via `INFORMATION_SCHEMA` using a back-reference set by `SynapseDataSourceImpl._create_sql_dialect`. No other dialect is touched — they paginate with native OFFSET/LIMIT and `SELECT *` works fine.

### SQL-injection defense in the Synapse paginator
`SynapseSqlDialect._quote_identifier_safe` performs the standard T-SQL `]` → `]]` escape inside `[...]`-quoted identifiers. The paginator's `columns_csv` / `order_by_csv` interpolations now use this helper instead of `quote_default`, closing the Aikido-flagged identifier injection vector without changing `SqlServerSqlDialect.quote_default` (which is shared by other dialects). Addresses Aikido findings on lines 125/129.

### Misc cleanups
- `StreamingProducer.initialize_action` caches `produced_actions()` once instead of calling it twice (Copilot review + SonarQube confusing-string-concat fix).

## Test plan
- [x] Streaming unit tests pass (`soda-tests/tests/unit/test_streaming.py`)
- [x] Synapse rows_diff e2e (in [soda-extensions #353](https://github.com/sodadata/soda-extensions/pull/353)) passes
- [x] Postgres + SqlServer DWH suites pass (no regression from quote_default scope)

### DIFF
--- a/soda-core/src/soda_core/common/streaming/streaming_orchestrator.py
+++ b/soda-core/src/soda_core/common/streaming/streaming_orchestrator.py
@@ -10,9 +10,18 @@ logger = soda_logger
 
 
 class StreamingOrchestrator:
-    """Singleton (class-level state) that manages streaming consumer registration and action dispatch.
+    """Module-level registry of streaming consumers (class-level mutable state).
 
-    Follows the ContractVerificationHandlerRegistry pattern.
+    Same shape as `ContractVerificationHandlerRegistry`: consumers register at plugin-load
+    time and stay registered for the process lifetime; producers ask the orchestrator to
+    initialize an action and receive a handle that dispatches data to all matching consumers.
+
+    Concurrency: consumers carry per-instance state (e.g. `ReconRowsDiffDwhConsumer` holds
+    `_pending_rows`, `_dwh_data_source_impl`, `_context`). The current contract-verification
+    flow is sequential, so a single consumer instance is reused across actions without
+    overlap. If/when verifications run concurrently, this design must change — either
+    consumers must be stateless and key all per-action state on the action handle, or each
+    `initialize_action` must hand out fresh consumer instances.
     """
 
     _consumers: list[StreamingConsumer] = []
@@ -57,6 +66,10 @@ class StreamingOrchestrator:
 
     @classmethod
     def reset(cls) -> None:
-        """Clear all state. Used in tests."""
+        """Clear all registered consumers.
+
+        Primarily used by test fixtures to isolate consumer registration between tests, but
+        also valid in any setup that tears down and re-initializes the plugin layer.
+        """
         cls._consumers = []
         cls._action_consumers = {}
--- a/soda-core/src/soda_core/common/streaming/streaming_orchestrator.py
+++ b/soda-core/src/soda_core/common/streaming/streaming_orchestrator.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+from typing import Any
+
+from soda_core.common.logging_constants import soda_logger
+from soda_core.common.streaming.streaming_action_handle import StreamingActionHandle
+from soda_core.common.streaming.streaming_consumer import StreamingConsumer
+
+logger = soda_logger
+
+
+class StreamingOrchestrator:
+    """Singleton (class-level state) that manages streaming consumer registration and action dispatch.
+
+    Follows the ContractVerificationHandlerRegistry pattern.
+    """
+
+    _consumers: list[StreamingConsumer] = []
+    _action_consumers: dict[str, list[StreamingConsumer]] = {}
+
+    @classmethod
+    def register_consumer(cls, consumer: StreamingConsumer) -> None:
+        cls._consumers.append(consumer)
+        cls._rebuild_action_map()
+
+    @classmethod
+    def _rebuild_action_map(cls) -> None:
+        action_map: dict[str, list[StreamingConsumer]] = {}
+        for consumer in cls._consumers:
+            for action in consumer.supported_actions():
+                action_map.setdefault(action, []).append(consumer)
+        cls._action_consumers = action_map
+
+    @classmethod
+    def initialize_action(cls, action: str, context: dict[str, Any] | None = None) -> StreamingActionHandle:
+        """Initialize matching consumers and return a handle for streaming data.
+
+        Consumers that fail to initialize are excluded from the handle.
+        """
+        if context is None:
+            context = {}
+
+        matching_consumers = cls._action_consumers.get(action, [])
+        active_consumers: list[StreamingConsumer] = []
+
+        for consumer in matching_consumers:
+            try:
+                consumer.initialize(action, context)
+                active_consumers.append(consumer)
+            except Exception:
+                logger.warning(
+                    f"Streaming consumer {type(consumer).__name__} failed to initialize for action '{action}'",
+                    exc_info=True,
+                )
+
+        return StreamingActionHandle(action=action, consumers=active_consumers)
+
+    @classmethod
+    def reset(cls) -> None:
+        """Clear all state. Used in tests."""
+        cls._consumers = []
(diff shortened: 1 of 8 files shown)


soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
```diff
@@ -72,6 +78,18 @@ def _build_insert_into_values_row_sql(self, values: VALUES_ROW) -> str:
...
+        """T-SQL-safe identifier quoting that escapes closing brackets.
+
+        `quote_default` (inherited from SqlServerSqlDialect) wraps in `[...]` without escaping
+        any `]` characters embedded in the identifier — fine in practice for trusted
+        identifiers, but allows SQL injection if the identifier is not pre-validated. The
+        Synapse paginator below interpolates column names directly into a hand-built SQL
+        string (rather than going through the AST), so it must defend against this. The T-SQL
+        rule for bracket-quoted identifiers is to double any embedded `]`.
+        """
+        return f"[{identifier.replace(']', ']]')}]"
```
m1n0, line 108: should this fix happen in sqlserver then perhaps?

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
```diff
@@ -81,32 +99,56 @@ def select_all_paginated_sql(
...
+        return (
+            f"WITH paginated AS (\n"
+            f"    SELECT {columns_csv}, ROW_NUMBER() OVER (ORDER BY {order_by_csv}) AS {rn_alias}\n"
+            f"    FROM {qualified_table}\n"
+            f"    {where_sql}\n"
+            f")\n"
+            f"SELECT {columns_csv}\n"
+            f"FROM paginated\n"
+            f"WHERE {rn_alias} > {offset} AND {rn_alias} <= {offset + limit}\n"
+            f"ORDER BY {rn_alias};"
```
m1n0, line 166: thanks for addressing this. It would be a head-scratcher if it had to be done by hand 😅

soda-core/src/soda_core/common/streaming/streaming_orchestrator.py
```diff
@@ -0,0 +1,62 @@
...
+
+from soda_core.common.logging_constants import soda_logger
+from soda_core.common.streaming.streaming_action_handle import StreamingActionHandle
+from soda_core.common.streaming.streaming_consumer import StreamingConsumer
+
+logger = soda_logger
+
+
+class StreamingOrchestrator:
+    """Singleton (class-level state) that manages streaming consumer registration and action dispatch.
```
m1n0: 1. The docstring claim is misleading. This is class-level mutable state, not a singleton. Any subclass would share the same list; any caller who writes StreamingOrchestrator._consumers = ... from outside replaces it.
  Compare to the cited precedent ContractVerificationHandlerRegistry — same pattern, but I'd call it a "module-level registry" in the docstring rather than "singleton."
  2. The state is keyed by class, so two checks running concurrently (e.g. the future case where contract verifications parallelize) all share consumers and rely on the consumer's per-instance state being fresh. The
  consumer in this PR (ReconRowsDiffDwhConsumer) holds instance state — _active, _pending_rows, _dwh_data_source_impl, _context — that gets clobbered if initialize() is re-invoked while a previous finish() hasn't run.
  As-is, this works because checks are sequential. The pipeline doesn't enforce or document that, and the next person who copies the pattern won't know.
  3. reset() is documented "Used in tests" (:60). It's also called in two production-adjacent places I'd want explicit (test conftests). The class lifecycle is "configure once at plugin init, never tear down" — fine, but
   say so.

  If you want a defensible singleton, instantiate StreamingOrchestrator once and pass it through; if you want a process-global registry, drop "Singleton" and own that. This is the kind of decision worth pinning before
  this gets templated.

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
```diff
@@ -32,7 +32,13 @@ def __init__(self, data_source_model: SynapseDataSourceModel, connection: Option
         super().__init__(data_source_model=data_source_model, connection=connection)
 
     def _create_sql_dialect(self) -> SqlDialect:
-        return SynapseSqlDialect()
+        # Hand the dialect a back-reference to this data source impl so its paginator can
+        # resolve the table's column list via INFORMATION_SCHEMA when the caller didn't
+        # enumerate columns. This is Synapse-specific because the ROW_NUMBER()-based
+        # paginator below needs explicit columns to avoid leaking the synthetic rn column.
+        dialect = SynapseSqlDialect()
+        dialect._data_source_impl = self
```
m1n0: - The fallback to "fetch all columns from INFORMATION_SCHEMA" silently makes a network round-trip that callers don't expect. If a caller is paginating a 10K-call loop without supplying columns, the dialect will issue
  10K metadata queries (unless get_columns_metadata is cached upstream — worth verifying for the Synapse data source).
  - Cleaner option: make columns mandatory at the interface level (it already needs to be non-empty for Synapse). That breaks two callers but makes the contract explicit. Or: have SynapseDataSourceImpl build the dialect
  with a constructor parameter, e.g. SynapseSqlDialect(data_source_impl=self) rather than monkey-patching.
  - The getattr(self, "_data_source_impl", None) defensive read is a code smell — only SynapseDataSourceImpl constructs SynapseSqlDialect, so the attribute will always be set in practice. The defensive None branch raises
   a ValueError that no caller expects to handle.

m1n0 (approved):
looks good overall, a few minor things

soda-synapse/src/soda_synapse/common/data_sources/synapse_data_source.py
Niels-b: Tried that earlier and it blew up — modifying `SqlServerSqlDialect.quote_default` to escape `]` triggered an unrelated regression in `test_multiple_runs_in_sequence` (something downstream is passing already-bracketed names through `quote_default` and would get double-escaped). Reverted to the Synapse-only helper to keep this PR focused.

PR #2671, opened 2026-04-23
feat(DES-294): validate column type parameters at contract parse time

m1n0:
## Summary
- Adds parse-time validation in `ColumnYaml` rejecting type params that don't match the column's data_type family (e.g. `numeric_precision` on VARCHAR, `character_maximum_length` on DECIMAL).
- Two-tier resolution: direct match against `SodaDataTypeName` enum values (tolerates parenthesized suffixes like `varchar(255)`), then dialect-aware fallback via `SqlDialect.get_soda_data_type_name_by_data_source_data_type_names()` when a `primary_data_source_impl` is provided (so Databricks `string` resolves to canonical TEXT).
- The yes/no on whether a param applies to a data_type is delegated to the bound `SqlDialect.data_type_has_parameter_*` methods (already used by `extract_*` on info_schema), not a hardcoded list — one override per dialect now drives both schema-check extraction *and* contract validation.
- For non-dialect paths (`contract publish`, `contract verify` without a data source), a private `_DefaultSqlDialect` fallback keeps parse-time validation alive so contracts with obvious mismatches still get flagged before publish.
- Types that resolve to neither are skipped — the schema check still catches real DB mismatches at scan time.

## Why
Before this change, contracts like

```yaml
columns:
  - name: amount
    data_type: DECIMAL
    character_maximum_length: 255
```

were silently accepted at parse time and only surfaced at scan time as a confusing "data type mismatch" against the live DB. Follow-up to sodadata/soda#12029 where we added form-side sanitization — this closes the CLI gap.

## Test plan
- [x] Added 13 unit tests in `test_contract_yaml_parsing.py` covering:
  - Valid params on each family (VARCHAR, DECIMAL/NUMERIC, TIMESTAMP) — no error
  - Parenthesized `varchar(255)` normalized before lookup
  - Invalid mismatches log an error naming the param, the column, and the offending data_type
  - DECIMAL↔NUMERIC same-family passes
  - Missing data_type skips validation (parser handles that elsewhere)
  - Dialect-aware: Databricks-style `string` resolves to TEXT — valid char param passes, numeric param flagged
  - No dialect + unknown native type (`string`) is skipped
- [x] `pytest soda-tests/tests/unit/contract_yaml/test_contract_yaml_parsing.py` → 23/23 passed
- [x] `pre-commit run` → all hooks pass (isort, black, autoflake, etc.)

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/contract_yaml.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_yaml.py
@@ -79,6 +79,12 @@ class ContractYaml:
         self.contract_yaml_source: ContractYamlSource = contract_yaml_source
         self.contract_yaml_object: YamlObject = contract_yaml_source.parse()
         self.primary_data_source_impl: Optional[DataSourceImpl] = primary_data_source_impl
+        # Cache the dialect's native→canonical data type mapping once — dialects construct this dict fresh on every call.
+        self._native_to_soda_data_type_mapping: Optional[dict[str, SodaDataTypeName]] = (
+            primary_data_source_impl.sql_dialect.get_soda_data_type_name_by_data_source_data_type_names()
+            if primary_data_source_impl is not None
+            else None
+        )
 
         self.variables: list[VariableYaml] = self._parse_variable_yamls(contract_yaml_source, provided_variable_values)
 
@@ -435,13 +441,16 @@ _DATETIME_PRECISION_FAMILY: frozenset[SodaDataTypeName] = frozenset(
 )
 
 
-def _resolve_to_soda_type(data_type: Optional[str], sql_dialect: Optional["SqlDialect"]) -> Optional[SodaDataTypeName]:
+def _resolve_to_soda_type(
+    data_type: Optional[str],
+    native_to_soda_mapping: Optional[dict[str, SodaDataTypeName]],
+) -> Optional[SodaDataTypeName]:
     """
     Resolve a raw YAML data_type string to a canonical SodaDataTypeName.
 
     Order:
-    1. Direct match against SodaDataTypeName values (no dialect needed).
-    2. Dialect-aware resolution via native→canonical mapping (e.g. Databricks "string" → TEXT).
+    1. Direct match against SodaDataTypeName values (no mapping needed).
+    2. Dialect-aware resolution via the provided native→canonical mapping (e.g. Databricks "string" → TEXT).
     3. Give up (return None) — caller skips type-param validation for unknown types.
     """
     if not data_type:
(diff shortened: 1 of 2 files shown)


m1n0:
Addressed both Copilot points in f8ff5bd:

1. **Caching** — confirmed dialects build the mapping dict fresh on every call (e.g. `postgres_data_source.py`, `duckdb_data_source.py` both `return { ... }`). Moved the lookup to `ContractYaml.__init__` so it runs once per parse; `ColumnYaml._validate_type_parameters` now receives the cached `dict[str, SodaDataTypeName]` directly and passes it to `_resolve_to_soda_type`. For a 100-column contract with unresolvable native types this saves 99 dict constructions.

2. **Naming** — renamed the local `checks` list to `type_param_validations` to avoid confusion with `ContractYaml._parse_checks` / `ColumnYaml.check_yamls`.

All 23 unit tests still pass; pre-commit clean.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -411,6 +419,52 @@ def __init__(self, yaml_object: YamlObject):
...
+        SodaDataTypeName.CHAR,
+        SodaDataTypeName.TEXT,
+    }
+)
+_NUMERIC_PRECISION_FAMILY: frozenset[SodaDataTypeName] = frozenset(
+    {
+        SodaDataTypeName.DECIMAL,
+        SodaDataTypeName.NUMERIC,
+    }
+)
```
Niels-b: Do we maybe need to include `FLOAT` and `DECIMAL` here as well? (Can't remember of the top of my head, but may be worth looking into).

Niels-b (approved):
LGTM!

Only one small remark/thought

soda-core/src/soda_core/contracts/impl/contract_yaml.py
m1n0: Good catch — addressed in c5625cf, though differently than just adding FLOAT/DOUBLE.

`DECIMAL` was already in the family (line 431, alongside `NUMERIC`). For `FLOAT`/`DOUBLE` the dialect's existing `data_type_has_parameter_numeric_precision`/`numeric_scale` (sql_dialect.py:252-256) only return True for `numeric`/`number`/`decimal` — so adding them only to the contract-yaml family would have diverged from how schema-check extraction actually behaves.

Instead, I removed the hardcoded `_CHAR_LENGTH_FAMILY`/`_NUMERIC_PRECISION_FAMILY`/`_DATETIME_PRECISION_FAMILY` frozensets entirely and made parse-time validation defer to `dialect.data_type_has_parameter_*` directly. The dialect is now the single source of truth — if a future dialect decides FLOAT does take precision (e.g. SQL standard FLOAT(p) for binary precision), one override does both extraction *and* contract validation in one go.

For non-dialect paths (`contract publish`, `contract verify` without a data source) there's a private `_DefaultSqlDialect` fallback so those paths still get parse-time validation rather than silently skipping it. The class docstring explains the rationale.

m1n0:
triggering re-review @Niels-b as I decided to tweak the approach a bit

Niels-b (approved):
New approach looks good. 

I was debating using postgres/duckdb dialects instead of a "default" one. But that introduces risk when we change something in postgres/duckdb. So this seems like a good option 👍 . Well done 👏

PR #2672, opened 2026-04-26
fix: return 1-based column index in TrinoDataSource to match Trino ordinal_position (#2491)

### LINKED ISSUE #2491: SODA core fails Trino column index check
soda-core==3.5.6
soda-core-trino==3.5.6

SODA core fails Trino column index check
because SODA returns 0-based order while Trino starts from number 1
so all column index/order checks fail

devimano2011:
Fixes #2491

## Problem

Trino uses 1-based `ordinal_position` in `INFORMATION_SCHEMA.COLUMNS`, but the base `get_column_index()` in `SqlDialect` returns a 0-based Python list index. This mismatch caused all column index/order checks to fail in Trino.

## Fix

Added a single-line override of `get_column_index()` in `TrinoDataSource` that adds 1 to the base result.

## Testing

Column index checks against Trino will now return the correct 1-based index matching


devimano2011:
pre-commit.ci autofix

PR #2675, opened 2026-04-27
feat(core): add DataSourceConnection.create_additional_dbapi_connection [DES-308]

m1n0:
## Summary

Adds a single base-class method `create_additional_dbapi_connection(self) -> object` to `DataSourceConnection`. Default implementation delegates to the existing abstract `_create_connection(self.connection_properties)`. No changes to any of the 17 concrete subclasses.

Enables features that need **two independent DBAPI cursors against the same data source**, where sharing `self.connection` would require concurrent cursors — unreliable across drivers (Snowflake, BigQuery, MSSQL, etc.).

## Why

The immediate consumer is the within-database `rows_diff` reconciliation check in soda-extensions ([DES-309](https://linear.app/sodadata/issue/DES-309) — paired PR: sodadata/soda-extensions#TBD). `rows_diff` opens two raw DBAPI cursors and runs a paginated merge-join. When source DS == target DS, both cursors would land on the same shared connection — which several drivers cannot reliably support. This method gives the caller a fresh DBAPI connection sharing the same connection properties, so each cursor lives on its own connection.

The caller owns the returned connection and must close it.

## What changed

`soda-core/src/soda_core/common/data_source_connection.py`:

```python
def create_additional_dbapi_connection(self) -> object:
    """
    Open and return a fresh raw DBAPI connection using the same
    connection properties as this one. The caller owns the returned
    connection and MUST close it.
    ...
    """
    return self._create_connection(self.connection_properties)
```

Inserted between `close_connection` and `__exit__`. Returns a raw DBAPI connection object — matches the existing return type of `_create_connection`. No new imports.

## What did not change

- Any concrete `DataSourceConnection` subclass.
- The primary connection (`self.connection`) lifecycle.
- Any existing call site.

## Tests

New file `soda-tests/tests/unit/test_additional_dbapi_connection.py` with 3 unit tests using a minimal in-process `DataSourceConnection` subclass:

1. `test_create_additional_dbapi_connection_returns_fresh_connection` — returns a new object distinct from `self.connection`.
2. `test_create_additional_dbapi_connection_uses_stored_properties` — passes the stored `connection_properties` to `_create_connection`.
3. `test_create_additional_dbapi_connection_does_not_affect_primary_on_close` — closing the additional connection does not affect the primary.

```
$ uv run pytest soda-tests/tests/unit/test_additional_dbapi_connection.py -v
3 passed in 0.01s
```

The full `soda-tests/tests/unit` suite passes (614 passed, 1 skipped). 16 unrelated tests ERROR on this branch and on `main` due to no local Postgres on `:5432`; not regressions.

## Test plan

- [x] New unit tests pass
- [x] Full unit suite has no new failures (existing postgres-connection ERRORs are pre-existing environmental)
- [x] Manual smoke: `hasattr(DataSourceConnection, 'create_additional_dbapi_connection') is True`
- [ ] Reviewer: confirm the default base-class implementation is appropriate for all 17 concrete subclasses

## Linear

- [DES-308](https://linear.app/sodadata/issue/DES-308) — this PR
- [DES-309](https://linear.app/sodadata/issue/DES-309) — paired soda-extensions consumer (blocked by this)
- Project: [Within database recon check](https://linear.app/sodadata/project/within-database-recon-check-c5e6871c9265/overview)

## Out of scope

- Connection pooling / caching across calls (one-shot use-case; pool overhead unjustified).
- Per-driver overrides (default implementation works uniformly via existing `_create_connection`).
- A typed `Protocol` for the return value (the existing `_create_connection -> object` is unchanged; nominal typing across DBAPI drivers is a separate cleanup).


Niels-b (commented):
We already have something similar for DWH in soda-extensions.

https://github.com/sodadata/soda-extensions/blob/c3fc4b90b2e42161f73d1efadfebe45e677a26e3/soda/src/soda/data_source_extensions/sqlserver_data_source_extension.py#L78

Maybe we can re-use this or combine the two approaches?

m1n0:
Good catch — agreed, two parallel mechanisms isn't great. Pushed `87bebfde` that reshapes the API:

- **Method moved from `DataSourceConnection` → `DataSourceImpl`** and renamed to `create_additional_connection() -> DataSourceConnection`. It delegates to the existing abstract `_create_data_source_connection()`, so still zero per-subclass changes.
- Returns a wrapped `DataSourceConnection` (not raw DBAPI) — consistent with what your `ExtraDwhConnection` expects.

Paired soda-extensions update on sodadata/soda-extensions#359 (`23c2ae78`):

- `SqlServerDataSourceExtension.get_connection_for_dwh()` now uses the new core method — drops the `_create_data_source_connection()` + `.open_connection()` pair (the `open_connection()` was already redundant since `DataSourceConnection.__init__` auto-opens):
  ```python
  def get_connection_for_dwh(self) -> ExtraDwhConnection:
      return ExtraDwhConnection(
          data_source_connection=self.data_source_impl.create_additional_connection(),
          main_dwh_connection=self.data_source_impl.connection,
      )
  ```
- `rows_diff` (within-database) uses the same primitive directly. Manual close in `try/finally` since rows_diff doesn't need the conditional-close-vs-main-connection logic that `ExtraDwhConnection` provides.

Net: one canonical "open another connection" primitive in core; the `ExtraDwhConnection` RAII wrapper stays in extensions where the conditional-close semantics are useful. Tests still pass on both branches (3 unit tests in core; 49 reconciliation tests + 7 new within-DB tests on duckdb).

Niels-b (approved):
LGTM

Issue #356, opened 2021-06-18
[sql server] Soda analyse  can't handle columns that are reserved names


pbuxtonccuk:
**Describe the bug**
Running soda analyze against a SQLServer database where some tables have columns with the same name as a reserved keyword fails with 'incorrect syntax'

There should be an option to encapsulate columns names with [], or it should do it  always

**To Reproduce**
1. Create a test database with a table containing the columns 'myid' and a column named 'order'
1. Modify warehouse.yml and the credentials yml to suit to conenct to the database
2. Run `soda analyze'

**OS**: 
Centos 7

**Python Version**:
3.8.1

**Soda SQL Version**:
Latest Soda SQLServer beta

**Warehouse Type**:
SQLServer 2017

pbuxtonccuk:
I am aware that I could manually create the various scan.ymls per table, but I'd rather not.  There's a lot if them .........

vijaykiran:
Fixed in latest tag!

PR #2678, opened 2026-04-28
feat: add first-class SELECT DISTINCT support to SQL AST

paulteehan:
## Summary

- Add `distinct: bool = False` to the `SELECT` AST node so callers can write `SELECT(fields=[...], distinct=True)` and get `SELECT DISTINCT col1, col2 FROM ...` natively
- `_build_select_sql_lines` emits `SELECT DISTINCT` when any incoming `SELECT` element has `distinct=True`
- Existing `DISTINCT` expression node and `_build_distinct_sql` are intentionally untouched — they remain correct for aggregate-level use (`COUNT(DISTINCT x)`, `SUM(DISTINCT x)`)

## Why

SQL uses `DISTINCT` in two semantically different positions:

| Feature | Example | Semantics |
|---|---|---|
| Set quantifier on SELECT | `SELECT DISTINCT a, b FROM t` | Dedupes the whole result set |
| Aggregate-level modifier | `COUNT(DISTINCT a)` | Per-aggregate input dedup |

Today the AST only models the second. The first has no representation, and the workaround of nesting `DISTINCT([a, b, c])` into `SELECT.fields` renders as `SELECT DISTINCT(a, b, c)` — accepted loosely by Postgres/MySQL but rejected by DB2, Athena/Presto/Trino, and BigQuery, where parens around the field list are interpreted as a row constructor and change semantics. The existing warning at `sql_ast.py:61-67` flags this exact shape as wrong.

This was discovered during review of [soda-extensions PR #356](https://github.com/sodadata/soda-extensions/pull/356), which currently ships:

```python
paginated_sql = data_source.sql_dialect.build_select_sql(statements)
paginated_sql = paginated_sql.replace(\"SELECT\", \"SELECT DISTINCT\", 1)
```

After this change lands, that caller (and any future ones) can be migrated to `SELECT(fields=columns, distinct=True)` and the string-replace hack deleted.

## What changed

- `soda-core/src/soda_core/common/sql_ast.py` — `SELECT` gets `distinct: bool = False`; warning message updated to point callers at the new flag
- `soda-core/src/soda_core/common/sql_dialect.py` — `_build_select_sql_lines` reads `select_element.distinct` and prepends `DISTINCT` to the keyword. Continuation-line indent kept at 7 spaces (no realignment to `SELECT DISTINCT `'s width — still readable)
- `soda-postgres/tests/unit/test_postgres_dialect.py` — 6 new tests

## Acceptance criteria (from internal plan)

- [x] `SELECT(fields=[\"a\", \"b\"], distinct=True)` renders `SELECT DISTINCT a, b` (no parens)
- [x] `SELECT(fields=[\"a\"])` continues to render `SELECT a` unchanged
- [x] `COUNT(DISTINCT(expression=\"x\"))` continues to render `COUNT(DISTINCT(x))`
- [x] All existing call sites work without change (default `distinct=False`)
- [x] Unit tests cover: single col, multiple cols, full paginated shape (WHERE/ORDER BY/LIMIT/OFFSET), aggregate-level DISTINCT preserved, default-false

## Test plan

- [x] `uv run pytest soda-{postgres,bigquery,athena,databricks,duckdb,redshift,snowflake,sparkdf,sqlserver,synapse,trino,fabric}/tests/unit` — 30/30 passed
- [x] Full-suite collection: 905 tests collect cleanly, no import regressions
- [ ] Integration matrix (run in CI) — especially DB2, Athena, BigQuery, Trino once soda-extensions caller is migrated
- [ ] Follow-up: migrate `soda-reconciliation/.../reference_diff_check.py` once [soda-extensions PR #356](https://github.com/sodadata/soda-extensions/pull/356) merges, then `grep -rn '.replace(\"SELECT\", \"SELECT DISTINCT\"' src/` should return zero hits

## Out of scope

- Removing/renaming the existing `DISTINCT` expression node — it has legitimate aggregate-level users
- The `TODO: refactor build_select_sql to use AST` at `sql_dialect.py:669` — independent and larger
- `SELECT DISTINCT ON (...)` (Postgres-specific) — not needed here

### DIFF
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -53,17 +53,22 @@ class BaseSqlExpression:
 @dataclass
 class SELECT(BaseSqlExpression):
     fields: SqlExpression | str | list[SqlExpression | str]
+    distinct: bool = False
 
     def __post_init__(self):
         super().__post_init__()
         self.handle_parent_node_update(self.fields)
 
-        # Check that the select contains a distinct and has multiple fields -> give a warning
+        # A DISTINCT expression inside SELECT.fields renders as DISTINCT(...) with
+        # parens, which is only correct inside an aggregate (e.g. COUNT(DISTINCT x)).
+        # For the set-quantifier form (SELECT DISTINCT a, b FROM t) use distinct=True.
         if isinstance(self.fields, list) and len(self.fields) > 1:
             if any(isinstance(field, DISTINCT) for field in self.fields):
                 logger.warning(
-                    """Found DISTINCT in a SELECT statement with multiple fields.
-                               This might have unintended consequences."""
+                    "DISTINCT expression inside SELECT fields. "
+                    "For set-quantifier deduplication, use SELECT(..., distinct=True). "
+                    "The DISTINCT expression node is intended for aggregate-level use "
+                    "(e.g. COUNT(DISTINCT x))."
                 )
 
 
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -692,8 +692,7 @@ class SqlDialect:
         select_field_sqls: list[str] = []
         for select_element in select_elements:
             if isinstance(select_element, SELECT):
-                if select_element.distinct:
-                    distinct = True
+                distinct = select_element.distinct
                 if isinstance(select_element.fields, str) or isinstance(select_element.fields, SqlExpression):
                     select_element.fields = [select_element.fields]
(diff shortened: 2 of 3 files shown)


Lepo00:
Thank you very much for this @paulteehan!

Niels-b (changes requested):
Not sure if we should do it like this. We already have a `DISTINCT` in the AST, you should just use that, no? 

In my opinion we should be able to do `SELECT(DISTINCT([my_elements]))`. Open for discussion.

Also, what are we assuming will happen here? If we have a `SELECT DISTINCT` that the entire row is "distinct"?

PR #2679, opened 2026-04-28
fix(core): qualify check_filter columns in valid_reference_data JOIN (DTL-1517)

paulteehan:
## Summary

Closes [DTL-1517](https://linear.app/sodadata/issue/DTL-1517). 

A check-level \`filter:\` on a \`valid_reference_data\` check produced ambiguous SQL whenever the filter referenced a column that also exists in the **reference** dataset. Concrete shape:

\`\`\`yaml
columns:
  - name: country
    valid_reference_data:
      dataset: <ds>/<schema>/valid_countries
      column: country_code
    checks:
      - invalid:
          filter: |
            id > 1            # `id` exists in both source AND reference table
\`\`\`

The reference-check \`InvalidReferenceCountQuery\` rendered the filter un-aliased:

\`\`\`sql
SELECT COUNT(*)
FROM "_soda_filtered_dataset" AS "C"
     LEFT JOIN "_soda_filtered_referenced_dataset" AS "R" ON "C"."country" = "R"."country_code"
WHERE (id > 1)                  -- ambiguous: id is in both C and R
  AND NOT("C"."country" IS NULL) AND "R"."country_code" IS NULL;
\`\`\`

DuckDB rejects with \`Binder Error: Ambiguous reference to column name "id"\`. Postgres / Snowflake / DB2 all produce the same error. The check ends up \`NOT_EVALUATED\`.

A previous attempt at this fix ([#2535](https://github.com/sodadata/soda-core/pull/2535), closed without merging) renamed the JOIN aliases from \`C\` / \`R\` to \`TARGET\` / \`REFERENCE\` and required users to manually qualify their filter columns. That approach was abandoned — Milan confirmed in the recent Slack thread that "this is known and there is no current workaround." This PR takes a different, transparent approach: filters keep their plain source-only column references, and we automatically qualify them via sqlglot when rendering the JOIN query.

## How

1. Add \`qualify_unqualified_columns_with_alias\` helper in \`soda_core/common/sql_utils.py\`. Uses sqlglot (already a dependency) to walk \`exp.Column\` nodes and set the source alias \`"C"\` on any column without one. Already-qualified columns / identifiers inside string literals / function names are left alone. Falls back to returning the input unchanged on parse error.
2. Add \`SqlDialect.qualify_unqualified_columns_with_alias(sql, alias)\` wrapper that injects the dialect's \`SQLGLOT_DIALECT\`. Mirrors \`SqlDialect.apply_sampling\`.
3. Call the wrapper in \`InvalidReferenceCountQuery.build_query\` (\`invalidity_check.py:255\`) to qualify the filter against the source alias \`"C"\` before passing it into \`WHERE.optional(SqlExpressionStr.optional(...))\`.

The metric's \`self.check_filter\` string is left untouched: the optimized aggregate row-count query (\`RowCountMetricImpl.sql_expression\` rendered into \`AggregationQuery.build_sql\`) runs against the source CTE alone with no JOIN and no \`AS "C"\` on the FROM, so a literal \`"C".id > 1\` filter would not resolve there.

## Why sqlglot over a regex

The filter can reference any source column, not just the check's column. We don't always have an enumerated column list — the contract YAML may not list every column. sqlglot is dialect-aware and skips identifiers inside string literals, function names, etc.

## DB2 — companion change in soda-extensions

DB2 LUW and z/OS do not have a sqlglot dialect. A companion soda-extensions PR adds a native DB2 sqlglot dialect (registered as \`"db2"\`). The DB2 PR can be reviewed independently — this one is correct on every adapter today regardless.

## Test plan

- \`soda-tests/tests/integration/test_invalid_reference_check.py\` — added \`test_invalid_count_with_check_filter_on_shared_column\`, reproducing the exact ambiguous-column case (referenced table shares the \`id\` column with the source). Pre-fix this test asserts \`assert None == 3\` because the reference query errors with \`Ambiguous reference to column name "id"\`. Post-fix the emitted JOIN query is \`WHERE ("C".id > 1) AND ...\`, and all 6 reference-check tests pass.
- [x] Full DuckDB suite (\`soda-tests/tests\` + \`soda-duckdb/tests\` with \`TEST_DATASOURCE=duckdb\`): 800 passed, 14 skipped, 0 failed.
- [x] Full DB2 LUW integration suite (\`soda-tests/tests/integration\` with \`TEST_DATASOURCE=db2\`, requires the companion soda-extensions PR for the native DB2 dialect): 144 passed, 8 skipped, 0 failed. Every SQL statement that flowed through \`DataSourceConnection.execute_query\` was captured (205 unique) and round-tripped through the DB2 dialect; 196 / 196 of the SELECT-shape statements were re-accepted by live DB2.
- [ ] Reviewer to verify on Postgres / Snowflake / BigQuery in CI.

## Note for reviewers

This was AI-written. Bug analysis, fix design, and tests are mine; please scrutinise.

### DIFF
--- a/soda-core/src/soda_core/common/sql_utils.py
+++ b/soda-core/src/soda_core/common/sql_utils.py
@@ -48,14 +48,22 @@ def qualify_unqualified_columns_with_alias(
 
     Already-qualified columns are left untouched. If the expression cannot be parsed as SQL,
     it is returned unchanged so any database error surfaces unmodified to the user.
+
+    Data sources without a native sqlglot dialect (DB2, for example, sets
+    `sqlglot_dialect=""`) are not rewritten, because the round-trip through sqlglot's default
+    parser/printer can mangle dialect-specific syntax (e.g. DB2 timestamp literals like
+    `'2024-01-01-12.30.00'`). On those data sources the filter is returned unchanged and the
+    pre-fix behaviour applies, so a filter that references a column shared with the reference
+    table will still error at the database — no new regression vs today.
     """
     if not sql_expression or not sql_expression.strip():
         return sql_expression
 
+    if not read_dialect:
+        return sql_expression
+
     try:
-        tree = (
-            sqlglot.parse_one(sql_expression, read=read_dialect) if read_dialect else sqlglot.parse_one(sql_expression)
-        )
+        tree = sqlglot.parse_one(sql_expression, read=read_dialect)
     except sqlglot.errors.ParseError:
         return sql_expression
 
--- a/soda-core/src/soda_core/common/sql_utils.py
+++ b/soda-core/src/soda_core/common/sql_utils.py
@@ -37,6 +37,36 @@ def attach_sample_to_relation(rel: exp.Expression, sampler_limit: Number, sample
         rel.set("sample", build_sample_clause(sampler_limit, sampler_type))
 
 
+def qualify_unqualified_columns_with_alias(
+    sql_expression: str,
+    alias: str,
+    read_dialect: str | None = None,
+    write_dialect: str | None = None,
+) -> str:
+    """
+    Qualify every unqualified column reference in a SQL expression with the given table alias.
+
+    Already-qualified columns are left untouched. If the expression cannot be parsed as SQL,
+    it is returned unchanged so any database error surfaces unmodified to the user.
+    """
+    if not sql_expression or not sql_expression.strip():
+        return sql_expression
+
+    try:
+        tree = (
+            sqlglot.parse_one(sql_expression, read=read_dialect) if read_dialect else sqlglot.parse_one(sql_expression)
+        )
+    except sqlglot.errors.ParseError:
+        return sql_expression
+
+    for column in tree.find_all(exp.Column):
+        if column.table:
+            continue
+        column.set("table", exp.to_identifier(alias, quoted=True))
+
+    return tree.sql(dialect=write_dialect) if write_dialect else tree.sql()
+
+
 def apply_sampling_to_sql(
     sql: str,
     sampler_limit: Number,
--- a/soda-core/src/soda_core/common/sql_utils.py
+++ b/soda-core/src/soda_core/common/sql_utils.py
@@ -44,16 +44,21 @@ def qualify_unqualified_columns_with_alias(
     write_dialect: str | None = None,
 ) -> str:
     """
-    Qualify every unqualified column reference in a SQL expression with the given table alias.
+    Qualify every unqualified top-level column reference in a SQL expression with the given
+    table alias.
 
-    Already-qualified columns are left untouched. If the expression cannot be parsed as SQL,
-    it is returned unchanged so any database error surfaces unmodified to the user.
+    Already-qualified columns are left untouched. Columns inside subqueries are also left
+    untouched — they belong to the subquery's own scope, not the outer alias. If sqlglot
+    cannot parse or print the expression for any reason (parse error, unknown dialect),
+    the input is returned unchanged so any database error surfaces unmodified to the user.
+
+    Note: a successful parse round-trips through sqlglot's printer, which canonicalises some
+    syntax (e.g. ``col::int`` becomes ``CAST(col AS INT)``, function names get uppercased).
+    On the same read/write dialect the rewrite is semantically equivalent in standard cases,
+    but recognised functions may be rewritten into canonical equivalents.
 
     If ``read_dialect`` / ``write_dialect`` are not provided (or are empty), sqlglot's
-    default parser/printer is used. Every Soda data source ships a non-empty
-    ``SQLGLOT_DIALECT`` today — including DB2 LUW + z/OS, both registered as ``"db2"``
-    via ``soda_db2.base.sqlglot_dialect``. If you add a new adapter, make sure its
-    ``SqlDialect`` subclass passes a real sqlglot dialect name in ``sqlglot_dialect=``.
+    default parser/printer is used.
     """
     if not sql_expression or not sql_expression.strip():
         return sql_expression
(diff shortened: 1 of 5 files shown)


soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -332,9 +347,39 @@ def query_join(self) -> SqlExpression:
     def execute(self) -> list[Measurement]:
         try:
             query_result: QueryResult = self.data_source_impl.execute_query(self.sql)
-        except Exception as e:
-            logger.error(msg=f"Could not execute invalid reference query {self.sql}: {e}", exc_info=True)
-            return []
+        except Exception as initial_error:
```
Niels-b, line 350: This will make the retry trigger for any error, even if the connection is closed (or something like that).

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -332,9 +347,39 @@ def query_join(self) -> SqlExpression:
...
+                        f"Retried with the check_filter columns qualified against source alias "
+                        f"{self.referencing_alias!r} (common fix for ambiguous-column errors when "
+                        f"the filter references a column that exists in both the source and "
+                        f"reference datasets); the retry also failed.\n"
+                        f"Retry query: {retry_sql}\n"
+                        f"  Retry error: {retry_error}"
+                    ),
+                    exc_info=True,
+                )
+                return []
```
Niels-b, line 382: Is there a way that we can detect this beforehand? Not really the biggest fan of "try, error, retry" if we can avoid it. These things show up in DB logs. The error message itself is clear. But what if it's not related to the filter? Then we're still showing this error in call cases

soda-core/src/soda_core/contracts/impl/check_types/invalidity_check.py
```diff
@@ -244,21 +244,36 @@ def __init__(
...
             FROM(cte.alias).AS(self.referencing_alias),
-            WHERE.optional(SqlExpressionStr.optional(self.check_filter)),
+            WHERE.optional(SqlExpressionStr.optional(check_filter)),
         ]
 
         query.extend(self.query_join())
 
         return query
 
+    def _build_qualified_filter_sql(self) -> str:
```
Niels-b, line 264: Name is confusing to me, does this build only the filter or the full SQL? (Yes it's in the docs below 😉 ). 
Maybe `_build_query_with_qualified_filter`?

soda-core/src/soda_core/common/sql_utils.py
```diff
@@ -37,6 +37,52 @@ def attach_sample_to_relation(rel: exp.Expression, sampler_limit: Number, sample
...
+
+    Already-qualified columns are left untouched. Columns inside subqueries are also left
+    untouched — they belong to the subquery's own scope, not the outer alias. If sqlglot
+    cannot parse or print the expression for any reason (parse error, unknown dialect),
+    the input is returned unchanged so any database error surfaces unmodified to the user.
+
+    Note: a successful parse round-trips through sqlglot's printer, which canonicalises some
+    syntax (e.g. ``col::int`` becomes ``CAST(col AS INT)``, function names get uppercased).
+    On the same read/write dialect the rewrite is semantically equivalent in standard cases,
+    but recognised functions may be rewritten into canonical equivalents.
```
Niels-b, line 58: Should we also add some test cases for this?

Niels-b (changes requested):
Some remarks that I'd like to see discussed/addressed.

m1n0:
Thanks for picking this up @paulteehan! Sorry for posting a Claude-generated comment but I have some concerns about the approach

```
Retry has a silent-wrong-answer class. Hardcoding "C" as the qualification target is a guess. When wrong, no warning logged:

  - filter: country_code IS NOT NULL — user means R, retry rewrites to "C", plausible wrong number.
  - filter: country_code IS NULL for "didn't match reference" — silently flips to "source value is null."
  - Pseudo-columns (LEVEL, _PARTITIONTIME, METADATA$FILENAME) parsed as exp.Column and prefixed.
  - try/except Exception is broad — typos, transient errors, permission denied all double round-trip and produce a log that implies ambiguity.
  - DB2 without companion PR: unknown-dialect path returns input unchanged, retry runs same SQL, no fix.

  If the shape stays, log a WARNING on successful retry with both filters. Today there's no signal at all.

  Root cause is wider than the PR. check_filter is a scope-blind string used in multiple query shapes:

  - InvalidReferenceCountQuery — JOIN scope, "C"/"R" bound, ambiguity possible.
  - AggregationQuery for check_rows_tested — single-table scope, no aliases, qualified columns wouldn't resolve.

  That asymmetry is why retry-on-error was chosen — pre-qualifying the shared string would break the second consumer. Bug class will recur in any future JOIN-shape consumer (failed-rows sampling, recon, etc.).

  Alternatives

  A. Per-consumer filtered CTE. Wrap source CTE before the JOIN:

  WITH source_cte   AS (SELECT ... FROM source),
       ref_cte      AS (SELECT ... FROM ref),
       filtered_src AS (SELECT * FROM source_cte WHERE <check_filter>)
  SELECT COUNT(*)
  FROM filtered_src AS C
  LEFT JOIN ref_cte AS R ON C.col = R.refcol
  WHERE NOT (C.col IS NULL) AND R.refcol IS NULL;

  Filter runs in single-table scope. No ambiguity. AggregationQuery untouched. Drops sqlglot from this codepath, drops retry, drops false-positive class. Diff comparable to current.

  B. Split into source_filter / target_filter.

  checks:
    - invalid:
        source_filter: id > 1
        target_filter: country_code IS NOT NULL

  Each pushes into its CTE's WHERE. Ambiguity impossible by construction. Backward compat: alias filter → source_filter at parse time (per the PR, no users are qualifying today, so post-JOIN semantics aren't load-bearing
   — confirm with a grep through known contracts).

  Different shape from #2535: that one exposed TARGET/REFERENCE aliases and made users qualify columns. This hides aliases entirely.

  target_filter is also a useful feature on its own (active-only references, soft-delete exclusion). Generalises to any two-relation check.

  Doesn't cover cross-table predicates (source.x AND ref.y); that's an upstream-view problem.

  Net diff smaller than current PR, deletes the new helper rather than adding it.
```

I am not sure those alternatives are 100% better or the way to go, but the silent assumption to qualify with C we are adding here worries me

PR #2682, opened 2026-04-30
fix(core): resolve datasource YAML variables post-parse

paulteehan:
Fixes SAS-11735. soda-core's V4 datasource YAML loader substituted `${ns.var}` references into the YAML byte stream before parsing, which broke ruamel when a secret contained YAML control characters (`'`, `[`, `#`, `&`, ...) — the substituted value re-tokenised and the parser failed at the substitution column. Ports soda-library `f2a926c5` (PR #697) into the equivalent soda-core path: parse first, walk the parsed structure, substitute only string scalars.

🤖 AI-written

### DIFF
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -20,7 +20,7 @@ from soda_core.common.statements.metadata_tables_query import (
     MetadataTablesQuery,
 )
 from soda_core.common.statements.table_types import FullyQualifiedObjectName, TableType
-from soda_core.common.yaml import DataSourceYamlSource, YamlObject
+from soda_core.common.yaml import DataSourceYamlSource, VariableResolver, YamlObject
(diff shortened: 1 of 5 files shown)


m1n0 (approved):
cautiously approving

mivds (approved):
Would consider removing the `YamlSource.resolve` method, or replace it with `YamlSource.parse_and_resolve` so that we don't end up using it incorrectly again. We've hit this bug multiple times already, in different codebases. There is similar (but not quite same) logic in soda-library, soda-scan-launcher, soda-core and soda-contracts-launcher

paulteehan:
> Would consider removing the `YamlSource.resolve` method, or replace it with `YamlSource.parse_and_resolve` so that we don't end up using it incorrectly again. We've hit this bug multiple times already, in different codebases. There is similar (but not quite same) logic in soda-library, soda-scan-launcher, soda-core and soda-contracts-launcher

done, thanks

PR #2688, opened 2026-05-05
Fix: Athena dqn containing forward slash in catalog

Niels-b:
## Description

  - Athena DQNs may contain forward slashes
  - Maybe, our test suite should uncover such issues.

## Checklist

  - [ ] CI run: https://github.com/sodadata/soda-extensions/actions/runs/25368860205 and https://github.com/sodadata/soda-extensions/actions/runs/25440849371


m1n0:
A few suggestions, none blocking:

**1. Collapse the two normalization helpers into one.** `AthenaDataSourceImpl._normalize_athena_prefixes` (returns `(catalog, schema)`) and `AthenaSqlDialect._normalize_prefix` (returns `list[str]`) encode the same rule — "everything-but-last → catalog, last → schema" — in two places, with subtly different thresholds (`len > 2` vs. always-join). They're also inconsistent in shape because they feed different override seams in different base classes.

Worth unifying to a single module-level helper:

```python
def _collapse_athena_prefixes(prefixes: Optional[list[str]]) -> tuple[Optional[str], Optional[str]]:
    if not prefixes:
        return None, None
    if len(prefixes) == 1:
        return prefixes[0], None
    return "/".join(prefixes[:-1]), prefixes[-1]
```

…and have both the `extract_*_from_prefix` overrides and the dialect's qualify methods route through it (the dialect just needs a tiny list-shaped adapter on top). It also removes the `Class._method(None, …)` hack the unit tests use today, which is itself a tell that the method should be a free function or `@staticmethod`.

**2. Add a comment on the `extract_*_from_prefix` overrides** explaining *why* the index-based seam is bypassed — i.e. that the base relies on `sql_dialect.get_database_prefix_index()` / `get_schema_prefix_index()` (fixed indices), and Athena can't use that here because a slash-containing catalog spans an arbitrary number of post-parse prefix elements. Otherwise the next person to touch this will wonder why we're not using the index machinery the rest of the codebase uses.

**3. Suggest filing a follow-up ticket for the root-cause fix.** The post-parse recovery only works because Athena has a fixed 2-level structure; it doesn't generalize, and anyone iterating `DatasetIdentifier.prefixes` directly (outside of the overridden paths) silently gets the over-split form. The proper fix is to escape `/` (or pick a separator that can't appear in identifiers) in DQN segments at construction time so `DatasetIdentifier.parse()` recovers the right shape without connector-specific post-processing. Migration is non-trivial (persisted DQNs in Cloud, hashes via `to_hash()`), so it's its own piece of work — but worth tracking so this workaround doesn't quietly become the long-term design. Reference the ticket from the helper's docstring once filed.

Also unrelated: the `execute_update(..., log_query: bool = True)` change has nothing to do with the slash-catalog fix — could probably be split into its own PR.

m1n0 (approved):
approved with caveats, I am not a huge fan of forking and post-processing just for athena

Niels-b:
Thanks for the review @m1n0 . I agree, it's a bandaid fix for now. But we have to verify it first. Then we can think about fixing the root cause properly. I think we may have something similar for Dremio 🤔 .

PR #2689, opened 2026-05-06
feat: expose live session timezone on each DataSourceConnection

Niels-b:
Add a `get_session_timezone()` accessor (cached, with UTC fallback) on the base `DataSourceConnection`, plus per-adapter `_fetch_session_timezone()` overrides that issue the appropriate vendor SQL: SHOW timezone (postgres/redshift), SHOW PARAMETERS (snowflake), current_timezone() (databricks/trino), current_setting (duckdb), SYSDATETIMEOFFSET() (the SQL Server family — works on SQL Server proper as well as Fabric DW and Synapse where CURRENT_TIMEZONE_ID is unsupported). BigQuery, Athena, Dremio and SparkDF return UTC directly.

This is purely additive infrastructure for the diagnostics-warehouse value-mapper layer to canonicalize TIMESTAMP_TZ values across heterogeneous source/target pairs. No existing behavior changes — nothing reads from the new method yet.

A new integration test under soda-tests/tests/integration verifies for every adapter that (a) the call returns a usable tzinfo, (b) repeat calls are cached, (c) the live session TZ matches an EXPECTED_SESSION_TIMEZONES mapping (catches silent drift), and (d) the underlying SQL produces no warning/error log records.

## Description

## Checklist

- [ ] I added a test to verify the new functionality.

### DIFF
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -81,9 +81,7 @@ class DataSourceConnection(ABC):
             try:
                 self._session_timezone_cache = self._fetch_session_timezone()
             except Exception as e:
-                logger.warning(
-                    f"Failed to query session timezone on '{self.name}'; defaulting to UTC: {e}"
-                )
+                logger.warning(f"Failed to query session timezone on '{self.name}'; defaulting to UTC: {e}")
                 self._session_timezone_cache = timezone.utc
         return self._session_timezone_cache
 
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -3,9 +3,17 @@ from __future__ import annotations
 import contextlib
 import logging
 import os
+import re
 from abc import ABC, abstractmethod
 from collections.abc import Iterator
+from datetime import timedelta, timezone, tzinfo
 from typing import Any, Callable, Optional
+
+try:
+    from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+except ImportError:
+    ZoneInfo = None
+    ZoneInfoNotFoundError = Exception
 
 from soda_core.common.data_source_results import QueryResult, QueryResultIterator
 from soda_core.common.logging_constants import soda_logger
@@ -13,6 +21,35 @@ from soda_core.common.logging_constants import soda_logger
 logger: logging.Logger = soda_logger
 
 from tabulate import tabulate
+
+
+def parse_session_timezone(value: str) -> tzinfo:
+    """Parse a session-timezone string returned by a database driver into a tzinfo.
+
+    Accepts IANA names ('America/Los_Angeles'), 'UTC'/'GMT', or numeric offsets like
+    '+00:00', '-08:00', '+0530', 'UTC+02:00'. Falls back to UTC if unparseable.
+    """
+    if not value:
+        return timezone.utc
+    text = value.strip().strip("'\"")
+    if text.upper() in ("UTC", "GMT", "Z"):
+        return timezone.utc
+
+    if ZoneInfo is not None:
+        try:
+            return ZoneInfo(text)
+        except (ZoneInfoNotFoundError, ValueError, OSError):
+            pass
+
+    match = re.match(r"^(?:UTC|GMT)?([+-])(\d{1,2}):?(\d{2})?$", text)
+    if match:
+        sign = -1 if match.group(1) == "-" else 1
+        hours = int(match.group(2))
+        minutes = int(match.group(3) or 0)
+        return timezone(sign * timedelta(hours=hours, minutes=minutes))
+
+    logger.warning(f"Could not parse session timezone {text!r}; defaulting to UTC")
+    return timezone.utc
 
 
 class DataSourceConnection(ABC):
@@ -29,9 +66,29 @@ class DataSourceConnection(ABC):
         self.name: str = name
         self.connection_properties: dict = connection_properties
         self.connection: Optional[object] = connection
+        self._session_timezone_cache: Optional[tzinfo] = None
 
         # Auto-open on creation if no connection already supplied. See DataSource.open_connection()
         self.open_connection()
+
+    def get_session_timezone(self) -> tzinfo:
+        """Return the live session timezone of this connection.
+
+        Result is cached for the connection's lifetime. Subclasses override
+        :meth:`_fetch_session_timezone` to query their backend; the default is UTC.
+        """
+        if self._session_timezone_cache is None:
+            try:
+                self._session_timezone_cache = self._fetch_session_timezone()
+            except Exception as e:
+                logger.warning(
+                    f"Failed to query session timezone on '{self.name}'; defaulting to UTC: {e}"
+                )
+                self._session_timezone_cache = timezone.utc
+        return self._session_timezone_cache
+
+    def _fetch_session_timezone(self) -> tzinfo:
+        return timezone.utc
 
     def __enter__(self):
         pass
--- a/soda-core/src/soda_core/common/data_source_connection.py
+++ b/soda-core/src/soda_core/common/data_source_connection.py
@@ -23,16 +23,25 @@ logger: logging.Logger = soda_logger
 from tabulate import tabulate
 
 
-def parse_session_timezone(value: str) -> tzinfo:
+# IANA names that mean "UTC, no DST, no historical offsets". When ZoneInfo returns one
+# of these, we collapse to ``timezone.utc`` so adapter outputs are uniform regardless of
+# whether the driver reports the literal string ``"UTC"`` or the IANA alias ``"Etc/UTC"``.
+_UTC_ALIAS_ZONE_KEYS = frozenset({"UTC", "Etc/UTC", "Universal", "Zulu", "Etc/Zulu"})
+
+
+def parse_session_timezone(value: Optional[str]) -> tzinfo:
     """Parse a session-timezone string returned by a database driver into a tzinfo.
 
-    Accepts IANA names ('America/Los_Angeles'), 'UTC'/'GMT'/'Z', or numeric offsets like
-    '+00:00', '-08:00', '+0530', 'UTC+02:00'. Zero offsets ('+00:00', 'UTC+00:00', etc.)
-    are normalized to ``timezone.utc`` so adapter outputs are uniform across vendors
-    that report UTC by name vs by offset. No production code currently depends on this
-    by-identity (``is timezone.utc``) check — call it future-proofing for any
-    fast-path branches that may want to skip ``astimezone`` on a guaranteed-UTC tzinfo,
-    and treat it as cosmetic uniformity rather than a correctness fix.
+    Accepts IANA names ('America/Los_Angeles'), 'UTC'/'GMT'/'Z', UTC-equivalent IANA
+    aliases (``'Etc/UTC'``, ``'Universal'``, ``'Zulu'``, ``'Etc/Zulu'``), or numeric
+    offsets like '+00:00', '-08:00', '+0530', 'UTC+02:00'. Every UTC-equivalent input
+    — the literal strings, the alias zones, and zero numeric offsets — is normalized
+    to ``timezone.utc`` (the singleton) so adapter outputs are uniform across vendors
+    that report UTC by name, by alias, or by offset. No production code currently
+    depends on this by-identity (``is timezone.utc``) check; treat the normalization
+    as future-proofing for any fast-path branches that may want to skip ``astimezone``
+    on a guaranteed-UTC tzinfo, and as cosmetic uniformity rather than a correctness
+    fix.
 
     Empty / None / whitespace-only input is treated as "no session TZ configured" and
     returns ``timezone.utc`` silently. This is the conventional fallback every adapter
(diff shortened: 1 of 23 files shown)


soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
 
 from tabulate import tabulate
 
+# IANA names that mean "UTC, no DST, no historical offsets". When ZoneInfo returns one
+# of these, we collapse to ``timezone.utc`` so adapter outputs are uniform regardless of
+# whether the driver reports the literal string ``"UTC"`` or one of the IANA aliases
+# below. The Etc/GMT entries are tricky — note that Etc/GMT+0 / Etc/GMT-0 are both
+# zero-offset variants (Etc/GMT+N has *negative* offset N hours per POSIX inversion,
+# but +0/-0 collapse to UTC). GMT0 is a System V-style zero-offset zone alias.
+_UTC_ALIAS_ZONE_KEYS = frozenset(
```
paulteehan, line 26: 😮‍💨  wow this turned out to be finicky

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+    depends on this by-identity (``is timezone.utc``) check; treat the normalization
+    as future-proofing for any fast-path branches that may want to skip ``astimezone``
+    on a guaranteed-UTC tzinfo, and as cosmetic uniformity rather than a correctness
+    fix.
+
+    Empty / None / whitespace-only input is treated as "no session TZ configured" and
+    returns ``timezone.utc`` silently. This is the conventional fallback every adapter
+    uses on a missing-row query result; promoting it to an error would emit warnings on
+    every cold start.
+
```
paulteehan, line 70: this comment is a bit suspicious to me, I get the reasoning for not surfacing false positive warnings, but does this not open the possibility to false negatives, i.e. returning UTC when we should be throwing an error or a warning?  I don't know enough about the flow to meaningfully comment, just something to think about

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -29,10 +151,28 @@ def __init__(
...
+        if self._session_timezone_cache is None:
+            try:
+                self._session_timezone_cache = self._fetch_session_timezone()
+            except Exception as e:
+                logger.warning(f"Failed to query session timezone on '{self.name}'; defaulting to UTC: {e}")
+                self._session_timezone_cache = timezone.utc
+        return self._session_timezone_cache
+
+    def _fetch_session_timezone(self) -> tzinfo:
+        return timezone.utc
```
paulteehan: wondering if this should raise a NotImplementedError rather than silently defaulting to UTC

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
```diff
@@ -175,9 +175,17 @@ def get_soda_data_type_name_by_data_source_data_type_names(
...
-            "timestamp": SodaDataTypeName.TIMESTAMP,
+            # Native ``timestamp`` in Databricks is TZ-aware in semantics (stored as
+            # an instant, displayed in session TZ — see Databricks "Timestamp" docs).
+            # The truly-naive variant is ``timestamp_ntz`` (Databricks Runtime 13.3+).
+            # Map ``timestamp`` to TIMESTAMP_TZ so cross-source DWH dispatch picks the
+            # TZ-aware value mapper for Databricks-as-source, otherwise driver-returned
+            # tz-aware values would be sent through the naive mapper which strips
+            # tzinfo, leaving a session-local wallclock in the warehouse instead of
+            # the canonical UTC instant.
+            "timestamp": SodaDataTypeName.TIMESTAMP_TZ,
```
paulteehan, line 186: This was broken before, huh?   good thing you did this overhaul

soda-sqlserver/tests/unit/test_sqlserver_session_timezone.py
```diff
@@ -0,0 +1,93 @@
...
+    cursor.close.return_value = None
+
+    pyodbc_conn = MagicMock()
+    pyodbc_conn.cursor.return_value = cursor
+    instance.connection = pyodbc_conn
+    return instance, cursor
+
+
+class TestSqlServerSessionTimezoneNormalization:
+    def test_zero_offset_returns_timezone_utc_identity(self) -> None:
```
paulteehan, line 47: the complexity here is making we think we need several CI instances of each database, each one configured with different session timezones.  obviously that's not feasible but I wonder if we could do something like that with docker containers for local testing.  It's hard to be confident we've covered all the edge cases we'd find in the wild with our relatively limited population of test database instances.  Bigger question for future

soda-tests/tests/integration/test_session_timezone.py
```diff
@@ -0,0 +1,183 @@
...
+    # Snapshot replay returns ``timezone.utc`` (recorded when the CI Snowflake account
+    # was on UTC); live runs against the current CI account return
+    # ``America/Los_Angeles``. Both are correct values that can appear depending on
+    # whether snapshots are active for this run, so accept either. A future flip of
+    # the CI account would still need a deliberate update here — this is not a free
+    # pass for arbitrary drift.
+    "snowflake": (ZoneInfo("America/Los_Angeles"), timezone.utc),
+    "bigquery": timezone.utc,  # BigQuery is UTC-only by design
+    "athena": timezone.utc,  # Athena is UTC-only
+    "fabric": timezone.utc,  # Inherits SQL Server adapter; test instance is UTC
```
paulteehan, line 46: I think anything that depends on the test instance state should be passed in as an environment variable.  Fine to make it default  to the instance current value, just for better maintainability

paulteehan (approved):
Very nice, a tricky one indeed.  Left a few comments for you to consider, nothing blocking.

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
 from abc import ABC, abstractmethod
 from collections.abc import Iterator
+from datetime import datetime, timedelta, timezone, tzinfo
 from typing import Any, Callable, Optional
 
+try:
+    from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+except ImportError:
+    ZoneInfo = None
+    ZoneInfoNotFoundError = Exception
```
mivds: Why do we need `try: ... except ImportError` here? From what I find `zoneinfo` is part of the standard library and should be available since python 3.9?

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+    A2: junk like ``'Mars/Olympus'`` previously degraded to UTC with only a debug-level
+    log; now it surfaces clearly so adapter authors and operators see the problem). The
+    single legitimate fallback site is ``DataSourceConnection.get_session_timezone()``,
+    which catches and logs one warning before returning UTC.
+    """
+    if value is None or not str(value).strip():
+        return timezone.utc
+    # Strip outer whitespace, outer quotes, then inner whitespace. Drivers may quote
+    # their result and the body of those quotes may also be empty / whitespace-only.
+    text = value.strip().strip("'\"").strip()
```
mivds: Could be simplified as:
```suggestion
    text = value.strip(" \r\n\t'\"")
```

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: Ooh yes, this whole operation was 😅

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: We can probably make it more explicit to return a `NoTimezoneDefined` kind of thing, but that would still result in `UTC` being used further down the line. We're just moving the responsibility and complexity more upstream in that case.

It already throws an error when there is a timezone that we couldn't parse somehow (`ValueError(f"could not parse session timezone {text!r}")`).

The code is only used in the datasourceconnection in `_fetch_session_timezone`, which already handles such edge cases. The `parse_session_timezone` is a "helper" function

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: Hmm yes, I think that's a good idea. Especially when we add new datasources 👍

soda-sqlserver/tests/unit/test_sqlserver_session_timezone.py
Niels-b: For some datasources, yes this definitely should be possible with containers. But I think that's out of scope for this PR to set up

soda-tests/tests/integration/test_session_timezone.py
Niels-b: Not really a fan of splitting it up to env variables (we'd also have to do some more conversion since we're passing a `timezone.utc` now). I want it to fail "hard" and that we have to fix it in the code itself, rather than in a separate location in the env variables (thinking we may forget to do/update this).

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+    fix.
+
+    Empty / None / whitespace-only input is treated as "no session TZ configured" and
+    returns ``timezone.utc`` silently. This is the conventional fallback every adapter
+    uses on a missing-row query result; promoting it to an error would emit warnings on
+    every cold start.
+
+    Raises ``ValueError`` on non-empty unparseable input (the colleague-review concern
+    A2: junk like ``'Mars/Olympus'`` previously degraded to UTC with only a debug-level
+    log; now it surfaces clearly so adapter authors and operators see the problem). The
```
mivds: The stuff in parentheses feels like noise that doesn't belong in docstring

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+    single legitimate fallback site is ``DataSourceConnection.get_session_timezone()``,
+    which catches and logs one warning before returning UTC.
+    """
+    if value is None or not str(value).strip():
```
mivds: The `str(value).strip()` clause here seems to strictly be a duplicate of the clause below, no?

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+        return timezone.utc
+    # Strip outer whitespace, outer quotes, then inner whitespace. Drivers may quote
+    # their result and the body of those quotes may also be empty / whitespace-only.
+    text = value.strip().strip("'\"").strip()
+    if not text:
+        # Outer quotes wrapped an empty (or whitespace-only) string — same intent as
+        # a raw-empty input on the conventional empty-fallback path.
+        return timezone.utc
+    if text.upper() in ("UTC", "GMT", "Z"):
+        return timezone.utc
```
mivds, line 87: Why is this a separate case from the `ZoneInfo` parsing below?

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+
+    if text.lower() in _HOST_LOCAL_LITERAL_VALUES:
+        # Driver said "use the host local TZ". Resolve via the OS now. The result is a
+        # fixed-offset tzinfo derived from the current local time — DST drift across a
+        # long-lived connection is the same as for SQL Server / DB2 (out of scope per
+        # the team's "DWH connections are short-lived" stance), but the wallclock at
+        # canonicalization time is correct.
+        return datetime.now().astimezone().tzinfo
+
+    if ZoneInfo is not None:
```
mivds: Can be dropped if the import structure above is changed

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -3,17 +3,139 @@
...
+        try:
+            zi = ZoneInfo(text)
+        except (ZoneInfoNotFoundError, ValueError, OSError):
+            zi = None
+        if zi is not None:
+            # Collapse UTC-equivalent IANA aliases to ``timezone.utc`` so an adapter
+            # reporting ``"Etc/UTC"`` returns the same singleton as one reporting the
+            # literal ``"UTC"`` or the ``+00:00`` numeric offset. Without this, the
+            # docstring's identity-parity claim would be false for one specific input.
+            if zi.key in _UTC_ALIAS_ZONE_KEYS:
```
mivds: Could possibly be replaced by looking at the offset relative to UTC rather than keeping a hardcoded list of UTC-like timezones. Not sure it's better, just something to consider.


```suggestion
            if zi.utcoffset(datetime.now()).total_seconds() == 0:
```

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
```diff
@@ -28,6 +32,12 @@ def _create_connection(
             **config.to_connection_kwargs(),
         )
 
+    def _fetch_session_timezone(self) -> tzinfo:
+        with self.connection.cursor() as cursor:
+            cursor.execute("SELECT current_timezone()")
+            row = cursor.fetchone()
+        return parse_session_timezone(row[0] if row else "")
```
mivds: Nit, feel free to ignore: outcome will be the same, but we could be more explicit here:


```suggestion
        if not row:
            return datetime.timezone.utc
        return parse_session_timezone(row[0])
```

soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source_connection.py
```diff
@@ -145,3 +149,33 @@ def _create_connection(
             application="Soda",
             **config.to_connection_kwargs(),
         )
+
+    def _fetch_session_timezone(self) -> tzinfo:
+        cursor = self.connection.cursor()
+        try:
```
mivds: Any reason why we're not using `with self.connection.cursor()` here?

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
```diff
@@ -196,5 +196,31 @@ def _create_connection(
...
+    def _fetch_session_timezone(self) -> tzinfo:
+        # Use SYSDATETIMEOFFSET() instead of CURRENT_TIMEZONE_ID() so the same query works
+        # across the whole SQL Server family: SQL Server proper, Microsoft Fabric DW, Azure
+        # Synapse Dedicated/Serverless. CURRENT_TIMEZONE_ID is unsupported in Fabric and
+        # Synapse Dedicated; SYSDATETIMEOFFSET is universally supported.
+        # The connection registers ``handle_datetimeoffset`` as a pyodbc output converter for
+        # SQL_TYPE -155, so the returned value is a tz-aware ``datetime`` whose tzinfo is the
+        # exact offset reported by the server.
+        cursor = self.connection.cursor()
+        try:
```
mivds: Same question here around using `with cursor`

mivds (approved):
LGTM, nice work finally tackling our timezone issues at the root.

I left some comments for consideration, but there's nothing blocking there.

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: Yes, good catch! Also makes the comments below a bit simpler. Thank you.

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: Good catch! Only the `"Z"` one should be as an exception here, the rest should be handled normally by the timezones.

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: As discussed on Slack, this one we probably shouldn't do as it introduces a bug.

soda-snowflake/src/soda_snowflake/common/data_sources/snowflake_data_source_connection.py
Niels-b: with is always better; changing it. Thanks!

PR #2690, opened 2026-05-07
refactor(DES-315): extract CheckCollection* base classes for extensible check execution

m1n0:
## Summary

Structural refactor: extract a generic `CheckCollection*` base layer from the existing `Contract*` classes, and open a real extension seam (class-attribute polymorphism hooks) so future check-collection subtypes can plug into the same orchestration without forking it. **No change to the programmatic API for contract callers.**

The current contract verification engine is soda-core's most general check-execution machinery, but every class is named for the specific contract case (`ContractYaml`, `ContractImpl`, `ContractVerificationSession`, …). This PR promotes those classes to a `CheckCollection*` base layer in a new `soda_core.check_collections` subpackage, with `Contract*` remaining as concrete subtypes that the base orchestration constructs by default via four polymorphism hooks.

A contract becomes one specialization of "a YAML of checks executed against a dataset" — the most general primitive. The `Contract*` subtypes preserve every public import path, attribute access, constructor signature, and isinstance check that programmatic callers rely on. The one intentional exception is documented below.

## What this PR delivers

**1. Base layer** — new `soda_core.check_collections` subpackage:
| Module | Base class (canonical) | Concrete `Contract*` subtype (kept for BC) |
|---|---|---|
| `check_collections.check_collection_verification` | `CheckCollectionVerificationSession` | `ContractVerificationSession` |
| `check_collections.check_collection_verification` | `CheckCollectionSessionResult` | `ContractVerificationSessionResult` |
| `check_collections.check_collection_verification` | `CheckCollectionResult` | `ContractVerificationResult` |
| `check_collections.impl.check_collection_yaml` | `CheckCollectionYaml` | `ContractYaml` |
| `check_collections.impl.check_collection_yaml` | `CheckCollectionYamlExtension` (Protocol) | — (renamed from `ContractYamlExtension`, no BC alias — see "Intentional break" below) |
| `check_collections.impl.check_collection_verification_impl` | `CheckCollectionVerificationSessionImpl` | `ContractVerificationSessionImpl` |
| `check_collections.impl.check_collection_verification_impl` | `CheckCollectionImpl` | `ContractImpl` |
| `check_collections.impl.check_collection_verification_impl` | `CheckCollectionImplExtension` (Protocol) | — (renamed from `ContractImplExtension`, no BC alias — see "Intentional break" below) |
| `check_collections.impl.check_collection_verification_impl` | `CheckCollectionVerificationHandler` | `ContractVerificationHandler` (BC alias) |
| `check_collections.impl.check_collection_verification_impl` | `CheckCollectionVerificationHandlerRegistry` | `ContractVerificationHandlerRegistry` (BC alias) |
| `common.yaml` | `CheckCollectionYamlSource` | `ContractYamlSource` |

**2. Extension seam** — four class-attribute hooks on the base classes:
```python
class CheckCollectionVerificationSessionImpl:
    _YAML_CLASS: Optional[type[CheckCollectionYaml]] = None
    _IMPL_CLASS: Optional[type[CheckCollectionImpl]] = None
    _RESULT_CLASS: Optional[type] = None
    _SESSION_RESULT_CLASS: Optional[type] = None
```
The base orchestration resolves these via `cls._YAML_CLASS or _default_yaml_cls()` at every construction site (`execute()`, `_execute_locally()`, `_execute_on_agent()`, `verify()`). The `Contract*` path wires the hooks to itself at module load (`contract_verification_impl.py:142-147`, `contract_yaml.py:82`), so existing callers see identical behavior. A future `DataStandard*` subtype subclasses these and sets its own four hooks; orchestration then constructs its types with no other change. Proven by `soda-tests/tests/unit/test_check_collection_hooks.py` — two sentinel-subclass tests that exercise all four hooks end-to-end.

**3. BC bridges on the `Contract*` shim subclasses** — accept both legacy and canonical kwarg names, with `TypeError` if both are passed:
- `ContractVerificationSession.execute(contract_yaml_sources=...)` and `(check_collection_yaml_sources=...)` (`contract_verification.py:60`)
- `ContractVerificationSessionResult(contract_verification_results=...)` and `(check_collection_results=...)` (`contract_verification.py:75`)
- `ContractVerificationSessionImpl.execute(contract_yaml_sources=...)` and `(check_collection_yaml_sources=...)` (`contract_verification_impl.py:60`)
- `ContractImpl(contract_yaml=...)` and `(check_collection_yaml=...)` (`contract_verification_impl.py:99`)
- `ContractYaml.parse(contract_yaml_source=...)` and `(check_collection_yaml_source=...)` (`contract_yaml.py:28`)
- `ContractYaml(contract_yaml_source=...)` and `(check_collection_yaml_source=...)` (`contract_yaml.py:28`)

**4. `@property` aliases on `ContractYaml`** for the renamed instance attributes (`contract_yaml.py:66-74`):
- `contract_yaml.contract_yaml_source` → `check_collection_yaml_source`
- `contract_yaml.contract_yaml_object` → `check_collection_yaml_object`

**5. Other re-exports preserved** — `__all__` on both `contract_verification.py` and `contract_verification_impl.py` lists every legacy name; the shim re-imports them from the new base layer so `from soda_core.contracts.contract_verification import X` and `from soda_core.contracts.impl.contract_verification_impl import X` continue to work for all 47 legacy names (verified by introspection — see "Proof" below).

## Backwards compatibility

**Intent**: every public surface a programmatic user might rely on continues to work without code changes, *except* the two internal-only extension Protocols renamed for naming consistency (see below). The top-level user-facing API in `soda_core.contracts.api/` is untouched — zero diff.

**Mechanisms in place** (proof in the next section):

| Concern | Mechanism |
|---|---|
| Class imports from `contracts.contract_verification` | Shim re-exports all 16 legacy class names |
| Class imports from `contracts.impl.contract_verification_impl` | Shim re-exports all 31 legacy class names |
| `ContractVerificationSession.execute(contract_yaml_sources=...)` legacy kwarg | Accepted by BC bridge |
| `ContractYaml.parse(contract_yaml_source=...)` legacy kwarg | Accepted by BC bridge |
| `result.contract_verification_results` legacy attribute | `@property` alias |
| `contract_yaml.contract_yaml_object` / `.contract_yaml_source` legacy attributes | `@property` aliases |
| `ContractVerificationHandler` / `ContractVerificationHandlerRegistry` (post-processing handler extension surface) | Renamed canonically, BC alias kept |
| `contract_verification_handlers` classvar on the registry | BC alias references the canonical list by identity |
| Calling `execute()` with both legacy + canonical kwarg | Raises `TypeError` (loud, not silent precedence) |
| `isinstance(x, ContractVerificationResult)` etc. | Still works — `Contract*` is now a subclass of `CheckCollection*`, so isinstance against the legacy class succeeds for both legacy and new instances |

## Intentional break (the one exception, documented)

Two Protocol classes are renamed **without** a BC alias:

| Removed | Replacement |
|---|---|
| `ContractYamlExtension` in `soda_core.contracts.impl.contract_yaml` | `CheckCollectionYamlExtension` |
| `ContractImplExtension` in `soda_core.contracts.impl.contract_verification_impl` | `CheckCollectionImplExtension` |

These are internal extension hooks used to plug custom YAML / impl extensions into the engine. The only known consumers are `soda-extensions/soda-reconciliation` (updated in the companion PR — see below) and any external customer who built their own such Protocol implementation against an older soda-core. The break is **loud at import time** (`ImportError`), not silent. The migration is one line per implementation: change the import to the new name. No call-site changes required.

This was the deliberate naming-cleanup goal; keeping the old Protocol names would have meant carrying contract-flavoured names on a base-layer Protocol forever.

## Proof of equivalence

**Test matrix** (`TEST_DATASOURCE=duckdb pytest soda-tests/`):
- Pre-refactor baseline (merge-base `2dfcff27`): 811 passed, 14 skipped
- After all refactor commits: 813 passed, 14 skipped (+2 new tests for the polymorphism hooks)
- After merging current `main` (which added unrelated tests for SCS-1172): **900 passed, 14 skipped** ✅

**CI matrix** (most recent push): all 12 non-pending connectors green on Python 3.14 (athena, bigquery, duckdb, fabric, postgres, redshift, snowflake, sparkdf, sqlserver, synapse, trino-postgres, trino-s3). Pre-commit / lockfile / Aikido / SonarCloud / CLA all green.

**Public import audit** (introspection on the merged branch — 47 legacy class names across three public modules):
```
Checked 47 legacy imports — 0 missing.
Intentional renames (documented):
  contract_yaml.ContractYamlExtension: removed (renamed)
  contract_verification_impl.ContractImplExtension: removed (renamed)
```

**Top-level user-facing API** (`soda_core/contracts/api/`): **zero diff** between merge-base and HEAD. Anyone calling `verify_contract*` or `ContractVerificationSession.execute(...)` sees no difference.

**Hook substitutability proven**: `soda-tests/tests/unit/test_check_collection_hooks.py` defines five sentinel subclasses (one per base) and asserts that wiring the four hooks routes orchestration through them — proving the extension seam is real, not just renamed.

## Companion PR

[sodadata/soda-extensions#370](https://github.com/sodadata/soda-extensions/pull/370) — `des-315-phase-1-soda-core-structural-refactor-checkcollection-base` branch — updates the known consumers of the renamed Protocols and the renamed attribute names:

- `soda-reconciliation` — imports the new Protocol names; renames extension-method parameter names to match the new Protocol signatures
- `soda-migration` — uses the renamed `check_collection_yaml_source` / `check_collection_yaml_object` attributes directly (rather than relying on the BC `@property` aliases)
- `soda/failed_rows_extractor` — same attribute rename

The companion PR depends on this PR landing first, then a new soda-core release publishing the renames, then a lockfile bump in soda-extensions. Documented on that PR.

Refs DES-315


paulteehan (approved):

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
```diff
@@ -0,0 +1,1978 @@
...
+        only_validate_without_execute: bool,
+        provided_variable_values: dict[str, str],
+        data_timestamp: Optional[str],
+        data_source_impls: list[DataSourceImpl],
+        data_source_yaml_sources: list[DataSourceYamlSource],
+        soda_cloud_impl: Optional[SodaCloud],
+        soda_cloud_publish_results: bool,
+        check_selectors: list[CheckSelector],
+        dwh_data_source_file_path: Optional[str] = None,
+    ) -> list["ContractVerificationResult"]:
```
mivds: Should this also be renamed? Still specifies `ContractVerificationResult`

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
```diff
@@ -0,0 +1,1978 @@
...
+    # Subtype hooks: subclasses (e.g. ContractVerificationSessionImpl, a future
+    # DataStandardVerificationSessionImpl) set these to their concrete YAML / impl /
+    # result / session-result classes. Resolved at use sites in execute() and its
+    # helpers; ``None`` falls back to the default Contract* types so the base layer
+    # remains usable directly. This makes the structural seam polymorphic — a
+    # subclass that sets the four hooks gets its own types constructed and returned
+    # without re-implementing orchestration.
+    _YAML_CLASS: Optional[type["CheckCollectionYaml"]] = None
+    _IMPL_CLASS: Optional[type["CheckCollectionImpl"]] = None
+    _SESSION_RESULT_CLASS: Optional[type] = None
```
mivds, line 168: If the goal is to enable polymorphism and proper typing, I think this implementation is incomplete. The return types for this class are still referencing the base class, not the actual implementation classes. To fix that the class should be made generic.

```python
from typing import Generic, TypeVar

CheckCollectionYamlType = TypeVar("CheckCollectionYamlType", bound=CheckCollectionYaml)
CheckCollectionImplType = TypeVar("CheckCollectionImplType", bound=CheckCollectionImpl)
CheckCollectionVerificationResultType = TypeVar("CheckCollectionVerificationResultType", bound=ContractVerificationResult)  # This one probably still needs to renamed?

class CheckCollectionVerificationSessionImpl(Generic[CheckCollectionYamlType, CheckCollectionImplType, CheckCollectionVerificationResultType]):
    @classmethod
    def _execute_locally(...): -> list[CheckCollectionVerificationResultType]
        ...
```

I'd also argue using subclass parameter initialization is cleaner (more explicit) than overriding class variables the way it is now, but that's more of a flavor thing. Example of what that could look like:

```python
class CheckCollectionVerificationSessionImpl:
    def __init_subclass__(
        cls,
        check_collection_yaml_type: type[CheckCollectionYamlType] = CheckCollectionYaml,
        check_collection_impl_type: type[CheckCollectionImplType] = CheckCollectionImpl,
        check_collection_verificaton_result_type: type[CheckCollectionVerificationResultType] = CheckCollectionVerificationResult
        **kwargs
    ):
        super().__init_subclass__(**kwargs)
        cls._YAML_CLASS = check_collection_yaml_type
        cls._IMPL_CLASS = check_collection_impl_type
        cls._RESULT_CLASS = check_collection_verificaton_result_type

class ContractVerificationSessionImpl(
    check_collection_yaml_type=ContractYaml,
    check_collection_impl_type=ContractImpl,
    check_collection_verificaton_result_type=ContractVerificationResult
):
    ...
```

Of if we combine both and extract the type arguments from the `Generic` typing (more magic here, less work in inherited classes):
```python
from typing import Generic, TypeVar, get_args

CheckCollectionYamlType = TypeVar("CheckCollectionYamlType", bound=CheckCollectionYaml)
CheckCollectionImplType = TypeVar("CheckCollectionImplType", bound=CheckCollectionImpl)
CheckCollectionVerificationResultType = TypeVar("CheckCollectionVerificationResultType", bound=ContractVerificationResult)  # This one probably still needs to renamed?

class CheckCollectionVerificationSessionImpl(Generic[CheckCollectionYamlType, CheckCollectionImplType, CheckCollectionVerificationResultType]):
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        # Note: probably needs more logic to deal with multiple/nested inheritance
        cls._YAML_CLASS, cls._IMPL_CLASS, cls._RESULT_CLASS = get_args(cls.__orig_bases__)
```

Disclaimer:
* Not validated any of it, mostly going off of memory and some quick interactive terminal trials
* That last implementation with extracting types from the typing info may need to resolve strings to actual types (using `typing.get_type_hints`)

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
```diff
@@ -0,0 +1,1978 @@
...
+    SodaCloudYamlSource,
+)
+from soda_core.contracts.impl.check_selector import CheckSelector
+from tabulate import tabulate
+
+logger: logging.Logger = soda_logger
+
+
+def _default_yaml_cls():
+    """Lazy-resolve the default concrete CheckCollectionYaml subclass."""
```
mivds, line 62: What's the rationale for doing lazy resolving? Are we trying to avoid import conflicts, or doing this for performance reasons?

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
```diff
@@ -0,0 +1,1978 @@
...
+
+
+class CheckCollectionImpl:
+    check_collection_impl_extensions: dict[str, type[CheckCollectionImplExtension]] = {}
+
+    # Subtype hook: subclasses (e.g. ContractImpl, a future DataStandardImpl) set this
+    # at module level to the concrete result class they want ``verify()`` to construct
+    # and return. ``None`` falls back to ContractVerificationResult for backwards
+    # compatibility.
+    _RESULT_CLASS: Optional[type] = None
```
mivds, line 438: Same note here as above. `CheckCollectionImpl` should probably become `Generic`

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
```diff
@@ -0,0 +1,1978 @@
...
+    def _parse_columns(self, check_collection_yaml: CheckCollectionYaml) -> list[ColumnImpl]:
+        columns: list[ColumnImpl] = []
+        if check_collection_yaml.columns:
+            for column_yaml in check_collection_yaml.columns:
+                if column_yaml:
+                    column = ColumnImpl(check_collection_impl=self, column_yaml=column_yaml)
+                    columns.append(column)
+        return columns
+
+    def verify(self) -> "ContractVerificationResult":
```
mivds: Any specific reason to quote the result type here (and elsewhere)? I'd be inclined to go with `from __future__ import annotations` at the top of the file instead

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
m1n0: yes, this was intentionally incomplete/minimal, just to enable moving all the code around, as all variants require some changes in the callers, which I very strictly want to avoid in this PR. Subsequent PR coming right after this will do the change, and I have previously chosen subclass parameter initialization.

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
m1n0: indeed, and there were a few more, renamed

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
m1n0: Only circular import avoidance. These will disappear in subsequent PR once init subclass is done

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
m1n0: indeed, sublcass init will solve this fully

soda-core/src/soda_core/check_collections/impl/check_collection_verification_impl.py
m1n0: leftover, it was needed previously because of circular imports, gone now. I was too strict in speccing this phase one with things like "no changes to code whatsoever" etc, as I was trying to minimize risk of drift. Resolved

mivds (approved):

Issue #373, opened 2021-07-02
Inconsistent handling of the arguments for dialects' `sql_tables_metadata_query`


milanaleksic:
**Describe the bug**

Postgres, Hive, Athena, SqlServer dialects don't enforce the `limit` and `filter` like Snowflake does in the call `sql_tables_metadata_query`. We should align across all dialects the usage of the `LIMIT`, especially since it is by default set. Alternatively we can just remove the parameter altogether from all dialects?

Context: 
- [only Snowflake into consideration the arguments](https://github.com/sodadata/soda-sql/blob/27c9e226a73c4034a703e3ab7e03b65f7db274d4/packages/snowflake/sodasql/dialects/snowflake_dialect.py#L121)
- [example of postgres that ignores both arguments](https://github.com/sodadata/soda-sql/blob/27c9e226a73c4034a703e3ab7e03b65f7db274d4/packages/postgresql/sodasql/dialects/postgres_dialect.py#L49)

vijaykiran:
Related: #398

PR #2693, opened 2026-05-11
Chore: enable cross source testing for all data sources 

Niels-b:
## Summary

Five narrow dialect / test-helper fixes that the expanded cross-source DWH matrix in soda-extensions surfaces. Each fix is scoped to a single dialect or shared helper — no cross-cutting refactors.

### 1. Generic precision-strip in base CREATE TABLE rendering
Postgres source extension populates both `numeric_precision` and `datetime_precision` from psycopg's polymorphic `column.precision`, which leaks `numeric_precision` onto timestamp columns. After the dialect-level `datetime_precision` strip on athena / oracle / redshift, `SqlDataType` rendering fell through to the still-set `numeric_precision` branch and emitted `timestamp(2)` — rejected by all three.

`_build_create_table_column_type` in the base layer now strips `numeric_precision` from non-numeric columns and `datetime_precision` from non-datetime columns *before* dialect overrides run (via the existing `data_type_has_parameter_*` helpers). Athena's override delegates to `super()` so the safeguards apply.

Fixes cross-source DWH failed_rows_query on postgres → athena / oracle / redshift.

### 2. Snowflake VARCHAR / TEXT normalisation
Snowflake reports unbounded text columns with `CHARACTER_MAXIMUM_LENGTH=16777216` (its native max). The dialect now normalises that to a length-less canonical TEXT, so downstream targets don't choke on a literal `VARCHAR(16777216)` they can't represent.

### 3. BigQuery NUMERIC / BIGNUMERIC precision & scale
BigQuery's `INFORMATION_SCHEMA` doesn't expose precision/scale for NUMERIC / BIGNUMERIC columns. Override `extract_numeric_precision` / `extract_numeric_scale` to return the documented fixed values `(38,9)` and `(76,38)`. Without this, every fractional value was silently truncated to scale 0 on cross-source moves.

### 4. SQL Server / Synapse / Fabric datetime precision clamp
SQL Server's `datetime2` / `time` / `datetimeoffset` accept precision 0..7 (Fabric's max is 6). Cross-source flows from higher-precision sources (e.g. Snowflake `TIMESTAMP_NTZ` defaults to 9) used to emit e.g. `datetime2(9)` and fail CREATE TABLE. Clamped at the dialect level; Synapse inherits via SQL Server.

### 5. `test_drop_old_schemas` fallback
Strict fallback for the new cross-source schema-name shape `soda_diagnostics_<datasource>_<YYYYMMDD>_<hash>`, which broke the fixed-offset date extraction in the old parser. Only triggers when the schema has exactly 5 underscore-parts and parts[3] is 8 digits; existing schema names parse identically.

## Description

- **What problem are you solving?**
  Unblocking the full 14-source cross-source DWH matrix (full cartesian product) that soda-extensions PR #373 enables in CI. The matrix exposed dialect mismatches that the previous hand-picked compatibility list never exercised.

- **Any expected impact on downstream packages/services?**
  Yes — paired with soda-extensions PR #373, which contains the matrix expansion + CI plumbing + a couple of soda-extensions-specific dialect fixes (BigQuery cursor description, DB2 decimal clamp).

- **Reference issue:** R-581.

## Verification

Local validation: 44 previously-failing cross-source test cases now pass across 11 source/target combinations; 44 regression test cases unaffected. CI on soda-extensions PR #373 shows 182/182 cross-source cells green (14 sources × 13 targets, duckdb excluded as DWH target by design).

## Checklist

  - This work is *for* soda-extensions PR #373.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -190,6 +190,13 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
     # precision/scale) carry the correct decimal scale through; otherwise a
     # `numeric` column on the target defaults to scale 0 and silently truncates
     # the fractional part of every value.
+    #
+    # Note: these overrides intentionally bypass the base layer's
+    # `data_type_has_parameter_numeric_*` / `supports_data_type_numeric_*` gates
+    # (both return False for BigQuery in the CREATE TABLE path). The asymmetry
+    # is source-vs-target: BigQuery-as-source needs to *carry* the precision so
+    # downstream target dialects that do require it can render correctly, even
+    # though BigQuery-as-target would never render it itself.
     _BIGQUERY_NUMERIC_PRECISION = 38
     _BIGQUERY_NUMERIC_SCALE = 9
     _BIGQUERY_BIGNUMERIC_PRECISION = 76
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -183,6 +183,34 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
     def supports_data_type_datetime_precision(self) -> bool:
         return False
 
+    # BigQuery's NUMERIC / BIGNUMERIC are fixed-precision (no per-column tuning),
+    # so INFORMATION_SCHEMA.COLUMNS doesn't expose NUMERIC_PRECISION / NUMERIC_SCALE.
+    # Hard-code BigQuery's documented values so consumers of the column metadata
+    # (e.g. cross-source CREATE TABLE on dialects that *do* require explicit
+    # precision/scale) carry the correct decimal scale through; otherwise a
+    # `numeric` column on the target defaults to scale 0 and silently truncates
+    # the fractional part of every value.
+    _BIGQUERY_NUMERIC_PRECISION = 38
+    _BIGQUERY_NUMERIC_SCALE = 9
+    _BIGQUERY_BIGNUMERIC_PRECISION = 76
+    _BIGQUERY_BIGNUMERIC_SCALE = 38
+
+    def extract_numeric_precision(self, row, columns) -> Optional[int]:
+        data_type_name: str = self.extract_data_type_name(row, columns).lower()
+        if data_type_name == "numeric":
+            return self._BIGQUERY_NUMERIC_PRECISION
+        if data_type_name == "bignumeric":
+            return self._BIGQUERY_BIGNUMERIC_PRECISION
+        return super().extract_numeric_precision(row, columns)
+
+    def extract_numeric_scale(self, row, columns) -> Optional[int]:
+        data_type_name: str = self.extract_data_type_name(row, columns).lower()
+        if data_type_name == "numeric":
+            return self._BIGQUERY_NUMERIC_SCALE
+        if data_type_name == "bignumeric":
+            return self._BIGQUERY_BIGNUMERIC_SCALE
+        return super().extract_numeric_scale(row, columns)
+
     def sql_expr_timestamp_literal(self, datetime_in_iso8601: str) -> str:
         return f"timestamp('{datetime_in_iso8601}')"
 
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -465,6 +465,23 @@ class SqlDialect:
     def _build_create_table_column_type(self, create_table_column: CREATE_TABLE_COLUMN) -> str:
         assert isinstance(create_table_column.type, SqlDataType)
 
+        # Strip precision/length fields that don't apply to this column's data type.
+        # Some source extensions populate multiple precision fields from a single
+        # cursor-description value (e.g. postgres sets numeric_precision AND
+        # datetime_precision from column.precision); without this filter,
+        # get_sql_data_type_str_with_parameters() picks the first non-None and would
+        # emit e.g. `timestamp(2)` from a leaked numeric_precision even after a
+        # subsequent dialect-level safeguard zeros datetime_precision.
+        type_name: str = create_table_column.type.name
+        if not self.data_type_has_parameter_character_maximum_length(type_name):
+            create_table_column.type.character_maximum_length = None
+        if not self.data_type_has_parameter_numeric_precision(type_name):
+            create_table_column.type.numeric_precision = None
+        if not self.data_type_has_parameter_numeric_scale(type_name):
+            create_table_column.type.numeric_scale = None
+        if not self.data_type_has_parameter_datetime_precision(type_name):
+            create_table_column.type.datetime_precision = None
+
         if not self.supports_data_type_character_maximum_length():
             create_table_column.type.character_maximum_length = None
         if not self.supports_data_type_numeric_precision():
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import logging
 from abc import abstractmethod
 from datetime import date, datetime, time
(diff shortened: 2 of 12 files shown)


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -465,6 +465,23 @@ def _build_create_table_column(self, create_table_column: CREATE_TABLE_COLUMN) -
     def _build_create_table_column_type(self, create_table_column: CREATE_TABLE_COLUMN) -> str:
```
m1n0, line 466: hmm this feels a bit too dwh specific, actually cross-dwh specific, it doesn't look like a generic builder for create table statements 🤔

m1n0 (commented):
(superseded — see inline comments)

soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source.py
```diff
@@ -76,8 +76,17 @@ def _build_create_table_column_type(self, create_table_column: CREATE_TABLE_COLU
...
+        # Cross-source flows from sources with higher native precision (e.g. Snowflake's
+        # TIMESTAMP_NTZ defaults to 9) would otherwise produce e.g. `datetime2(9)` and
+        # fail CREATE TABLE with "precision value between 0 and 6 must be specified".
+        if create_table_column.type.name.lower() in ("datetime2", "datetimeoffset", "time"):
+            if (
+                create_table_column.type.datetime_precision is not None
+                and create_table_column.type.datetime_precision > 6
+            ):
+                create_table_column.type.datetime_precision = 6
+        return create_table_column.type.get_sql_data_type_str_with_parameters()
```
m1n0: Bypasses the new base-dialect safeguards. Mirror Athena/SQL Server's post-PR shape: `return super()._build_create_table_column_type(create_table_column)`. Otherwise the new strip pass in `sql_dialect._build_create_table_column_type` and Fabric's own `supports_data_type_*` strip don't run for Fabric, so e.g. a Postgres-source `timestamp(2)` leaked via `numeric_precision` still escapes to a Fabric DWH `CREATE TABLE`.

soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source.py
```diff
@@ -76,8 +76,17 @@ def _build_create_table_column_type(self, create_table_column: CREATE_TABLE_COLU
...
+        # Clamp datetime precision to Fabric's max (0..6, lower than SQL Server's 0..7).
+        # Cross-source flows from sources with higher native precision (e.g. Snowflake's
+        # TIMESTAMP_NTZ defaults to 9) would otherwise produce e.g. `datetime2(9)` and
+        # fail CREATE TABLE with "precision value between 0 and 6 must be specified".
+        if create_table_column.type.name.lower() in ("datetime2", "datetimeoffset", "time"):
+            if (
+                create_table_column.type.datetime_precision is not None
+                and create_table_column.type.datetime_precision > 6
+            ):
+                create_table_column.type.datetime_precision = 6
```
m1n0, line 91: Cosmetic: SQL Server uses a named constant (`_MAX_DATETIME_PRECISION = 7`) for the same clamp pattern. Consider `_MAX_DATETIME_PRECISION = 6` on the Fabric class for symmetry.

soda-tests/tests/integration/test_drop_old_schema.py
```diff
@@ -41,6 +41,16 @@ def determine_if_schema_needs_to_be_dropped(schema_name: str) -> bool:
...
             must_have_date = True
+            # Fallback for the cross-source variant: soda_diagnostics_<datasource>_<YYYYMMDD>_<8hex>
+            # Variable-length datasource names shift the date off the fixed offset above,
+            # so the substring above won't be a date. Only kick in when the fixed-offset
+            # extraction didn't produce an 8-digit string AND the name has the exact
+            # 5-underscore-part shape with parts[3] being 8 digits — otherwise behavior
+            # for all other schema names is unchanged.
+            if not (len(potential_date_string) == 8 and potential_date_string.isdigit()):
+                parts = schema_name.split("_")
+                if len(parts) == 5 and len(parts[3]) == 8 and parts[3].isdigit():
```
m1n0, line 52: `len(parts) == 5` rejects xdist-suffixed schemas. soda-extensions `dwh_setup.get_cross_datasource_dwh_schema_name` appends `_<xdist_worker>` (e.g. `_gw3`), so `soda_diagnostics_postgres_20251119_abcd1234_gw3` is 6 parts and falls through. Contradicts the PR body's "exactly 5 underscore-parts" claim.

```python
if len(parts) >= 5:
    for p in parts[3:]:
        if len(p) == 8 and p.isdigit():
            potential_date_string = p
            break
```

Low blast radius (safety-net code path), but worth fixing so xdist orphans aren't immortal.

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -465,6 +465,23 @@ def _build_create_table_column(self, create_table_column: CREATE_TABLE_COLUMN) -
...
         assert isinstance(create_table_column.type, SqlDataType)
 
+        # Strip precision/length fields that don't apply to this column's data type.
```
m1n0, line 476: Two notes:

1. **In-place mutation amplified.** The existing block at lines 485-490 already mutated `create_table_column.type` in place; this new strip adds four more. Cross-source flows that reuse a single `SqlDataType` across two dialects would see fields already stripped by the first render. Pre-existing pattern, not introduced here — but worth considering a defensive `replace(...)` copy at the top of the method now that the surface has doubled.

2. **No unit test pins this strip.** PR checklist "added a test" is ticked, but `rg '_build_create_table_column_type' soda-core/soda-tests/` returns nothing. A small parametrized test (build a `CREATE_TABLE_COLUMN` with `name='timestamp', numeric_precision=2` → assert no `(2)` in the SQL) would catch regressions on every commit rather than only in the matrix.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -183,6 +183,34 @@ def supports_data_type_numeric_scale(self) -> bool:
...
+    # precision/scale) carry the correct decimal scale through; otherwise a
+    # `numeric` column on the target defaults to scale 0 and silently truncates
+    # the fractional part of every value.
+    _BIGQUERY_NUMERIC_PRECISION = 38
+    _BIGQUERY_NUMERIC_SCALE = 9
+    _BIGQUERY_BIGNUMERIC_PRECISION = 76
+    _BIGQUERY_BIGNUMERIC_SCALE = 38
+
+    def extract_numeric_precision(self, row, columns) -> Optional[int]:
+        data_type_name: str = self.extract_data_type_name(row, columns).lower()
```
m1n0, line 206: Worth a one-liner that this intentionally bypasses base's `data_type_has_parameter_numeric_precision` + `supports_data_type_numeric_precision` gates (both return `False` for BigQuery — that's the source-vs-target asymmetry: source-side metadata needs to carry precision so cross-source *target* dialects render it). Without the comment, the next reader is going to "fix" this by adding the gate back. No behaviour change requested.

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: I disagree; this is cleaning/preparing of the column type for the create table statement. It should be able to handle non-correct input, which it didn't do correctly before.

m1n0: ok fair

soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source.py
Niels-b: That was done explicitly to not change the behavior of Fabric. But it is definitely worth doing this indeed.

soda-fabric/src/soda_fabric/common/data_sources/fabric_data_source.py
Niels-b: Fabric has a max precision of `6`. So we can't change it to 7

soda-tests/tests/integration/test_drop_old_schema.py
Niels-b: Discrepancy between PR body and this. There is parts>=5 which will cover the case for xdist. This is fine.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: Yes, will add docs for this

m1n0 (approved):

PR #2694, opened 2026-05-11
Chore: Intercept raw DBAPI cursors in snapshot wrapper

Niels-b:
Testing code that bypasses the high-level execute_query family — notably rows_diff in soda-reconciliation, which does data_source.connection.connection.cursor() and uses fetchmany() in a paginated merge-join — was incompatible with the snapshot record/replay system because SnapshotDataSourceConnection.connection was an opaque sentinel. This change records and replays that path without requiring any production code change.

snapshot_connection.py:
- _SnapshotDbapiProxy replaces the sentinel on self.connection; only cursor() is exposed, all other attributes raise AttributeError so existing fallback paths (e.g. _optimized_insert -> AST_INSERT) keep working unchanged.
- _SnapshotCursor records/replays at the execute() boundary. Narrow surface: execute(sql) (no params), fetchmany(size), description, close(). Eager-drain on execute() — fetchmany serves from the in-memory buffer.
- New op_type "cursor_execute"; handled by _activate_fallback as read-only (no DB state to re-create on fallback re-setup).
- primary_snapshot delegation: secondary wrappers route _handle_test_boundary / _record_entry / _next_replay_entry through the primary so interleaved cursor executes (rows_diff merge-join) share one ordered snapshot stream.

data_source_test_helper.py:
- Reference-counted monkey-patch of DataSourceImpl.create_additional_connection. When the patched method is invoked and the data source's primary connection is a SnapshotDataSourceConnection, the returned additional DataSourceConnection is wrapped in a secondary SnapshotDataSourceConnection linked to the primary. Required for the within-DB rows_diff path.

## Description

  - Recon check not being compatible with snapshot recordings
  - No, testing in soda-extensions as well

## Checklist

- [ ] I added a test to verify the new functionality.
  - PR will be created


m1n0:
Ok this is super fancy, thanks for this work Niels. It did make me think - wouldn't it be easier to adjust recon checks to be compatible with the existing system? 🤔

soda-tests/src/helpers/snapshot_manager.py
```diff
@@ -36,18 +36,41 @@ def _serialize_tzinfo(tz: tzinfo) -> str:
...
+    real DB can be materialised to the matching state — no more, no less —
+    when the upstream falls back mid-test. ``seq=None`` on a loaded entry
+    means it predates the seq mechanism; cascade falls back to the
+    coarser "only-if-consumed" heuristic in that case.
+    """
 
     op_type: str  # "query", "update", "query_one_by_one", "query_iterate"
     sql: str
     result: Any = None  # QueryResult, (description, rows), etc.
+    seq: Optional[int] = None
```
m1n0: **H1 — `seq` is dead but documented.** I grepped every `SnapshotEntry(...)` constructor in this PR (10+ sites in `snapshot_connection.py` + tests) and `seq` is never set; it always inherits `None`. The docstring above describes a "precise seq-based cascade for snapshot fallback," but that cascade was removed when `passthrough/fallback` was replaced with the swap-based rerun (commit `465e794e`). The sidecar at line 115 still writes `"seq": entry.seq` → every recorded JSON contains `"seq": null` per entry forever.

Either delete the field + serialization + docstring, or wire it up. Leaving it is misleading for whoever next extends this — they'll assume the cascade is live and build around it.

soda-tests/src/helpers/snapshot_connection.py
```diff
@@ -44,6 +45,88 @@
...
+    """
+    _RERUN_IN_PROGRESS.add(test_id)
+
+
+def end_rerun_in_progress(test_id: str) -> None:
+    """Plugin hook: clear the rerun-in-progress marker for ``test_id``."""
+    _RERUN_IN_PROGRESS.discard(test_id)
+
+
+def get_rerun_in_progress_record() -> set[str]:
```
m1n0: **H2 — name/signature say "membership," callers do truthiness.** The function returns the global set, callers at `data_source_test_helper.py:86` and `:537` do `if get_rerun_in_progress_record():` — i.e. "is *any* test in rerun?", not "is *this* test in rerun?". It works today because the plugin only ever sets one entry, but:

- The name and the return type both suggest membership-by-test_id.
- Leaking the internal mutable container lets a future caller `.add(...)` to it by accident.
- If any path opens a helper from inside a primary test body during a future rerun, the bypass would fire for an unrelated path.

Two cleaner shapes:
```python
def is_any_rerun_in_progress() -> bool: ...
# or, better:
def is_rerun_in_progress(test_id: str) -> bool: ...
```
The second takes `PYTEST_CURRENT_TEST` from the caller and is robust to future cross-test scenarios.

soda-tests/src/helpers/snapshot_connection.py
```diff
@@ -497,10 +737,19 @@ def _finalize_current_test(self) -> None:
...
         if unconsumed_error:
+            # Unconsumed entries at teardown mean an earlier mid-test mismatch was
+            # swallowed by user code (e.g. the contract verification handler).
+            # Queue the rerun signal and log a warning rather than raising —
+            # raising here would block the plugin from running the rerun.
+            # F9: only downgrade when the rerun plugin is actually wired up.
+            # Without the plugin, _PENDING_RERUN is read by nobody and the
+            # original test failure would silently disappear.
+            if self._allow_fallback and self._current_test_id is not None and is_pytest_plugin_active():
+                _PENDING_RERUN.setdefault(self._current_test_id, str(unconsumed_error))
```
m1n0, line 786: **H3 — silent discard of unconsumed-entry errors on rerun-mode test boundaries.** When `_finalize_current_test` is hit from `_handle_test_boundary` (the *next* test's first `execute_query`), the queue entry is keyed to the *previous* test's id — but the previous test's reports have already been emitted. The plugin's `_pop_rerun_reason(current_nodeid)` then can't find the entry, and `_reset_per_test_rerun_state` (line 223 in the plugin) clears the queue silently on the next iteration. The author's commit `82d021e5` confirms this is intentional ("don't fail teardown on unconsumed snapshot under rerun mode").

The trade-off is reasonable, but right now the only telemetry is a single `logger.warning(...)` per discard. `pytest_terminal_summary` enumerates `_RERAN_TEST_RECORD` but discarded queue entries don't appear anywhere in the summary. For a system whose explicit purpose is to **not** silently drop SQL drift, please surface a count: e.g. `_DISCARDED_UNCONSUMED_RECORD: list[tuple[test_id, reason]]` printed alongside the rerun summary. Otherwise this is exactly the kind of soft failure that disappears across many runs.

soda-tests/src/helpers/snapshot_connection.py
```diff
@@ -969,3 +1085,62 @@ def format_rows(self, rows: list[tuple]) -> list[tuple]:
         if self._real is not None:
             return self._real.format_rows(rows)
         return rows
+
+    # -------------------------------------------------------------------------
+    # Swap hooks (used by snapshot_pytest_plugin between attempts)
+    # -------------------------------------------------------------------------
+
+    def swap_to_real_for_rerun(self) -> bool:
```
m1n0, line 1137: **H6 — `_current_test_id` is not reset in `swap_to_real_for_rerun`.** `swap_to_real_for_rerun` clears `_replay_data` / `_replay_index` but leaves `_current_test_id` set. `restore_from_rerun` (~line 1145) eventually clears it on the way back. In between, anyone who holds a reference to the wrapper (the `_LIVE_SNAPSHOT_WRAPPERS` registry, or any `__getattr__` lookup) observes the wrapper with a `_current_test_id` that matches a test in fact running in off-mode against the real DB.

Not a known bug today, but the invariant is implicit and load-bearing. Suggest:
```python
def swap_to_real_for_rerun(self) -> bool:
    ...
    self._replay_data = None
    self._replay_index = 0
    self._current_test_id = None  # detached for the duration of the rerun
```
Plus a one-line class doc: "during a rerun, the wrapper is detached and observes `_current_test_id is None` until `restore_from_rerun`."

soda-tests/src/helpers/snapshot_pytest_plugin.py
```diff
@@ -0,0 +1,348 @@
...
+def _runtestprotocol(item, nextitem):
+    """Indirection around the pytest runner so the private import is local.
+
+    F12: ``_pytest.runner.runtestprotocol`` is private API. Importing here
+    raises a clear error on incompatible pytest versions rather than at
+    plugin load time, and keeps the import out of the module's public
+    surface.
+    """
+    try:
+        from _pytest.runner import runtestprotocol
```
m1n0, line 188: **H4 — private pytest API, no pin.** The plugin imports `_pytest.runner.runtestprotocol`, and the wrap error message says "Pin a compatible pytest version." But `soda-tests/pyproject.toml` declares only an unbounded `pytest` dep. A pytest minor that restructures `_pytest.runner` (it has churned across 7.x → 8.x) will break every snapshot run on `main` and look like a flaky-tests outage, not a dep upgrade. Suggest:

- Add an upper bound in `soda-tests/pyproject.toml` (`pytest>=8.0,<9.0` if that's what you've validated against).
- Add a CI matrix cell that runs strict mode against the lower + upper bounds — at minimum on the snapshot-plugin tests, so an upgrade attempt fails loudly.
- Optionally vendor a tiny `_runtestprotocol` reimplementation (it's ~30 lines) to remove the private-API dependency entirely.

soda-tests/src/helpers/snapshot_pytest_plugin.py
```diff
@@ -0,0 +1,348 @@
...
+def _deactivate_lingering_dwh_interceptor() -> None:
+    """If a DWH interceptor outlived the failed attempt, deactivate it.
+
+    soda-extensions ships ``DwhSnapshotInterceptor``; its ``_active_instance``
+    class attribute points at any live interceptor. We deactivate before the
+    rerun so the next attempt re-installs cleanly. Import is lazy + best-
+    effort so soda-core (which has no DWH module) doesn't break.
+    """
+    try:
+        from test_helpers.dwh_setup import DwhSnapshotInterceptor  # type: ignore
```
m1n0: **New foundational lens — replace this `getattr` with a registry.** The plugin reaches into soda-extensions via `from test_helpers.dwh_setup import DwhSnapshotInterceptor` + `getattr(DwhSnapshotInterceptor, "_active_instance", None)`. Two problems for the template:

1. If soda-extensions renames the module or the attribute, this lazy import silently no-ops (you'd find out via a flaky/duplicate-interceptor symptom, not a clear failure).
2. Any future "thing that needs to know about reruns" in another extension has to be hard-coded in this same `try/except`. That doesn't scale to the pattern this PR is establishing.

A registry mirrors what you already do for wrappers (`_LIVE_SNAPSHOT_WRAPPERS`):
```python
# in soda-tests/src/helpers/snapshot_interceptor.py
class SnapshotInterceptor(Protocol):
    def deactivate(self) -> None: ...

_REGISTERED: "WeakSet[SnapshotInterceptor]" = WeakSet()
def register_interceptor(i): _REGISTERED.add(i)
```
Then `DwhSnapshotInterceptor` calls `register_interceptor(self)` at activation, and this plugin walks `_REGISTERED` — no reflection, no string-based contract.

m1n0 (commented):
Inline notes — most are H-level (worth addressing before this becomes the template for soda-extensions). No production code touched, verified.

Niels-b:
> Ok this is super fancy, thanks for this work Niels. It did make me think - wouldn't it be easier to adjust recon checks to be compatible with the existing system? 🤔

That's for another time. Goal of this is really to never touch the production code and we only modify testing code. It should be compatible now with the cursor overrides 😄

m1n0 (approved):

Issue #379, opened 2021-07-05
Write scan result object to a file.


abhishek-khare:
- As a data engineer I want to write the scan results object to a file via an API.
 - It should write the errors, measurements and the test_results.
 - _Optional_ - choice of output format as a parameter to the API.

JCZuurmond:
This is relevant to #504, the outputs of a scan is input for a diff.

PR #2699, opened 2026-05-12
refactor(DTL-1780): consolidate aliasing under ALIAS — drop COLUMN/COUNT field_alias

m1n0:
## Summary

The codebase had three aliasing mechanisms in the SQL AST:

1. `COLUMN.field_alias` (per-class attribute)
2. `COUNT.field_alias` (per-class attribute)
3. `ALIAS(expression, alias)` (general wrapper, introduced in PR #2695)

`ALIAS` is strictly more powerful — wraps any `SqlExpression`. The per-class attrs only worked for `COLUMN`/`COUNT` and existed for historical reasons. This PR removes them in favor of `ALIAS`.

Tracking: [DTL-1780](https://linear.app/sodadata/issue/DTL-1780).

## Changes

- Drop `COLUMN.field_alias` and `COUNT.field_alias` dataclass attributes.
- Drop dispatcher branches for them in `sql_dialect.py` (base + athena override).
- `COLUMN.AS(alias)` returns `ALIAS(self, alias)` instead of a new `COLUMN`.
- `COLUMN.IN()` no longer propagates `field_alias`.
- `STAR.convert_to_standard_column` drops the unused `field_alias` parameter.
- Migrate call sites: 11 in soda-core (metadata queries, postgres, redshift). Soda-extensions migrated in a [matching branch](https://github.com/sodadata/soda-extensions/pull-or-branch/DTL-1780-consolidate-aliasing-under-alias).

## Why now

- Single canonical mechanism for output-column aliasing. Reduces cognitive load and the surface area for inconsistencies.
- Removes redundant dispatcher branches and dataclass attrs.

## Test plan

- [x] soda-core 911 passed on duckdb. No SQL output changes — metadata queries produce identical strings.
- [x] soda-extensions 136 passed (reconciliation + groupby + migration unit suites).
- [ ] Reviewer to verify diff is purely mechanical and emits identical SQL.

## Related

- Introduced `ALIAS`: PR #2695

### DIFF
--- a/soda-core/src/soda_core/common/statements/metadata_tables_query.py
+++ b/soda-core/src/soda_core/common/statements/metadata_tables_query.py
@@ -71,10 +71,10 @@ class MetadataTablesQuery:
             ),
             SELECT(
                 [
-                    self.sql_dialect.column_table_catalog() or ALIAS(LITERAL(None), "database_name"),
+                    self.sql_dialect.column_table_catalog() or LITERAL(None).AS("database_name"),
                     self.sql_dialect.column_table_schema(),
                     self.sql_dialect.column_table_name(),
-                    self.sql_dialect.column_table_type() or ALIAS(LITERAL(None), "table_type"),
+                    self.sql_dialect.column_table_type() or LITERAL(None).AS("table_type"),
                 ]
             ),
         ]
--- a/soda-core/src/soda_core/common/statements/metadata_tables_query.py
+++ b/soda-core/src/soda_core/common/statements/metadata_tables_query.py
@@ -71,10 +71,10 @@ class MetadataTablesQuery:
             ),
             SELECT(
                 [
-                    self.sql_dialect.column_table_catalog() or COLUMN(LITERAL(None), field_alias="database_name"),
+                    self.sql_dialect.column_table_catalog() or ALIAS(LITERAL(None), "database_name"),
                     self.sql_dialect.column_table_schema(),
                     self.sql_dialect.column_table_name(),
-                    self.sql_dialect.column_table_type() or COLUMN(LITERAL(None), field_alias="table_type"),
+                    self.sql_dialect.column_table_type() or ALIAS(LITERAL(None), "table_type"),
                 ]
             ),
         ]
--- a/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
+++ b/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
@@ -346,7 +346,7 @@ class PostgresSqlDialect(SqlDialect, sqlglot_dialect="postgres"):
                     END AS "datetime_precision"
                 """
                 ),
-                ALIAS(current_database_expression, "table_catalog"),
+                current_database_expression.AS("table_catalog"),
                 COLUMN("nspname", table_alias="n").AS("table_schema"),
                 COLUMN("relname", table_alias="c").AS("table_name"),
                 RAW_SQL(self.relkind_table_type_sql_expression()),
--- a/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
+++ b/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
@@ -12,7 +12,6 @@ from soda_core.common.metadata_types import (
     SqlDataType,
 )
 from soda_core.common.sql_ast import (
-    ALIAS,
     AND,
     CAST,
     COLUMN,
--- a/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
+++ b/soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
@@ -12,6 +12,7 @@ from soda_core.common.metadata_types import (
     SqlDataType,
 )
 from soda_core.common.sql_ast import (
+    ALIAS,
     AND,
     CAST,
     COLUMN,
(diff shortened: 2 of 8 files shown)


soda-core/src/soda_core/common/statements/metadata_tables_query.py
```diff
@@ -71,10 +71,10 @@ def build_sql_statement(
             ),
             SELECT(
                 [
-                    self.sql_dialect.column_table_catalog() or COLUMN(LITERAL(None), field_alias="database_name"),
+                    self.sql_dialect.column_table_catalog() or ALIAS(LITERAL(None), "database_name"),
```
Niels-b: NIT: not having `COLUMN` there anymore makes it a bit harder to read imo. It now requires an extra mental cycle to think if this is a column or not.

Just throwing the idea out here / thinking out loud, what if we go for `ALIAS(COLUMN(...), "table_type")`? 
Unsure of that one as it requires a lot more code to achieve the same thing.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
```diff
@@ -264,11 +265,11 @@ def _build_pg_columns_metadata_query(
 
         select_columns = []
         if include_table_name_column:
-            select_columns.append(COLUMN("relname", table_alias="c", field_alias="table_name"))
+            select_columns.append(COLUMN("relname", table_alias="c").AS("table_name"))
```
Niels-b, line 267: This seems inconsistent to me 🤔 . Maybe the suggestion I made above would help?

Niels-b: Ah, but that happens behind the scenes. `.AS()` calls `ALIAS`. Ok that makes sense. Just had to make the "click".

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
```diff
@@ -345,9 +346,9 @@ def _build_pg_columns_metadata_query(
                     END AS "datetime_precision"
                 """
                 ),
-                COLUMN(current_database_expression, field_alias="table_catalog"),
-                COLUMN("nspname", table_alias="n", field_alias="table_schema"),
-                COLUMN("relname", table_alias="c", field_alias="table_name"),
+                ALIAS(current_database_expression, "table_catalog"),
+                COLUMN("nspname", table_alias="n").AS("table_schema"),
+                COLUMN("relname", table_alias="c").AS("table_name"),
```
Niels-b: This still seems inconsistent to me 🤔 . Why do we have `ALIAS` for the first, but then `AS` everywhere else?

Niels-b (commented):
Thanks for following up on this.
Left some comments that intuitively felt a little off to me. No strong opinion, just want to open the discussion 😄 .

soda-core/src/soda_core/common/statements/metadata_tables_query.py
m1n0: Re-evaluated this one. The line is the fallback path — `column_table_catalog()` returns `None` for dialects that don't expose a catalog column, so the value being aliased is literally `NULL`, not a column:

```python
self.sql_dialect.column_table_catalog() or ALIAS(LITERAL(None), "database_name"),
```

Pre-PR the fallback was `COLUMN(LITERAL(None), field_alias="database_name")`, which leaned on `COLUMN` as a generic "alias-able expression" container — a bit dishonest since `LITERAL(None)` isn't a column reference. `ALIAS(LITERAL(None), …)` makes the shape explicit: we're emitting `NULL AS database_name`. Wrapping `LITERAL(None)` in `COLUMN(...)` would reintroduce the same misleading hint.

Leaving as-is.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
m1n0: Real reason: `.AS()` only lives on `COLUMN`. Line 349 is wrapping `current_database_expression`, which is a `RAW_SQL` — it has no `.AS()` method, so the explicit `ALIAS(...)` wrapper is the only option there. Lines 350-351 wrap `COLUMN` instances, so the `.AS()` sugar works.

So the rule today is: `COLUMN(...).AS("x")` when the expression is a column; `ALIAS(expr, "x")` for everything else (`RAW_SQL`, `COUNT`, `COMBINED_HASH`, `LITERAL`, …).

If we want full consistency we could promote `.AS()` onto `SqlExpression` (~2 lines) so `RAW_SQL(...).AS("x")` works. Happy to do that as a tiny follow-up — kept it out of this PR to keep the scope to the `field_alias` removal.

m1n0: Follow-up applied in cae62b54: promoted `.AS(alias)` onto `SqlExpression` so every expression type — `LITERAL`, `RAW_SQL`, `COUNT`, `COMBINED_HASH`, etc. — exposes the fluent form. Dropped the now-redundant `COLUMN.AS` override.

The metadata-query and postgres call sites in this PR are now uniform — `LITERAL(None).AS("database_name")`, `current_database_expression.AS("table_catalog")`, `COLUMN(...).AS("…")`. Matching change in soda-extensions PR #379 (commit 420f6d11) for db2/db2z/oracle and the failed-rows extractor. No SQL output changes.

m1n0:
@Niels-b — follow-up from your review is in. Recap:

1. **`metadata_tables_query.py:74`** (NIT, NULL-fallback readability) — replied on the thread. The fallback wraps `LITERAL(None)` (literal NULL), not a column reference, so wrapping in `COLUMN(...)` would re-introduce the same misleading hint the PR set out to remove. Left as-is, but migrated to the fluent form below.
2. **`postgres_data_source.py:268`** — your follow-up self-resolved (`.AS()` was the missing click).
3. **`postgres_data_source.py:351`** (ALIAS vs AS inconsistency) — fixed properly: promoted `.AS(alias)` onto `SqlExpression` (`b991799f`), so every expression type now exposes the fluent form. Migrated all PR-touched call sites in metadata queries and `postgres_data_source.py` to `<expr>.AS("name")` so the SELECT lists read uniformly. Added a unit test covering `LITERAL`, `COUNT`, `RAW_SQL`.

Matching commit in soda-extensions PR #379 (`dfd537ee`) for db2 / db2z / oracle metadata queries and the failed-rows extractor.

No SQL output changes — `.AS(x)` is just `ALIAS(self, x)`. Branch is rebased on current main and CI is green. Ready for another look.

Niels-b (approved):
Thanks for the follow up. It's also more clear to me now. Thanks 🙌 !

PR #2702, opened 2026-05-13
chore: add Dremio to schema cleanup test + dry-run mode

paulteehan:
## Summary

- Adds Dremio to `DATASOURCES_TO_RUN` in the CI `test_drop_old_schema` integration test
- Adds a `SODA_SCHEMA_CLEANUP_DRY_RUN=true` env var that logs targets without dropping them — useful for validating the matching logic before the first prod run
- Splits schema enumeration so Dremio uses a custom listing path: Dremio exposes nested folder paths in `INFORMATION_SCHEMA."SCHEMATA"` as dotted strings (e.g. `nas.bucket.host.ALTERNATE_DWH_…`), so we scope to the test-root path (`dataset_prefix[:-1]`) and match `LIST_OF_PREFIXES_TO_DROP` against the trailing component while passing the full dotted path to `drop_schema_if_exists`

Companion fix in soda-extensions: https://github.com/sodadata/soda-extensions/pull/378 (Dremio `DROP FOLDER` syntax in the test helper).

## Test plan

- [x] Dry-run against prod Dremio: `SODA_SCHEMA_CLEANUP_DRY_RUN=true TEST_DATASOURCE=dremio uv run pytest soda-tests/tests/integration/test_drop_old_schema.py` — passes, lists targets only
- [x] Real run against prod Dremio: same command without the dry-run var — actively dropping aged `ALTERNATE_DWH_*` folders without errors
- [ ] No regression against other data sources (logic for non-Dremio types is unchanged apart from variable renaming)

### DIFF
--- a/soda-tests/tests/integration/test_drop_old_schema.py
+++ b/soda-tests/tests/integration/test_drop_old_schema.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime
 import logging
+import os
 
 import pytest
 from dateutil.parser import parse
@@ -23,11 +24,16 @@ DATASOURCES_TO_RUN = [
     "redshift",
     "oracle",
     "databricks",
+    "dremio",
 ]
 # Only drop schemma's starting with these prefixes.
 LIST_OF_PREFIXES_TO_DROP = ["soda_diagnostics_", "ALTERNATE_DWH_", "ci_", "my_dwh_"]
 # Schema's starting with these prefixes are exempt from being dropped.
 LIST_OF_EXEMPTIONS = ["soda_diagnostics_dev_"]
+
+
+def _dry_run() -> bool:
+    return os.getenv("SODA_SCHEMA_CLEANUP_DRY_RUN", "false").lower() in ("1", "true", "yes")
 
 
 def determine_if_schema_needs_to_be_dropped(schema_name: str) -> bool:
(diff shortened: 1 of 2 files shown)


soda-tests/tests/integration/test_drop_old_schema.py
```diff
@@ -77,6 +83,41 @@ def determine_if_schema_needs_to_be_dropped(schema_name: str) -> bool:
...
+        full_name: str = row[0]
+        if not full_name or not full_name.startswith(test_root_prefix):
+            continue
+        leaf: str = full_name.rsplit(".", 1)[-1]
+        candidates.append((full_name, leaf))
+    logger.info(
+        f"Dremio: found {len(candidates)} schemas under test root '{test_root_dotted}' "
+        f"(out of total schemas in INFORMATION_SCHEMA)"
+    )
+    return candidates
```
Niels-b: This should really happen in a dremio specific file. (datasourcetesthelper perhaps)?

soda-tests/tests/integration/test_drop_old_schema.py
```diff
@@ -89,33 +130,47 @@ def test_drop_old_schemas(data_source_test_helper: DataSourceTestHelper):
...
-    # First get the list of all the schema's in the database.
-    table_namespace, schema_name = data_source_test_helper.data_source_impl._build_table_namespace_for_schema_query(
-        data_source_test_helper.dataset_prefix
-    )
-    schemas_query_sql = dialect.build_schemas_metadata_query_str(table_namespace=table_namespace)
-    query_result: QueryResult = data_source_test_helper.data_source_impl.execute_query(schemas_query_sql)
-    schema_names: list[str] = [row[0] for row in query_result.rows]
+    dry_run: bool = _dry_run()
+    if dry_run:
+        logger.info("SODA_SCHEMA_CLEANUP_DRY_RUN is set — will log targets but not execute DROP statements.")
```
Niels-b, line 98: I like the idea of doing a dry-run for debugging purposes. Nice addition 🙌 !

soda-tests/tests/integration/test_drop_old_schema.py
```diff
@@ -89,33 +130,47 @@ def test_drop_old_schemas(data_source_test_helper: DataSourceTestHelper):
...
+    if type_name == "dremio":
+        candidates = _list_dremio_candidate_schemas(data_source_test_helper)
+    else:
+        dialect: SqlDialect = data_source_test_helper.data_source_impl.sql_dialect
+        table_namespace, _ = data_source_test_helper.data_source_impl._build_table_namespace_for_schema_query(
+            data_source_test_helper.dataset_prefix
+        )
+        schemas_query_sql = dialect.build_schemas_metadata_query_str(table_namespace=table_namespace)
+        query_result: QueryResult = data_source_test_helper.data_source_impl.execute_query(schemas_query_sql)
+        candidates = [(row[0], row[0]) for row in query_result.rows if row[0]]
```
Niels-b: This may have to move to the `DatasourceTestHelper` I think, so we can "overwrite" it in Dremio.

soda-tests/tests/integration/test_drop_old_schema.py
```diff
@@ -89,33 +130,47 @@ def test_drop_old_schemas(data_source_test_helper: DataSourceTestHelper):
...
                     except (
                         Exception
                     ) as e:  # We catch any exception, because we don't want the CI itself to fail because of this.
                         logger.error(f"Error dropping schema: {e}")
                 else:
-                    logger.info(f"Schema name: {schema_name} is not old enough to be dropped or is an exemption")
-    logger.info(f"Number of deleted schemas: {num_deleted_schemas}")
+                    logger.info(f"Schema name: {match_name} is not old enough to be dropped or is an exemption")
+    if dry_run:
+        logger.info(f"[DRY RUN] Number of schemas that would be deleted: {num_matched_schemas}")
```
Niels-b: Might be good to give the actual list of schemas here (since it's a dry-run). I would like to see what is going to be dropped.

Note that there could always be a discrepancy between dry-run and actual run (the state of the DB may have changed in the meantime)

Niels-b (changes requested):
Nice ideas! Really like the work here, but some things will have to move to other files in my opinion.

Niels-b (approved):
Thanks for the work on this. LGTM 👍 !

PR #2704, opened 2026-05-14
fix(DTL-1799): isolate InvalidReferenceCountMetricImpl identity

m1n0:
Stacked on #2703 (DTL-1776). Merge that first.

## Summary

`InvalidReferenceCountMetricImpl` shared `metric_type=\"invalid_count\"` with the aggregation sibling `InvalidCountMetricImpl`, **and** did not include `check_filter` in its identity hash. Two reference-invalidity checks on the same column with the same `valid_reference_data` but different `filter:` clauses produced colliding metric ids; `MeasurementValues.metric_values_by_id` is a dict keyed by id, so the second `InvalidReferenceCountQuery` write clobbered the first and both checks reported the same count.

- Thread `check_filter` through `InvalidReferenceCountMetricImpl.__init__` to `super().__init__()` so it participates in the identity hash.
- Rename `metric_type` to `\"invalid_reference_count\"` so the two sibling classes can never collide on id even if all other base properties happen to match. The **public** diagnostic name `\"invalid_count\"` stays unchanged (it's the dict key in `evaluate()`; the rename only affects the internal `Measurement.metric_name`).

## Why not a shared base class with `RowsTestedQueryMetricImpl`?

Considered and rejected. The discriminators are fundamentally different:
- `RowsTestedQueryMetricImpl` needs the user-supplied raw SQL string in its identity — it overrides `_get_id_properties` to add it.
- `InvalidReferenceCountMetricImpl` only needs `check_filter` + `missing_and_validity` (including `ref_*` fields) — all already in `MetricImpl._get_id_properties`. No override needed.

A shared base would be a marker class with no shared behavior. Skipping.

## Test plan

- [x] New `test_two_invalid_reference_checks_with_different_filters_resolve_to_distinct_metrics` — two reference-invalid checks on `country` with `filter: id <= 4` and `filter: id > 4`; expected counts 1 and 2.
- [x] Test fails on pre-fix code (`assert 2 == 1` — both checks read the last-write-wins value), passes on this branch.
- [x] Full `soda-tests/tests/integration` suite green.
- [x] Full `soda-tests/tests/unit` suite green.
- [x] Pre-commit clean.
- [x] Independent reviewer pass — no findings.

## Linear

[DTL-1799](https://linear.app/sodadata/issue/DTL-1799/invalidreferencecountmetricimpl-shares-the-invalid-count-metric-type)

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
@@ -104,8 +104,17 @@ class FailedRowsCheckImpl(CheckImpl):
                 self.queries.append(failed_rows_count_query)
 
             if self.failed_rows_check_yaml.rows_tested_query and contract_impl.data_source_impl:
+                # DTL-1776: must NOT share identity with the contract-wide row_count
+                # metric. A plain RowCountMetricImpl with no filter hashes identical
+                # to ContractImpl.row_count_metric_impl, so RowsTestedQuery's
+                # measurement would overwrite dataset_rows_tested and leak this
+                # check's custom count into every other check's diagnostics.
                 self.check_rows_tested_metric_impl = self._resolve_metric(
-                    RowCountMetricImpl(contract_impl=contract_impl, check_impl=self)
+                    RowsTestedQueryMetricImpl(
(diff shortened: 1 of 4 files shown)


Niels-b (approved):
LGTM

PR #2709, opened 2026-05-18
feat(check-collections): lightweight CheckCollection abstraction (from scratch)

m1n0:
## Summary

Lightweight `CheckCollection` abstraction over `ContractImpl` so additional check-collection subtypes (data standards, future check suites, …) can plug into the same verification engine without copy-pasting the flow. From-scratch rewrite of the heavier #2707 approach.

Closes DES-315.
Absorbs DES-316 (all four Phase-2 deliverables landed here — see below).

## What lands

- **`CheckCollectionImpl`** (engine) lives in `soda-core/check_collections/base.py`. Subclasses declare five plain class attributes (`kind`, `wire_source`, `display_name`, `yaml_class`, `result_class`) and inherit `__init__` / `verify` / etc.
- **`execute_check_collections(yaml_sources, ...)`** dispatches per-item via the YAML's top-level `kind:` field (defaulting to `"contract"` when absent). Per-item error isolation with `abort_on_first_error` flag for the legacy single-input contract path.
- **`ContractImpl`** becomes a thin subclass declaring `kind = "contract"`, `wire_source = "soda-contract"`, `display_name = "contract"`, `yaml_class = ContractYaml`, `result_class = ContractVerificationResult`. Public API (`ContractVerificationSession.execute(...)`, `Contract`, `ContractVerificationStatus`, `Check.contract_file_*`, `DataSourceImpl.switch_warehouse(contract_impl=)`) is unchanged.
- **Cloud upload** reads `wire_source` from the impl instead of a hardcoded `"soda-contract"`, so future subtypes route correctly without touching upload code.

### Phase-2 features (DES-316) absorbed

- **Path prefixing** — `Check.full_path` carries `"{collection_id}.{path}"` for non-contract subtypes so backend's `firstSegmentOf(checkPath)` lookup matches `DataStandard.name`. Contracts emit bare `path` (byte-identical to today).
- **Selector / `--check-paths` matching** uses `Check.path` (yaml-internal). The wire-only `full_path` never affects selector matching.
- **Identity disambiguation** — `CheckCollectionImpl.identity_prefix()` returns `()` for contracts (byte-identical hashes) and `(wire_source, collection_id)` for non-contract subtypes by default. Locked by `test_identity_disambiguation.py`.
- **Alignment guard** — `_verify_check_sources_aligned` walks emitted checks before upload; any check whose `source` override disagrees with `self.wire_source` skips the upload entirely (defends against `DATA_STANDARDS_SOURCE_MISALIGNED` 500 from the backend).

## Kind dispatch shape

A minimal `__init_subclass__`-driven kind registry lives on `CheckCollectionImpl` (`_REGISTRY: dict[str, type]`, populated when a concrete subtype declares a non-empty `kind`). Lookup is `CheckCollectionImpl.for_kind(kind)`. No `CheckCollectionFamily` dataclass, no `register_family(...)` ceremony, no `OnAgentVerifier` Protocol, no `CheckCollectionSpec` wrapper, no `Generic[YamlT, ResultT]` seam. Subtype packages register implicitly via `class FooImpl(CheckCollectionImpl): kind = "foo"`. Side-effect-import behavior: importing the subtype's module runs the class body, which registers via `__init_subclass__` — same shape soda-autopilot uses.

## Backwards-compat shims

- **`@property contract`** on `CheckCollectionResult` returns `self.check_collection` (read-only). External callers reading `result.contract` keep working.
- **`Check.full_path: str = ""`** has a default so external constructors of `Check(...)` still work. Emission site falls back to `path` when empty.

## Test plan

- `pytest soda-tests/tests/unit/` — 791 passed, 2 skipped.
- Pre-commit hooks (autoflake/isort/black) green.
- Merge with `main` resolved: ports the new `dedup byte-identical aggregation metrics` logic (`74fecaba`) into the moved `CheckCollectionImpl._build_queries`.
- Contract verification path is byte-identical behaviorally (verified by the existing `test_check_path_emission`, `test_alignment_guard`, `test_identity_disambiguation`, `test_selector_unprefixing`, `test_build_contract_result_with_none_datasource`, etc. test suites all pass).

## Companion PRs

- soda-extensions [#386](https://github.com/sodadata/soda-extensions/pull/386) — soda-data-standard package + `verify_data_standards(...)` Python API (DES-317).
- soda-extensions [#391](https://github.com/sodadata/soda-extensions/pull/391) — `soda data-standard verify` CLI plugin (DES-317).
- soda-contracts-launcher [#215](https://github.com/sodadata/soda-contracts-launcher/pull/215) — launcher integration (DES-318).


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -448,14 +474,21 @@ def __init__(self, metric_id: str, value: any, metric_name: Optional[str]):
         self.value: any = value
 
 
-class ContractVerificationStatus(Enum):
+class CheckCollectionStatus(Enum):
```
Niels-b, line 476: Should this one not move to the Check Collections code? Or does it not make sense to have it there?

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -57,18 +59,20 @@ def register_extension(cls, name: str, extension_cls: type[ContractYamlExtension
...
+        # read their extra fields after calling ``super().__init__()``.
+        return cls(
+            contract_yaml_source=yaml_source,
             provided_variable_values=provided_variable_values,
             data_timestamp=data_timestamp,
             primary_data_source_impl=primary_data_source_impl,
         )
-        return contract_yaml
 
     def __init__(
```
Niels-b, line 58: Might be a silly remark: Is there a reason why we still overload this function? Is behavior different from the CheckCollection?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -332,10 +304,27 @@ def _execute_on_agent(
...
+                # from connecting to a data source or executing queries.
+                # Note that ``__init__`` is NOT a no-op here: it still parses
+                # the YAML's columns and checks, builds the metrics resolver,
+                # resolves the row-count metric, constructs the dataset CTE,
+                # and instantiates registered extensions. None of that hits
+                # the network — but it does run real parsing work, so a bad
+                # YAML still raises here rather than at agent dispatch time.
+                # The actual Soda Cloud client used to invoke the agent is
+                # passed below to ``verify_on_agent``.
+                contract_impl: ContractImpl = ContractImpl(
```
Niels-b, line 327: While I think the items described in the above comment are very good. Claude still pointed out a possible regression: 

Any errors logged during _parse_* land in self.logs but verify_on_agent never propagates them self.logs.pop_log_records() is never called. Mild regression if a YAML has check-shape issues that the YAML-level parse didn't catch.

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1406,23 +1417,37 @@ def to_jsonnable(o: Any, remove_null_values_in_dicts: bool = True) -> object:
...
+    # Fallback derives the scan-def name from the dataset's qualified name.
+    # For data-standards, the backend's lazy scan-def creation names them
+    # ``<dataset.contractIdentifier>_data_standards_scan`` (see
+    # ``DataStandardModule.ensureDataStandardsScanDefinition`` on the soda
+    # server). Match that convention so a local ``soda data-standard verify
+    # -p`` lands on a scan-def of the right type and clears the backend's
+    # ``DATA_STANDARDS_SOURCE_MISALIGNED`` guard.
+    qualified_name = contract_verification_result.check_collection.soda_qualified_dataset_name
+    if wire_source == "data-standard":
+        return f"{qualified_name}_data_standards_scan"
```
Niels-b: Comment from Claude:

This "leaks" some logic from soda-cloud into soda-core. It might be better to push this into the Impl for the check (e.g. `DataStandardsImpl`) and have that handle the logic.

It could also be completely fine to leave it here and have a centralized location where we do the mapping. I'm not really expecting there to be many other types of check/contract/...

soda-core/src/soda_core/check_collections/session.py
```diff
@@ -0,0 +1,164 @@
...
+    results: list[CheckCollectionResult] = []
+    for yaml_source in yaml_sources:
+        impl_class: Optional[type[CheckCollectionImpl]] = None
+        try:
+            # Peek the kind from the YAML root. Cheap parse (YAML library
+            # call); the subtype's full parse below repeats this work but
+            # also walks columns/checks/variables which is the heavy part.
+            yaml_object = yaml_source.parse()
+            kind = (yaml_object.read_string_opt("kind") if yaml_object is not None else None) or "contract"
+            impl_class = CheckCollectionImpl.for_kind(kind)
```
Niels-b, line 117: Claude flagged a discrepancy in the return types of these "kinds".
Pasting the full explanation here:

When the YAML's kind: is unknown, session.py:147 falls back to CheckCollectionImpl.build_error_result(...), which uses cls.result_class = CheckCollectionResult instead of the subtype's ContractVerificationResult. The docstring flags this ("Downstream consumers must not narrow on type"), but the ContractVerificationSessionImpl.execute return is typed list[ContractVerificationResult] —a base-class instance violates that contract. Either change the  return type to list[CheckCollectionResult], or default the fallback to ContractImpl.build_error_result when no other kind is known (since contract is the legacy default).

soda-core/src/soda_core/contracts/contract_verification.py
Niels-b: Interesting flag: we have `column_name: Optional[str]` twice 😅

Niels-b (commented):
Went trough it diagonally. Intent seems very clear and is good to me! Nice work.

Some comments though. And two general points:
- We may need to rebase on the latest main: soda-extensions CI will require the latest soda-core version and there could be a miss-match now.
- Do some check types in soda-extensions also need some changes to account for changes made here? I'm assuming that's done in other PRs (haven't managed to look at them yet).

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+    ``ContractVerificationResult`` is a subclass preserving the historical name.
+
+    ``check_collection`` carries the wire-source-neutral metadata about the
+    parsed file (data source name, dataset prefix, qualified name, file
+    source info). The field type is still ``Contract`` until the future
+    ``Contract`` → ``CheckCollection`` class rename lands; only the field
+    name is neutral in this PR.
+    """
+
+    check_collection: Contract
```
paulteehan, line 72: this was surprising to me, I was expecting to see a class named CheckCollection

I see that some renaming is pending, but it seems like a relatively small change to introduce a CheckCollection class and make Contract inherit from it, and that would make this PR more cohesive

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+        return "\n".join(self.get_errors())
+
+    def get_warnings(self) -> list[str]:
+        return [r.getMessage() for r in self.log_records if r.levelno == WARNING] if self.log_records else []
+
+    def get_warnings_str(self) -> str:
+        return "\n".join(self.get_warnings())
+
+    @property
+    def contract(self) -> Contract:
```
paulteehan: could also define a new ContractResult class which inherits from CheckCollectionResult and put the contract property on that one only, would be cleaner.

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+
+    check_collection: Contract
+    data_source: Optional[DataSource]
+    data_timestamp: Optional[datetime]
+    started_timestamp: datetime
+    ended_timestamp: datetime
+    status: CheckCollectionStatus
+    measurements: list[Measurement]
+    check_results: list[CheckResult]
+    sending_results_to_soda_cloud_failed: bool
```
paulteehan, line 80: kind of oddly specific, not a big deal, but I would expect something like soda_cloud_send_result: SodaCloudSendResult, with a new class to capture that.   pinning to a boolean could be restrictive

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+
+
+class CheckCollectionImpl:
+    """Engine that verifies one check-collection file against a data source.
+
+    Subclasses provide four plain class attributes; the engine inherits.
+
+    Example subtype declaration::
+
+        class DataStandardImpl(CheckCollectionImpl):
```
paulteehan: this is very nice

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+
+    # Subtype identity — declared per-subtype as a plain class attribute.
+    # The base default is empty string → not registered.
+    kind: str = ""
+
+    # Subclasses MUST override these. ``wire_source`` is guarded at the top of
+    # ``verify()`` so a missing override raises immediately rather than silently
+    # routing to no Cloud feature.
+    wire_source: str = ""
+    display_name: str = "check collection"
```
paulteehan: minor thing, how do subclass implementors know what style to follow?  I could imagine saying 'check collection', 'check_collection', 'Check Collection', etc

maybe display name should be a property which is by default a conversion of 'kind', and can be overridden if necessary

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+                    self.datasource_warehouse = data_source_impl.data_source_connection.connection_properties.warehouse
+
+                if self.datasource_warehouse is None:
+                    self.datasource_warehouse = data_source_impl.get_current_warehouse()
+
+        self.row_count_metric_impl = self.metrics_resolver.resolve_metric(RowCountMetricImpl(contract_impl=self))
+        self.dataset_rows_tested: Optional[int] = None
+
+        # Dataset defining CTE - used as basis for all queries in this collection
+        self.cte = CTE("_soda_filtered_dataset").AS(
```
paulteehan, line 399: I was a proponent of CTEs in the past but I'm wary of using them now because we hit some issues with  custom SQL.   some DBs (sql server) impose stricter requirements on CTE SQL than standard SQL.    I don't think it would be an issue in this context but I'd recommend you keep an eye on it

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -0,0 +1,1013 @@
...
+        if not self.wire_source:
+            raise ValueError(f"{type(self).__name__} did not declare wire_source class attribute")
+
+        # Non-contract subtypes MUST declare collection_id: the wire
+        # ``checkPath`` is prefixed with ``collection_id`` so the backend's
+        # ``firstSegmentOf(checkPath)`` filter can match ``DataStandard.name``.
+        # Without a collection_id the emitted checks would silently degrade
+        # to unprefixed paths the backend would then drop. Fail loudly here
+        # instead of producing a confused upload.
+        if self.wire_source != "soda-contract" and not self.collection_id:
```
paulteehan: again I think a ContractImpl class which handles the back-compatibility would be cleaner here, ideally for me CheckCollections wouldn't even know what a contract is

soda-core/src/soda_core/check_collections/session.py
```diff
@@ -0,0 +1,164 @@
...
+
+            yaml = impl_class.yaml_class.parse(
+                yaml_source=yaml_source,
+                provided_variable_values=variables,
+                data_timestamp=data_timestamp_str,
+                primary_data_source_impl=primary_data_source_impl,
```
paulteehan, line 123: is primary_data_source_impl distinct from data_source_impl?  can't tell from the code

soda-core/src/soda_core/contracts/contract_verification.py
paulteehan, line 278: maybe a future PR, I would expect contract*.py to only contain contract-related classes and move the more generic ones like Check elsewhere

paulteehan (approved):
Very nice idea and nice work overall, I feel a sense of peace reading this :)

My only real issue is that it feels incomplete having some Contract naming still sticking around, but I assume that's coming in a future PR

soda-core/src/soda_core/contracts/contract_verification.py
m1n0: Removed in e23351f4.

soda-core/src/soda_core/contracts/contract_verification.py
m1n0: Left in place in e23351f4. The enum is already neutrally named (`CheckCollectionStatus`) with a BC alias `ContractVerificationStatus = CheckCollectionStatus`. Moving just the enum into `check_collections/base.py` would leave `CheckOutcome`, `CheckResult`, `Contract`, `Measurement`, `PostProcessingStage`, `ScanTokenUsage`, `YamlFileContentInfo` in `contract_verification.py` — all imported by `base.py` the same way. The misnomer is the module name, not the enum's location; the real cleanup is renaming the module, deferred to a follow-up.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
m1n0: Good catch. The override existed solely to bridge a kwarg-name mismatch (`parse(yaml_source=...)` → `cls(contract_yaml_source=...)`). In e23351f4 I renamed `ContractYaml.__init__`'s parameter to `yaml_source` and dropped the override; `CheckCollectionYaml.parse` now provides a real default body. Cascaded the rename to `DataStandardYaml.__init__` in soda-extensions PR #386 (commit f3822b88).

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Confirmed regression. `_parse_columns`/`_parse_checks` errors flowed into the impl's `self.logs`, but `soda_cloud.verify_contract_on_agent` builds the returned result's `log_records` from a `Logs()` started AFTER impl construction, so init-time records were silently dropped.

Fix in e23351f4 (`_execute_on_agent`): pop the impl's init records via `self.logs.pop_log_records()`, detach the LogCapturer from root, then prepend them onto the returned result. Also closes a secondary leak where the impl's LogCapturer stayed attached across loop iterations. Pinned by two tests in `soda-tests/tests/unit/test_agent_path_log_propagation.py`.

soda-core/src/soda_core/common/soda_cloud.py
m1n0: Relocated in e23351f4. Added a `scan_definition_suffix: Optional[str] = None` class attribute on `CheckCollectionImpl`, threaded it alongside `wire_source` through `send_contract_result` → `_build_contract_result_json_dict` → `_build_scan_definition_name`. Removed the `if wire_source == "..."` branch from soda_cloud.py — subtypes now drive scan-def naming via the class attribute, and soda-core common stays subtype-neutral. Pinned by tests in `test_contract_wire_source_plumbed_to_upload.py` (suffix default, explicit suffix, no wire_source branching).

soda-core/src/soda_core/check_collections/session.py
m1n0: Real type-contract bug, fixed in e23351f4. Added `default_impl_class: Optional[type[CheckCollectionImpl]]` kwarg to `execute_check_collections`; `ContractVerificationSessionImpl.execute` passes `default_impl_class=ContractImpl` so the unknown-kind fallback uses `ContractImpl.build_error_result` and returns a `ContractVerificationResult` instance — matching the declared `list[ContractVerificationResult]` return. Pinned by `test_execute_check_collections_unknown_kind_fallback_uses_default_impl_class`.

m1n0:
Thanks for the thorough pass. All six inline threads addressed in e23351f4 — see each thread for specifics. Quick summary:

| Item | Resolution |
|---|---|
| N6 (dup `column_name`) | Removed. |
| N3 (agent log loss) | Pop init logs in `_execute_on_agent`, prepend onto result, detach capturer. New tests in `test_agent_path_log_propagation.py`. |
| N5 (build_error_result type) | `default_impl_class` kwarg on `execute_check_collections`; contract path passes `ContractImpl`. New test. |
| N2 (parse overload) | Renamed kwarg to `yaml_source`, dropped override. Cascaded to `DataStandardYaml` in soda-extensions #386 (f3822b88). |
| N4 (wire_source leak) | `scan_definition_suffix` class attribute on `CheckCollectionImpl`; soda_cloud branch removed. New tests. |
| N1 (enum location) | Left in place — module rename is the real cleanup, deferred. |

On your two general points:

- **Rebase**: already current. Merged `origin/main` 4 commits ago (`96f51147`); `git log HEAD..origin/main` is empty after a fresh fetch.
- **soda-extensions cascade**: companion PRs #386, #391, and soda-contracts-launcher #215 already use the new neutral names. The one remaining piece — `DataStandardYaml.__init__` kwarg — lands in soda-extensions/#386 (f3822b88) alongside this commit.

Suite green: 969 passed (soda-core unit + integration + feature against duckdb), 363 + 133 passed in soda-extensions. Pre-commit clean.

soda-core/src/soda_core/check_collections/base.py
m1n0: Agreed it would be more cohesive. The PR docstring already flags this as deferred ("only the field name is neutral in this PR") because introducing `CheckCollection` and making `Contract` inherit ripples through every check-collection result consumer (launchers, soda-extensions, soda-autopilot, public API). I'd rather land that as its own follow-up — same workstream as Niels's N1 about generic types living in `contract_verification.py` (see [#2709 comment](https://github.com/sodadata/soda-core/pull/2709#discussion_r3274166954)).

soda-core/src/soda_core/check_collections/base.py
m1n0: Done in a1b7b029. Moved `@property contract` off `CheckCollectionResult` and onto `ContractVerificationResult` — the wire-source-neutral base no longer carries the contract-named alias.

soda-core/src/soda_core/check_collections/base.py
m1n0: Fair point — pinning to `bool` discards the "why" if it ever becomes useful. Leaving as-is for this PR because no current consumer (engine, launchers, tests, autopilot) reads anything beyond pass/fail, so introducing `SodaCloudSendResult` adds surface area without a caller. Happy to upgrade when the first consumer needs structured failure detail.

soda-core/src/soda_core/check_collections/base.py
m1n0: Thanks! Restored the example block in a1b7b029 with a generic subtype name.

soda-core/src/soda_core/check_collections/base.py
m1n0: Done in a1b7b029. `display_name` is now a `@property` defaulting to `self.kind` with hyphens swapped for spaces. Subclasses override by declaring a plain `display_name` class attribute, which shadows the descriptor via normal Python MRO lookup. So a future `MyImpl` with `kind = "my-subtype"` automatically shows up as `"my subtype"` in logs unless overridden.

soda-core/src/soda_core/check_collections/base.py
m1n0: Noted — will keep an eye on it. The CTE here only wraps the dataset filter, which is straightforward SELECT/WHERE shape, so I don't expect SQL Server's stricter CTE rules to bite. If they do we have a clean fall-back to expanding the CTE inline at the dialect layer.

soda-core/src/soda_core/check_collections/base.py
m1n0: Good call — addressed in a1b7b029. The base now has `requires_collection_id: bool = True` and the guard is `if self.requires_collection_id and not self.collection_id`. `ContractImpl` sets `requires_collection_id = False` to opt out. Also applied the same pattern to `identity_prefix`: base default is `(wire_source, collection_id)`; `ContractImpl` overrides to return `()` for byte-identical historical hashes. The "soda-contract" string literal no longer appears anywhere in `CheckCollectionImpl`.

soda-core/src/soda_core/check_collections/session.py
m1n0: Yes, distinct on purpose — they can be the same instance but usually aren't:

- `data_source_impl`: the per-impl **execution** target. The DS the engine runs the check queries against.
- `primary_data_source_impl`: the dialect-bearing DS used **only during YAML parsing** in `ContractYaml.__init__` (for dialect-specific timestamp conversion in variable resolution — see the `f_convert_str_to_datetime = primary_data_source_impl.sql_dialect.convert_str_to_datetime` branch).

A multi-data-source contract has one primary (for parsing) and multiple per-impl execution targets. Naming was inherited from the pre-refactor convention; happy to rename to something clearer (e.g. `yaml_parse_data_source_impl`) in the same follow-up that does the broader module rename.

soda-core/src/soda_core/contracts/contract_verification.py
m1n0: Agreed and same theme as your P1 + Niels's [N1](https://github.com/sodadata/soda-core/pull/2709#discussion_r3273932180): `contract_verification.py` is now mis-named — it hosts shared public-API types (`Check`, `CheckResult`, `CheckCollectionStatus`, `Contract`, ...) that the new `check_collections/` engine imports from. The fix is a module-level rename + class-level `Contract → CheckCollection` rename, landing as its own follow-up. Tracked separately.

m1n0:
Thanks Paul. The umbrella point is fair — the abstraction stops one rename short of being fully cohesive, and that's deliberate to keep this PR reviewable. Inline replies on each thread; quick summary:

| Item | Resolution |
|---|---|
| P2 — `@property contract` on base | Moved to `ContractVerificationResult` subclass (a1b7b029). |
| P4 — example subtype block | Restored with a generic class name (a1b7b029). |
| P5 — `display_name` style | Now a `@property` defaulting to `kind` with hyphens→spaces; subclasses override via plain class attribute, shadows the descriptor via MRO (a1b7b029). |
| P7 — `wire_source == "soda-contract"` guard | Replaced with `requires_collection_id: bool = True` class attribute; `ContractImpl` opts out. Same shape applied to `identity_prefix()` — base default is the disambiguated tuple; `ContractImpl` overrides to `()`. The "soda-contract" literal no longer appears in `CheckCollectionImpl` (a1b7b029). |
| P1 / P9 — `Contract` → `CheckCollection` rename + `contract_verification.py` module rename | Deferred to a follow-up workstream — same theme as Niels's [N1](https://github.com/sodadata/soda-core/pull/2709#discussion_r3273932180). Tracked separately. |
| P3 — `SodaCloudSendResult` object | YAGNI for now — no current consumer reads more than pass/fail. Will upgrade when the first one needs structured failure detail. |
| P6 — SQL Server CTE caution | Noted. Current CTE is straightforward SELECT/WHERE shape; clean fall-back at the dialect layer if it bites. |
| P8 — `primary_data_source_impl` distinction | Explained in thread — they're distinct (parse-time dialect-bearing DS vs execution target). Naming follow-up bundled with P1/P9. |

After this round, `CheckCollectionImpl` no longer has any "soda-contract" string literals — subtypes are first-class peers configured purely via class attributes / overrides. Suite green: 969 passed in soda-core against duckdb, 352 in soda-extensions, pre-commit clean.

PR #2717, opened 2026-05-19
return None for identities in FIPS environments where blake2b is unavailable

### LINKED ISSUE #2354: Soda Core fails in FIPS-enabled environments due to use of hashlib.blake2b

bujjibabukatta:
Fixes #2354

## Problem
In FIPS 140-2 compliant environments (government/finance hardened Linux),
`hashlib.blake2b` is disabled at the OS level causing `ImportError` at
startup, making Soda Core completely unusable in these environments.

## Approach
Following the maintainer's agreed direction in #2354 — rather than
falling back to a different hash algorithm (which would make identities
environment-dependent), return `None` when `blake2b` is unavailable.

Identities are only meaningful for Soda Cloud users. Returning `None`
is safe and honest for soda-core-only users in FIPS environments.

## Changes
Single file change in `consistent_hash_builder.py`:
- Wrap `blake2b` import in `try/except ImportError`
- Guard all `blake2b` calls — skip hashing when unavailable
- `get_hash()` already returns `None` when nothing was hashed — no change needed

## Testing
To verify: run in a FIPS-enabled environment or mock the ImportError by
temporarily removing the `blake2b` import. `get_hash()` should return `None`.

### DIFF
--- a/soda-core/src/soda_core/cli/cli.py
+++ b/soda-core/src/soda_core/cli/cli.py
@@ -215,6 +215,14 @@ def _setup_contract_verify_command(contract_parsers) -> None:
         help="Specify the path to the diagnostics warehouse configuration file. ",
     )
 
+    verify_parser.add_argument(
+    "--dry-run",
+    const=True,
+    action="store_const",
+    default=False,
+    help="Print the SQL queries that would be executed without running them against the data source.",
+)
+
     def handle(args):
         contract_file_path = args.contract
         dataset_identifier = args.dataset
@@ -228,6 +236,7 @@ def _setup_contract_verify_command(contract_parsers) -> None:
         use_agent = args.use_agent
         blocking_timeout_in_minutes = args.blocking_timeout_in_minutes
         diagnostics_warehouse_file_path = args.diagnostics_warehouse
+        dry_run = args.dry_run
 
         # Parse --check-filter expressions
         try:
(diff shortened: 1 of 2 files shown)


bujjibabukatta:
Hi @CLAassistant  can you please approve and merge it

Issue #396, opened 2021-07-15
Snowflake Table containing columns with Non ascii characters are not parsed correctly when using Group by SQL query


sujithvarghese:
**Describe the bug**
I have trouble scanning a snowflake table containing columns with non ascii characters. 
For eg:  When I run the below sql query in snowflake , columns with non ascii characters are enclosed in quotes and it works fine.
` select "KSÆS",KHUSNO,KLEJH,ISCURRENT,COUNT(ISCURRENT) AS DUPLICATE_ISCURRENT_RECORDCOUNT
   from  MYTABLE
   GROUP by "KSÆS",KHUSNO,KLEJH,ISCURRENT
   HAVING COUNT(*) > 1  AND ISCURRENT = 1`

If I try to run the same sql query through Soda scan using sql metrics, I am getting an Exception. Looks like non ascii character is not decoded correctly. Getting exception as below
 ` | Errors occurred!
  |   [error] Exception during sql metric groups query select "KSÃ†S",KHUSNO,KLEJH,ISCURRENT,COUNT(ISCURRENT) AS DUPLICATE_ISCURRENT_RECORDCOUNT
from  MYTABLE
GROUP BY "KSÃ†S",KHUSNO,KLEJH,ISCURRENT
HAVING COUNT(*) > 1  AND ISCURRENT = 1
`

**To Reproduce**
1. Have a table with columns containing non ascii characters like Æ, Ø
2. Create some records. Construct a sql query  such that you can query the table with Group by on those columns containing non ascii characters (Attached the scan.yml )
3. Run `soda scan ...` on the table

**Context**
__Included  scan.yml 

**OS**: Windows
**Python Version**: 3.9.5
**Soda SQL Version**:  Soda CLI version 2.0.0b27
**Warehouse Type**: snowflake

vijaykiran:
Thanks for the issue @sujithvarghese ! Just a quick note `Soda CLI version 2.0.0b27` can you upgrade to latest version ?

sujithvarghese:
hi @vijaykiran 
Sce 1:
Below are the python packages and their corresponding versions that I am using where I got the Exception
snowflake-connector-python==2.4.0
soda-sql==2.0.0b27

Sce:2
I tried and upgraded to below versions. Still got the same exception
snowflake-connector-python==2.6.0
soda-sql==2.0.0b27 (This is the latest version)

Sce3:
I see that there is another package called 'soda-sql-core' with version as 2.1.0b11.
[...]

PR #2719, opened 2026-05-22
feat: combined uploads + session rollups for check-collection subtypes (DES-348)

m1n0:
## Summary

Engine-level changes to `CheckCollectionImpl` and `execute_check_collections` that let any subtype opt into session-level combined Cloud uploads, session-wide invariants, and rollup-based exit-code mapping. **No subtype-specific code in soda-core.** The first consumer is `DataStandardImpl` in soda-extensions (#396), but the seam is designed for arbitrary future subtypes (reconciliation, …).

Three opt-ins, each independent:

1. **`combine_uploads`** — class attribute. When `True`, N files of a single subtype in one session land in one `sodaCoreInsertScanResults` request instead of N. Default `False`; contract path is byte-identical.
2. **`requires_collection_id` + `(wire_source, collection_id)` uniqueness** — for `combine_uploads = True` subtypes that require an identifier, two files in the same session may not share the pair. Hard-reject with `InvalidArgumentException` before any `verify()` runs.
3. **`expected_kinds`** (executor kwarg) — opt-in allowlist of permitted top-level `kind:` values. Subtype-narrowed public APIs pin their dispatch with this; the universal entrypoint leaves it `None` and accepts every registered kind. Reuses the single phase-1 kind-dispatch parse — one `yaml_source.parse()` per yaml per session.

Plus session-result rollups lifted to a base `CheckCollectionSessionResult` and a module-level `session_result_to_exit_code` CLI helper, so every subtype CLI inherits the same exit-code mapping for free.

### Combined uploads
- New class attribute `CheckCollectionImpl.combine_uploads: bool` (default `False`).
- When `True`, `verify()` still uploads the YAML file and runs the alignment guard, then skips `send_contract_result` — the executor issues one combined upload at the session boundary.
- New `SodaCloud.send_combined_results(results, wire_source, scan_definition_suffix)` + module helper `_build_combined_results_json_dict(...)`. Combined payload mirrors the per-file shape: session-level fields come from the first result, `hasErrors`/`hasWarns`/`hasFailures` are ORs over per-file values, `checks` and `logs` flatten, `postProcessingStages` are de-duped by name, ingestion mode is `PARTIAL` if any check was excluded. `contract` stays null for non-contract sources.
- `verify()` stamps per-file `LogRecord.thread = \"{wire_source}.{collection_id}\"` at the end of method (after all in-method logging) so combined uploads can group log records by file via the wire's `thread` field.

### Executor: three explicit phases
`execute_check_collections` is now `phase 1 (construct-all) → phase 1.25 (expected_kinds filter) → phase 1.5 (session invariants) → phase 2 (verify-all) → phase 3 (combined uploads + post-processing handlers)`. Per-file isolation preserved on phases 1 and 2 (construct + verify failures become ERROR placeholders identical to today). `abort_on_first_error` semantics unchanged.

### `expected_kinds` filter
Optional `set[str]` argument to `execute_check_collections`. When set, the executor reads each yaml's `kind:` during the existing phase-1 dispatch parse and collects offenders; raises `InvalidArgumentException` after the construct loop, before any verify() runs. Subtype-narrowed APIs (`verify_data_standards`, future `verify_reconciliations`) pass their own `{kind}` so dispatch matches the function name's promise and yamls are parsed exactly once per session. `None` (default) accepts every registered kind — mixed-kinds sessions go through the universal executor unchanged.

### Phase 1.5: duplicate-`collection_id` validation
For impls in a `combine_uploads = True` subtype with a non-empty `collection_id`, `(wire_source, collection_id)` must be unique across the session. Duplicates raise `InvalidArgumentException` with a multi-line message listing every duplicate group and file paths. Unconditional raise (ignores `abort_on_first_error` — user-input error, not engine failure).

### Session-result rollups + exit-code helper
- `CheckCollectionSessionResult` gains rollup properties (`has_errors`, `is_failed`, `is_warned`, `sending_results_to_soda_cloud_failed`) computed across `results`. Subtypes get the rollups by inheriting.
- New module `soda_core/cli/exit_codes.py` houses `session_result_to_exit_code(_HasRollups) -> ExitCode` — subtype-neutral CLI exit-code mapping via a structural Protocol. Every subtype CLI uses the same priority order without duplicating logic.

### Misc
- `build_data_source_yaml_sources` promoted to public API in `soda_core.common.yaml` (used by contract verify-api and any future subtype CLI).
- First-class fields on `CheckCollectionYaml` base: `yaml_source`, `yaml_object`, `kind`, `data_timestamp`, `execution_timestamp` — subtypes inherit instead of duplicating the boilerplate.

## Tests

- **13 unit tests** in `test_combined_results_wire.py` cover the combined payload shape (check flattening, OR-aggregated status flags, min/max timestamps, contract-field omission, post-processing dedup, ingestion mode aggregation, scan-def suffix application) and the `send_combined_results` API (empty-list no-op, scan_id stamping on 200 OK, mark-all-failed on non-200 / missing scanId).
- **8 unit tests** in `test_dup_collection_id_validation.py` cover the phase 1.5 dup check.
- **5 unit tests** in `test_expected_kinds.py` cover the phase 1.25 filter (single offender, missing-kind default, `None` accepts all, multiple allowed kinds, multi-offender single raise).
- **5 unit tests** in `test_exit_codes.py` lock the precedence order of `session_result_to_exit_code` across multi-flag combinations.
- **Unit suite total: 853 passed, 2 skipped.**

## Backend coordination — none anticipated

Combined payload sets `contract: null` for non-contract sources (existing behaviour). Per-check resolution on the backend already happens via `firstSegmentOf(checkPath)` per check, so 1 vs N files of checks in `checks[]` is transparent.

Empirical pre-merge check: confirm one `sodaCoreInsertScanResults` request creates exactly one Scan row (not one per `checkPath` segment).

## Test plan
- [ ] CI: lint + unit tests pass
- [ ] Empirical: 2-file end-to-end verify against dev produces exactly one Scan row


m1n0:
**Follow-up: duplicate collection_id validation**

New commits on top of the combined-upload work:
- `refactor: split execute_check_collections into parse+construct / verify phases`
- `refactor: carry yaml_source in constructed tuple + restore why-comments`
- `feat: reject duplicate collection_id in combine-upload subtypes`
- `test: add edge-case coverage for dup collection_id validation`
- `chore: code-review polish on dup collection_id check`

Spec: `docs/superpowers/specs/data_standards/2026-05-25-unique-collection-id-validation-design.md`
Plan: `docs/superpowers/specs/data_standards/2026-05-25-unique-collection-id-validation-plan.md`

The refactor splits the executor into three explicit phases (construct-all → session-invariant checks → verify-all). The new phase 1.5 raises `InvalidArgumentException` when two yamls in one session produce impls sharing the same `(wire_source, collection_id)` inside a `combine_uploads=True` subtype, listing every duplicate group with file paths. Hard-rejects the session before any verify runs — no partial work.

Unit suite: 843 passed (8 new in `test_dup_collection_id_validation.py`).

soda-core/src/soda_core/check_collections/session.py
```diff
@@ -165,7 +253,68 @@ def execute_check_collections(
...
+    # Post-processing handlers — always run for every successfully-constructed
+    # combine-upload impl. The non-combine path runs handlers unconditionally
+    # at the end of ``verify()``; this loop is the combine-upload equivalent.
+    for (impl, impl_class, _, _), result in zip(constructed, results):
+        if impl is None or impl_class is None or not impl_class.combine_uploads:
+            continue
+        response_for_file = (
+            response_json_by_wire_source.get(impl_class.wire_source) if id(result) in uploaded_ids else None
+        )
+        impl.run_post_processing_handlers(result, response_for_file)
```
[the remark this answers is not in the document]
m1n0: Good catch — bug confirmed. Phase 3 iterated every successfully-constructed impl regardless of whether verify() raised, but the non-combine path's handlers live inside verify() so an exception bypasses them. Now skipped via `result.error is not None` guard with a dedicated regression test (`test_phase3_handler_skip_on_error.py`). Fixed in 1737da79.

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -793,41 +875,81 @@ def verify(self) -> CheckCollectionResult:
...
         verbose: bool,
     ) -> CheckCollectionResult:
-        """Agent-path verification. Default: raise NotImplementedError.
+        """Runner-path verification. Default: raise NotImplementedError.
 
-        Subtypes that support agent execution (``ContractImpl``) override.
+        Subtypes that support remote execution (``ContractImpl``) override.
         """
-        raise NotImplementedError(f"{self.display_name} does not support agent execution")
+        raise NotImplementedError(f"{self.display_name} does not support runner execution")
```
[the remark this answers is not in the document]
m1n0: Intentionally not adding the alias on the base. The base `verify_on_runner` only raises NotImplementedError — no external caller is hitting it (no real subtype that doesn't override it exists), so swapping AttributeError for NotImplementedError has no observable difference. The legitimate alias on `ContractImpl` is preserved for actual external callers. A spurious base-class alias was previously added and then removed deliberately in 6d6bafcf as scope creep relative to this PR's stated work.

soda-core/src/soda_core/cli/exit_codes.py
```diff
@@ -15,3 +16,49 @@ class ExitCode(IntEnum):
...
+        ...
+
+
+def session_result_to_exit_code(session_result: _HasRollups) -> ExitCode:
+    """Map a session result's rollup properties to a CLI exit code.
+
+    Subtype-neutral: works on any session-style result that exposes the
+    four rollup properties (``CheckCollectionSessionResult`` and its
+    subtypes do; ``ContractVerificationSessionResult`` does too, with
+    its own equivalents). Priority order:
```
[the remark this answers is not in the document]
m1n0: Correct — `ContractVerificationSessionResult` is the legacy session result, not a structural match (no `sending_results_to_soda_cloud_failed`). Docstring narrowed in 1737da79 to call out the actual covered types (`CheckCollectionSessionResult` + subtypes) and to explicitly note the legacy contract session result is NOT covered.

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -793,41 +875,81 @@ def verify(self) -> CheckCollectionResult:
...
                     dwh_files=self.dwh_files,
                 )
             except Exception as e:
                 logger.error(f"Error in {self.display_name} verification handler: {e}", exc_info=True)
                 self._handle_post_processing_failure(scan_id=scan_id, exc=e, contract_verification_handler=handler)
 
-        return verification_result
-
-    def verify_on_agent(
+    def verify_on_runner(
```
Niels-b, line 962: Missing a return here. Is this intentional?

Niels-b: Also; there is no fallback for the `verify_on_agent` call. This should be added to keep backwards compatibility

soda-core/src/soda_core/check_collections/session.py
```diff
@@ -109,31 +123,48 @@ def execute_check_collections(
...
-            # call); the subtype's full parse below repeats this work but
-            # also walks columns/checks/variables which is the heavy part.
+            # Parse the YAML once for kind dispatch; reuse the parsed
+            # object inside the subtype's ``yaml_class.parse(...)`` so the
+            # subtype's __init__ doesn't re-parse the same source.
+            # ``YamlSource.parse()`` either returns a ``YamlObject`` or
+            # raises; never ``None``.
             yaml_object = yaml_source.parse()
-            kind = (yaml_object.read_string_opt("kind") if yaml_object is not None else None) or "contract"
+            kind = yaml_object.read_string_opt("kind") or "contract"
```
Niels-b: Claude is pointing out that for missing files, the default will be `contract`. I believe you said this was the intentional default if nothing else has been specified? Just making sure.

Comment from Claude: 
kind = yaml_object.read_string_opt("kind") or "contract" defaults to "contract" before the offender check. The test test_expected_kinds_rejects_yaml_missing_kind locks this — but the user-facing message says kind='contract' for a file that actually had no kind: field at all. Minor UX issue; consider keeping the raw None for the offender report so the error reads kind=None (or "missing").

soda-tests/tests/unit/test_combined_results_wire.py
```diff
@@ -0,0 +1,254 @@
...
+    )
+    r2 = _make_result(
+        label="beta",
+        post_processing_stages=[
+            PostProcessingStage(name="dwh-stage", stage=PostProcessingStageState.ONGOING),
+            PostProcessingStage(name="other-stage", stage=PostProcessingStageState.ONGOING),
+        ],
+    )
+    payload = _build_combined_results_json_dict([r1, r2], wire_source="data-standard")
+    assert payload["postProcessingStages"] == [{"name": "dwh-stage"}, {"name": "other-stage"}]
```
Niels-b, line 169: Claude: 
Tests instantiate PostProcessingStage(name="dwh-stage", stage=PostProcessingStageState.ONGOING) with kwarg stage, but _handle_post_processing_failure in base.py uses state=PostProcessingStageState.FAILED. Worth verifying these aren't both accepted only because the dataclass tolerates extras — if one is wrong the tests still pass while production is broken.

soda-core/src/soda_core/check_collections/base.py
```diff
@@ -793,41 +875,81 @@ def verify(self) -> CheckCollectionResult:
...
+
+        # Override OS thread id with a per-file label so combined uploads
+        # group log records by file via the wire's ``thread`` field. Done
+        # at the end of verify() (not at pop_log_records time): pop returns
+        # the live list and code after it appends more records to the same
+        # list (YAML upload, handlers).
+        if log_records:
+            thread_label = f"{self.wire_source}.{self.collection_id}" if self.collection_id else self.wire_source
+            for record in log_records:
+                record.thread = thread_label
```
Niels-b: I'm missing the context here, so going to put the full comment from Claude here. This was marked as a bug:

For combine subtypes, handlers run in phase 3 after verify() already relabelled the records. Any log records emitted by handlers in phase 3 won't carry the per-file thread label — which is exactly the grouping signal the combined upload relies on. The non-combine path is fine because handlers run before the relabel inside verify(). Either relabel at the end of phase 3 too, or have run_post_processing_handlers re-apply the label on its own appended records.

Niels-b (approved):
Skimmed through this PR + agentic review. Idea looks good to me. Still some comments to address

soda-core/src/soda_core/check_collections/base.py
m1n0: Intentional — `raise NotImplementedError(...)` is the return mechanism here. The method has no successful-return path on the base; any subtype that needs runner support overrides (only `ContractImpl` does today). A return statement after the raise would be unreachable.

m1n0: Declining the BC alias on the base, same reasoning I posted to Copilot's earlier comment on this line: the base only raises NotImplementedError, no external caller is hitting the base `verify_on_*` path (no real subtype that doesn't override exists), so swapping AttributeError for NotImplementedError has no observable difference. `ContractImpl` keeps its alias where actual external callers live. The user explicitly removed a spurious base-class alias as scope creep in `6d6bafcf` so I'm holding that line.

soda-core/src/soda_core/check_collections/session.py
m1n0: Good catch — that was a real wart. The defaulting to `"contract"` is BC for legacy contract yamls without a kind line, but reporting `kind='contract'` for a file that genuinely had no `kind:` field is misleading. Fixed in 46a9feb9: `raw_kind` stays `None` for the offender report; the `"contract"` default is now applied only for impl-class dispatch. Test updated to assert `kind=None` for missing-kind.

soda-tests/tests/unit/test_combined_results_wire.py
m1n0: Double-checked — this is a false positive. The tests use `PostProcessingStage(name=..., stage=...)` which matches the dataclass field name (`contract_verification.py:533`: `__init__(self, name: str, stage: PostProcessingStageState, ...)`). `_handle_post_processing_failure` uses `self.soda_cloud.post_processing_update(stage=<name>, state=...)` — that's a different method on `SodaCloud` whose parameters happen to be named `stage` (the stage's name string) and `state` (the state enum). Two unrelated APIs that share a word; nothing for the dataclass to silently swallow.

soda-core/src/soda_core/check_collections/base.py
m1n0: Real bug — confirmed. Fixed in 46a9feb9: extracted `_apply_thread_label_to_log_records` and call it at the end of `run_post_processing_handlers` too. So:

- Non-combine: relabel inside verify() (existing) catches query/upload/in-verify-handler emissions; relabel inside run_post_processing_handlers (new) is idempotent on the same set.
- Combine: relabel inside verify() catches query/upload emissions; phase 3 invokes run_post_processing_handlers AFTER verify() returned, and its tail relabel pass catches any handler-emitted records.

Regression test in soda-extensions PR — `test_run_post_processing_handlers_relabels_records_emitted_inside` appends a raw-thread-id record and asserts the relabel kicks in.

PR #2725, opened 2026-05-29
refactor: unify single + multi-file send/build into one path (DES-348 follow-up)

m1n0:
## Summary

Stacked on top of #2719. Collapses ``send_contract_result`` + ``send_combined_results`` into a single ``send_check_collection_results(results: list, ...)`` and the two ``_build_*_json_dict`` helpers into one ``_build_check_collection_results_json_dict``. Contracts and combine-upload subtypes (data-standards) now flow through the same wire-building + sending code; the only divergence — the conditional ``contract`` field — moves into the unified function and stays gated on ``wire_source == "soda-contract"``.

The combined builder's aggregation operators degenerate to the single-result behaviour for N=1 (min/max over [t] is t; any() over [b] is b; flatten [r] is r) — so the unification doesn't change semantics on either path, just removes the duplicate code.

Companion: [soda-extensions PR](https://github.com/sodadata/soda-extensions/pull/407)

## API changes

- **New** ``SodaCloud.send_check_collection_results(results: list, wire_source=\"soda-contract\", scan_definition_suffix=None)``. Stamps ``scan_id`` and per-result ``dataset_id`` on the result list internally on 200; flips ``sending_results_to_soda_cloud_failed`` on every result on non-200 or missing ``scan_id``. Empty list is a no-op.
- **New** ``_build_check_collection_results_json_dict(results: list, wire_source, scan_definition_suffix)``.
- **Removed** ``SodaCloud.send_contract_result``, ``SodaCloud.send_combined_results``, ``_build_contract_result_json_dict``, ``_build_combined_results_json_dict``.

## Callers migrated

- ``CheckCollectionImpl.verify()`` (non-combine path) now calls the unified method with ``[verification_result]``. The post-call inline ``scan_id``/``dataset_id`` stamping is gone (internal now).
- ``execute_check_collections`` (combine-upload path) calls the unified method with the per-wire-source group. Behaviour identical.
- ``soda_autopilot/publish.py`` (extensions) wraps its single result in a list.

## Tests

- New ``test_single_result_wire_equivalence.py`` (7 tests) locks the N=1 contract:
  - ``contract`` field present iff ``wire_source == \"soda-contract\"``; ``None`` otherwise (and thus stripped by ``to_jsonnable``)
  - Session-level fields (timestamps, ``has_*``, ``dataTimestamp``, ``defaultDataSource``, ``definitionName``) come from the single result
  - ``checks`` list equals the single result's checks
  - ``postProcessingStages`` passes through (dedup is a no-op for unique names)
  - ``resultsIngestionMode`` is ``PARTIAL`` iff any check is EXCLUDED, else ``FULL``
- Existing combined-result tests updated to use the unified name + 1-element list call shape.
- Suite: **868 passed, 2 skipped** (was 861; 7 new).

## Test plan

- [ ] CI: lint + unit tests pass in both repos
- [ ] Empirical: single contract verify against dev still produces the byte-identical wire payload (key: ``contract`` field present, ``checkPath`` per check, ``scanId``/``dataset_id`` resolved on the result)

### DIFF
--- a/soda-core/src/soda_core/check_collections/base.py
+++ b/soda-core/src/soda_core/check_collections/base.py
@@ -698,7 +698,6 @@ class CheckCollectionImpl:
         return columns
 
     def verify(self) -> CheckCollectionResult:
-        from soda_core.common.soda_cloud import extract_dataset_id_from_response
         from soda_core.contracts.impl.contract_verification_impl import (
             ContractVerificationHandlerRegistry,
             DerivedMetricImpl,
@@ -864,20 +863,14 @@ class CheckCollectionImpl:
                 # Session-level combined upload — executor sends after the loop.
                 logger.debug(f"Deferring upload to session-level combined request " f"{Emoticons.FINGERS_CROSSED}")
             else:
-                # send_contract_result will use contract.source.soda_cloud_file_id
-                soda_cloud_response_json = self.soda_cloud.send_contract_result(
-                    verification_result,
+                # send_check_collection_results stamps scan_id + dataset_id
+                # on the result internally; we just hold the response_json
+                # for the post-processing-handler invocation below.
+                soda_cloud_response_json = self.soda_cloud.send_check_collection_results(
+                    [verification_result],
                     wire_source=self.wire_source,
                     scan_definition_suffix=type(self).scan_definition_suffix,
                 )
-                scan_id = soda_cloud_response_json.get("scanId") if soda_cloud_response_json else None
-                if not scan_id:
-                    verification_result.sending_results_to_soda_cloud_failed = True
-                else:
-                    verification_result.scan_id = scan_id
-                    verification_result.check_collection.dataset_id = extract_dataset_id_from_response(
-                        soda_cloud_response_json, self.soda_qualified_dataset_name
-                    )
         else:
             logger.debug(f"Not sending results to Soda Cloud {Emoticons.CROSS_MARK}")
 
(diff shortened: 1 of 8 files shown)


Niels-b (approved):
LGTM. I like the unification of this. Nice improvement 👏 !

Some (minor) comments from Claude:
  - Success-log format change for N=1. "Results sent to Soda Cloud" → "Results sent to Soda Cloud (1 file(s))". Harmless, but if anything greps for the exact old string it'll miss. The "file(s)" plural is also a small style nit — f"{len(results)} file" + ("s" if len(results) != 1 else "") reads cleaner, though not worth churn.                                            
  - No explicit N=1 stamping test. The renamed test_send_check_collection_results_stamps_scan_id_on_every_result uses two results. A 1-element variant would directly assert the migrated path (the autopilot caller and the verify() non-combine path both hit N=1).

m1n0:
Both addressed in 6aa04852:

- **Pluralization**: changed to `"1 file"` / `"N files"` conditional. Grepped the whole product (soda-core, soda-extensions, soda-library, soda, agent-orchestrator) for `"Results sent to Soda Cloud"` — nothing matches the old string outside the source file itself, so the wording is safe to evolve.

- **N=1 stamping test**: added `test_send_check_collection_results_n1_stamps_scan_id_and_dataset_id`. Good call — the pre-existing 2-element stamping test covered only scan_id; the dataset_id resolution moved INTO the unified sender during this unification, so the N=1 case wasn't directly asserted. The new test exercises the migrated path (autopilot + non-combine verify()) end-to-end with both scan_id and dataset_id assertions.

Suite: 869 / 2 skip.

PR #2727, opened 2026-06-01
chore(deps): bump urllib3 to 2.7.0 (security)

m1n0:
## What
Bump the transitive `urllib3` dependency **2.6.3 → 2.7.0** in `uv.lock` to resolve a security vulnerability.

## Why
`urllib3` is pulled in transitively (via `requests`, `botocore`, `databricks-sql-connector`). 2.7.0 is the patched release.

## How
`uv lock --upgrade-package urllib3` off latest `main`. Only the `urllib3` entry changes (version + sdist/wheel hashes); no other packages move.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2597,11 +2597,11 @@ wheels = [
 
 [[package]]
 name = "urllib3"
-version = "2.6.3"
+version = "2.7.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


Niels-b (commented):
Do we need a [REPLAY=OFF] run for this just to be sure? (It will also happen on main)

Niels-b (approved):

PR #2730, opened 2026-06-01
feat(sparkdf): catalog mode + Spark Connect modes for Databricks DWH

Niels-b:
## Summary

- Adds an opt-in ``use_catalog`` flag to the SparkDF dialect so DWH writes can land in a Unity-Catalog 3-level namespace (``catalog.schema.table``). Default off — local SparkDF behaviour unchanged.
- Adds two new YAML-friendly SparkDF connection modes:
  - ``remote_session`` (host / token / cluster_id) builds a Spark Connect session against a Databricks all-purpose cluster.
  - ``active_session`` (``use_active_session: true``) picks up ``SparkSession.getActiveSession()`` at connect time — no credentials in YAML, ideal for notebooks.
- ``token`` is a Pydantic ``SecretStr`` so the PAT no longer leaks into soda-core's DEBUG log of ``connection_properties`` under verbose logging.

Together these let Soda's Diagnostics Warehouse run end-to-end on Databricks via a single Spark session, with both source reads and DWH writes flowing through ``spark.sql(...)``. The companion PR in soda-extensions wires this into the DWH prefix builder and adds a live integration test against a real cluster.

## Test plan

- [x] ``soda-sparkdf/tests/unit/`` — 12 unit tests covering both modes (legacy + catalog), DDL generation, prefix indexing, token masking, active-session discriminator, and conditional ``SUPPORTS_DROP_TABLE_CASCADE``
- [x] ``test_dwh_sparkdf.py`` (legacy local-Spark DWH) still passes — no regression
- [x] Live integration test against a real Databricks cluster (in the soda-extensions companion PR): in-source + between-source (parametrised over ``remote_session`` and ``active_session`` DWH YAML modes)

### DIFF
--- a/soda-sparkdf/pyproject.toml
+++ b/soda-sparkdf/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.12.1rc0",
     "freezegun",
-    "pyspark>=3.5.0",
+    "pyspark[connect]>=3.5.0",
     "soda-databricks==4.12.1rc0",
 ]
 
--- a/soda-sparkdf/src/soda_sparkdf/common/data_sources/sparkdf_data_source.py
+++ b/soda-sparkdf/src/soda_sparkdf/common/data_sources/sparkdf_data_source.py
@@ -161,10 +161,16 @@ class SparkDataFrameSqlDialect(DatabricksSqlDialect, sqlglot_dialect="spark"):
 
     def create_schema_if_not_exists_sql(self, prefixes: list[str], add_semicolon: bool = True) -> str:
         if self.use_catalog:
+            if len(prefixes) < 2:
+                raise ValueError(
+                    f"Catalog-mode SparkDF requires 2 prefixes [catalog, schema]; got {len(prefixes)}: {prefixes}"
+                )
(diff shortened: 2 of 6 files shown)


m1n0 (approved):
LGTM overall, two issues flagged by claude that are I think worth exploring:

packaging:
```
remote_session calls SparkSession.builder.remote(uri) → needs the
  Spark Connect stack (grpcio, grpcio-status, googleapis-common-protos, pyarrow), which ships only via pyspark[connect].
  soda-sparkdf/pyproject.toml pins plain pyspark>=3.5.0. CI passes only because grpcio is present transitively — a clean pip install
  hits ImportError at connect time. Fix: pyspark[connect]>=3.5.0 or explicit deps, in #2730.
```

more of a NIT, but indeed very confusing error messages would happen if user uses `https://...` url format
```
conflicting config (use_active_session: true + host) is silently resolved by discriminator order rather
  than rejected; remote_session hard-codes :443/use_ssl=true so a pasted https://... host gives an opaque gRPC error; #413's live test
  leaves per-run schemas un-dropped in the live catalog (sibling test does _drop_dwh_schema).
```

PR #2737, opened 2026-06-04
PLATL-703: warn when account_info_json is provided alongside use_context_auth

Niels-b:
## Summary

The Soda Cloud UI work for PLATL-703 carries both auth modes' input fields in a single BigQuery connection form. Users can type a service-account JSON value, then switch to Application Default Credentials. Under that toggle the FE hides the JSON field and omits the value from the YAML — but if either the FE guard breaks or the YAML is hand-edited, the value would reach soda-core and be silently ignored, which is confusing.

This PR makes soda-core surface that situation:

- `BigQueryContextAuth` gains an explicit `Optional[SecretStr]` `account_info_json` field so the value is parsed and exposed on the model (previously it was dropped as a Pydantic extra), and the docstring documents the round-trip-for-form-state intent.
- `_load_project_id_and_credentials` logs a warning when both `use_context_auth` and a non-empty `account_info_json` are present, then proceeds with `default()` exactly as before.

## Tests

- New `soda-bigquery/tests/unit/test_bigquery_context_auth_logging.py` — caplog-based unit tests cover the warning path (ADC + JSON) and the no-warning path (ADC alone), with `google.auth.default()` monkeypatched.
- `soda-bigquery/tests/data_sources/test_bigquery.py` — new `app_default_creds_with_optional_json` `TestConnection` next to the existing `app_default_creds` case, pinning that the YAML parses cleanly under the new variant (same env caveat: requires no GCP credentials in the test environment).

## Cross-repo

Paired with `sodadata/soda` PR on the matching branch name `platl-703-show-bigquery-wif-settings-on-ui` (cross-repo branch-name convention). The Soda webapp also hides the Account Information JSON field under ADC and omits the value from the YAML serializer; this PR is defence-in-depth for the case where that guard isn't in effect.

## Test plan

- [x] `pytest tests/unit/test_bigquery_context_auth_logging.py tests/data_sources/test_bigquery.py -k 'context_auth_logging or app_default_creds' -v` — 4/4 pass locally.
- [ ] CI green.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
@@ -91,7 +91,7 @@ class BigQueryDataSourceConnection(DataSourceConnection):
     def _load_project_id_and_credentials(self, config: BigQueryConnectionProperties):
         if isinstance(config, BigQueryContextAuth):
             if config.account_info_json and config.account_info_json.get_secret_value():
-                logger.warning("account_info_json was provided but is ignored because use_context_auth is enabled.")
+                logger.info("account_info_json was provided but is ignored because use_context_auth is enabled.")
             logger.info("Using application default credentials.")
             self.credentials, self.project_id = default()
             return
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
@@ -62,10 +62,16 @@ class BigQueryContextAuth(BigQueryConnectionProperties):
     """BigQuery authentication using context.
 
     If use_context_auth is True, then application default credentials will be used.
-    The user may optionally provide JSON credentials; they will be ignored.
+    account_info_json is accepted (so a single connection form can carry both auth modes
+    without losing the user's input when toggling) but is ignored at credential resolution
+    time; a warning is logged when both are present.
     """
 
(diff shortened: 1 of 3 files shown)


m1n0 (approved):

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source_connection.py
```diff
@@ -62,10 +62,16 @@ class BigQueryContextAuth(BigQueryConnectionProperties):
...
-    The user may optionally provide JSON credentials; they will be ignored.
+    account_info_json is accepted (so a single connection form can carry both auth modes
+    without losing the user's input when toggling) but is ignored at credential resolution
+    time; a warning is logged when both are present.
     """
 
     use_context_auth: Literal[True] = Field(description=CONTEXT_AUTHENTICATION_DESCRIPTION)
+    account_info_json: Optional[SecretStr] = Field(
+        None,
+        description="Ignored when use_context_auth is True; accepted to round-trip form state",
```
mivds: Hmm, isn't our default to error out if there are unexpected parameters? Or do we accept & ignore unknown extra args? (in context of existing runner/core versions out there)

Also, what is the round-trip form state about? We don't send the incoming YAML anywhere as far as I know (so just a 1-way trip?)

Niels-b: There is a `model_config = ConfigDict(frozen=True, extra="allow", validate_by_name=True)` in the `DataSourceConnectionProperties`. So we allow for extra parameters to be passed (backwards compatible ✅ ). 

That round trip is a bit too much doc information 😬 . My assistant (Claude) was very very eager to commit and push this 😅 . I'll update it

PR #2739, opened 2026-06-07
fix: correctly parse migration poll response; use dict not Response a…

reachsridhard:
## Description

- Fixes a misuse of `Response` vs `dict` and an incorrect logical condition in [SodaCloud.migration_poll_status()](cci:1://file:///Users/satyasivasundarsalagrama/open-source/soda-core/soda-core/src/soda_core/common/soda_cloud.py:1281:4-1323:46) when parsing the migration status response from Soda Cloud.
- Previously, the code incorrectly accessed `response["migration"]` (treating the `Response` object as a dict) and used an `and` condition that could let through malformed JSON. This could cause runtime errors or incorrect polling behavior.

### Problem being solved
- After parsing the JSON response, the code checked `"state" not in response["migration"]` instead of using the parsed `response_body_dict`.
- The logical condition used `and` instead of `or`, which could allow the code to proceed even when keys were missing.

### Expected impact on downstream packages/services
- Low risk. This only affects the migration status polling logic on the error path (malformed or incomplete responses).
- The fix makes the parsing robust by consistently using the parsed dict and properly validating required keys.
- No CLI changes, no external API changes, and no plugin interfaces changed.

## Changes

- In [soda-core/src/soda_core/common/soda_cloud.py](cci:7://file:///Users/satyasivasundarsalagrama/open-source/soda-core/soda-core/src/soda_core/common/soda_cloud.py:0:0-0:0), method [migration_poll_status](cci:1://file:///Users/satyasivasundarsalagrama/open-source/soda-core/soda-core/src/soda_core/common/soda_cloud.py:1281:4-1323:46):
  - Changed the response validation to consistently use `response_body_dict` (the parsed JSON dict).
  - Updated the logical condition to use `or` so that any missing key is treated as a parse failure.
  - Ensures `response_body_dict` is not `None` before checking keys.

Before (problematic):
```python
if "migration" not in response_body_dict and "state" not in response["migration"]:


reachsridhard:
@m1n0 Could you please take another look when you have a chance?

I’d appreciate your feedback and review. Thanks for your time!

reachsridhard:
@Niels-b ,

 Could you please take another look when you have a chance?
I’d appreciate your feedback and review. Thanks for your time!

PR #2751, opened 2026-06-13
Fix decorator AST visitor dispatch

wunianze666-netizen:
﻿## Summary
- fix the `ast.NodeVisitor` callback registration so `FunctionDef` nodes are actually visited
- route decorator-argument diagnostics through structured debug logging instead of stdout
- add a regression test that exercises a real decorator call with an AST argument

## Why
`NodeVisitor` dispatches function definitions to `visit_FunctionDef`, but the helper registered `visit_function_def`. As a result, `get_decorators()` returned an empty mapping and the intended visitor logic never ran. Correcting the dispatch exposes the existing unconditional `print()`, so the diagnostic is now kept behind the project logger.

## Validation
- `py -m uv run --package soda-tests --group dev pytest soda-tests/tests/unit/test_soda_tracer.py -q` (`1 passed`)
- `py -m uv run black --check soda-tests/tests/unit/test_soda_tracer.py`
- `git diff --check`

## CI note
The GitHub Actions matrix currently stops before test execution in the upstream Get external secrets step because fork runs do not receive the AWS region/Secrets Manager configuration (expected region to be configured for aws.auth#sigv4). The fork-safe pre-commit, lockfile, CLA, Aikido, and SonarCloud checks pass; the targeted unit test above passes locally.

### DIFF
--- a/soda-core/src/soda_core/telemetry/soda_tracer.py
+++ b/soda-core/src/soda_core/telemetry/soda_tracer.py
@@ -36,7 +36,7 @@ def get_decorators(function):
                 name = n.func.attr if isinstance(n.func, ast.Attribute) else n.func.id
 
                 for a in n.args:
-                    print(ast.dump(a))
+                    logger.debug("Decorator arg AST: %s", ast.dump(a))
             else:
                 group = n.attr if isinstance(n, ast.Attribute) else n.id
                 name = None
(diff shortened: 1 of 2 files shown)


wunianze666-netizen:
CI follow-up: all 13 failing Python 3.14 matrix jobs stop in the `Get external secrets` step before tests run. The AWS Secrets Manager action receives empty fork credentials/region and reports `expected region to be configured for aws.auth#sigv4`. The pre-commit/lockfile checks and Sonar quality gate pass, so this does not appear to be a source-code failure that can be fixed in the PR branch. Could a maintainer rerun the matrix in a trusted context or apply the repository's fork-safe workflow path?

PR #2756, opened 2026-06-15
fix(cli): print the ASCII banner to stderr (interactive only), never stdout

m1n0:
The logo was printed to **stdout** at the top of `execute()` on every invocation, before argparse — contaminating piped/redirected stdout and any `--json` consumer (`soda … --json | jq` fails on the banner).

It is a cosmetic, interactive-only nicety, so this routes it to **stderr** and only when stderr is a TTY. `SODA_NO_BANNER=1` silences it entirely. stdout is now clean for machine consumption; the logo still shows in an interactive terminal.

No behavior change for interactive use; no test asserted the banner on stdout.

### DIFF
--- a/soda-core/src/soda_core/cli/cli.py
+++ b/soda-core/src/soda_core/cli/cli.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import os
 import signal
 import sys
 import traceback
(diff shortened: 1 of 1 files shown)


m1n0:
Closing — this will be reworked on a separate branch; not landing here for now.

PR #2757, opened 2026-06-15
r-594 Memory management: DWH streaming reads, copy-free value sizing & memory_container harness (core)

Niels-b:
## Summary

Hardens the diagnostics-warehouse (DWH) failed-rows pipeline against out-of-memory failures under tight Kubernetes pod caps, and adds a containerized memory-regression test harness to measure and gate peak memory. This is the **soda-core** half of a cross-repo change — the consumer-side DWH transfer logic lives in the matching `soda-extensions` branch of the same name.

The core problem: on the DWH transfer path, reading failed rows from the source with psycopg3's default (client-side) cursor materializes the *entire* result set in client memory before the first row is processed, driving peak RSS to several times the source size and OOM-killing pods. This branch introduces a bounded streaming read and copy-free value sizing so peak memory stays roughly flat regardless of result-set size.

## Core engine

- **Memory-optimized streaming reads.** `execute_query_one_by_one` is split into the proven buffered base plus a new `execute_query_one_by_one_memory_optimized` hook (`common/data_source_connection.py`). Postgres overrides it with a server-side named cursor (`withhold=True`, so it survives the per-flush commits the DWH writer makes on the same connection) and byte-budgeted `fetchmany` (8 MiB budget, adaptive batch size), bounding peak to ~one batch plus the widest row. Includes held-cursor cleanup on error and an autocommit fallback. (`soda-postgres/.../postgres_data_source_connection.py`)
- **Copy-free value sizing** (`common/value_size.py`, new). A budgeted recursive walker that estimates a value's in-memory size — including nested containers and parsed jsonb — without serializing it (which would allocate the very bytes the budgets exist to avoid), and counts `memoryview` buffers correctly. Used by the read-side fetch batching, the DWH buffer caps, and insert routing.
- **AST refcount fix** (`common/sql_ast.py`). `parent_node` back-references are now weakrefs, breaking a reference cycle with `INSERT_INTO` that delayed GC of large row graphs.

## Adapter statement-length caps

- Redshift: cap max SQL statement length at 15 MB.
- Databricks: 15 MiB statement cap; disable `DROP TABLE ... CASCADE` (unsupported by the dialect).

## Memory-regression test harness (`memory_container`)

- A pytest plugin (`soda-tests/src/helpers/memory_container_plugin.py`) re-dispatches `@pytest.mark.memory_container` tests into a Docker container with a cgroup memory cap, measures peak (kernel `memory.peak` plus pollers), and integrates memray for per-allocation analysis. A `setup_outside` hook runs heavy fixture creation on the host, outside the measured cap.
- **Local-only by hard guard:** the plugin force-skips these tests under `CI`/`GITHUB_ACTIONS`, and CI never collects the `memory_container` directories — memory tests never run in CI (verified: path-scoping + plugin guard).
- **Peak regression gate:** per-datasource `memory_baselines.json` with a tolerance band, measurement-sanity bands, and stale-baseline rename detection.
- **Docker images:** a canonical image (postgres/sqlserver/snowflake/bigquery/databricks + the group_by and reconciliation check-type plugins) and a separate `-extra` image (duckdb/redshift/synapse/fabric/athena/sparkdf/oracle) — kept apart so added adapter import floors don't shift recorded baselines. Credentials are kept out of image layers via `.dockerignore`, and env vars reach the container through a 0600 `--env-file`.

## Testing

- **Unit:** value sizing, peak gate, CI guard, plugin hooks, memory-optimized query batching, inner-summary parsing, fixture row-count validation.
- **Integration:** unbounded-text-column round-trip + length verification.
- **Real-DB (postgres):** streaming-read correctness across thin/fat row mixes, withhold-cursor survival across a mid-stream commit, autocommit fallback, and the documented rollback-before-held limitation.
- **Memory tiers:** DWH workloads across fat-row / scaling / FRQ / group-by shapes, run locally (gated by `SODA_MEMORY_TESTS=true`).

## Notes

- Memory tests are **local-only** and never execute in CI.
- Pairs with the same-named `soda-extensions` branch (the DWH transfer pipeline that consumes these primitives).


soda-tests/src/helpers/memory_container_plugin.py
```diff
@@ -0,0 +1,1528 @@
...
+    if outcome == "passed" and strict:
+        # XPASS in strict mode → real failure.
+        return "failed", f"[XPASS(strict)] {reason}", None
+    return outcome, longrepr, None
+
+
+def _run_in_container(item: pytest.Item, limit_mb: int) -> tuple[str, Optional[str]]:
+    image = os.environ.get("SODA_MEMTEST_IMAGE", DEFAULT_IMAGE_TAG)
+    bind_root = str(REPO_PARENT)
+    rootdir = str(item.config.rootpath)
```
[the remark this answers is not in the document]
Niels-b: Fixed in 9282c3d2 — the bind mount is now scoped to the `soda-core` and `soda-extensions` repo directories (each mounted at its host path so the editable installs still resolve) instead of the whole REPO_PARENT tree; the credential-mount helper now skips paths under any repo root. Validated end-to-end with a containerized DWH smoke test (warm peak within baseline).

soda-tests/src/helpers/memory_container_stub.py
```diff
@@ -0,0 +1,82 @@
...
+    if not raw:
+        return None
+    try:
+        names = json.loads(raw)
+    except (ValueError, TypeError):
+        return None
+    purpose_lower = table_purpose.lower()
+    for name in names:
+        if purpose_lower in name.lower():
+            return name
```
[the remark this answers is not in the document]
Niels-b: Fixed in 9282c3d2 — matching is now on the purpose as a `_`-delimited token (`_<purpose>_`), so a short purpose can no longer spuriously match inside a longer table name of the form `SODATEST_<purpose>_<hash>`.

soda-tests/src/helpers/data_source_test_helper.py
```diff
@@ -876,11 +1070,91 @@ def _create_test_table(self, test_table: TestTable) -> None:
...
+                    fully_qualified_table_name=table_name,
+                    values=[VALUES_ROW(row) for row in chunk],
+                    columns=columns,
+                )
+            )
+            if len(chunk_sql) > max_len and chunk_size > 1:
+                # Estimate was too loose (extreme escaping). Halve and retry.
+                chunk_size = max(1, chunk_size // 2)
+                continue
+            self.data_source_impl.execute_update(chunk_sql)
```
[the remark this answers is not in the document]
Niels-b: Kept by design, with a clarifying comment added in 9282c3d2. The base `get_max_sql_statement_length()` is 63 MB, but the fat-row fixtures deliberately push ~105 MB single-row INSERTs above it, and the drivers accept them (postgres allows ~1 GB statements). A single row cannot be split, so there is nothing left to shrink. It is also not silent: `execute_update` does not swallow errors — if a driver rejects the statement the exception propagates and fails fixture setup loudly. Failing fast here would break the fat-row fixtures the entire memory suite depends on.

soda-core/src/soda_core/common/value_size.py
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import annotations
+
+import sys
+
+NODE_BUDGET_DEFAULT: int = 1000
+# Recursion guard for adversarial nesting; the node budget governs cost.
+_MAX_DEPTH: int = 32
+
+_NONE_SIZE: int = 16  # pointer size
+
```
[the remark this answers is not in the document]
Niels-b: Fixed in 9282c3d2 — `_NONE_SIZE = sys.getsizeof(None)`.

soda-tests/tests/unit/test_value_size.py
```diff
@@ -0,0 +1,133 @@
...
+_MB = 1_000_000
+
+
+class TestLeaves:
+    def test_scalar_leaves_match_getsizeof(self):
+        for value in ("x" * 1000, b"y" * 1000, 42, 3.14, True):
+            assert estimate_value_size(value) == sys.getsizeof(value)
+
+    def test_none_is_pointer_size(self):
+        assert estimate_value_size(None) == 16
```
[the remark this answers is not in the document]
Niels-b: Fixed in 9282c3d2 — the test now asserts against `sys.getsizeof(None)` to stay aligned with the implementation.

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -17,6 +17,57 @@
...
+def set_memory_optimized_override(enabled: Optional[bool]) -> None:
+    """Enable the memory-optimized fetch driver from outside the env toggle.
+
+    Called by soda-extensions when Soda Cloud sends ``use_memory_optimized`` with
+    the diagnostics-warehouse configuration. ORed with
+    ``MEMORY_OPTIMIZED_DRIVER_ENABLED`` in ``is_memory_optimized_driver_enabled``,
+    so the backend flag can turn the driver on regardless of the env var. A falsy
+    value leaves the env var as the sole control.
+    """
+    global _memory_optimized_override
```
paulteehan, line 42: not too sure about the use of globals here, it feels a bit unusual

soda-core/src/soda_core/common/data_source_connection.py
```diff
@@ -358,6 +409,55 @@ def execute_query_one_by_one(
             cursor.close()
         return description
 
+    def execute_query_one_by_one_memory_optimized(
```
paulteehan, line 412: the function_memory_optimized() naming and pattern here also feels a bit off, feels more like there should be a class with subclasses that are dispatched based on application configuration

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -160,6 +160,20 @@ def execute_query_one_by_one(
...
+        self,
+        sql: str,
+        row_callback: Callable[[tuple, tuple[tuple]], None],
+        log_query: bool = True,
+        row_limit: Optional[int] = None,
+    ) -> tuple[tuple]:
+        # Memory-optimized streaming fetch — see
+        # DataSourceConnection.execute_query_one_by_one_memory_optimized.
+        # Returns the description of the query.
+        return self.data_source_connection.execute_query_one_by_one_memory_optimized(
```
paulteehan, line 173: this for example... architecture doesn't feel right

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
```diff
@@ -112,6 +113,14 @@ def _get_data_type_name_synonyms(self) -> list[list[str]]:
             ["int", "integer"],
         ]
 
+    def get_max_sql_statement_length(self) -> int:
+        # Databricks rejects large statements with INVALID_PARAMETER_VALUE
+        # "Query text size exceeds limit" (observed live at ~30 MB,
+        # 2026-06-12); the documented query-text cap is 16 MiB. Reserve
+        # 1 MiB because we count characters while the server limit is
+        # bytes (multi-byte UTF-8).
+        return 15 * 1024 * 1024
```
paulteehan, line 122: do we know that a 1 MiB buffer is sufficient?  i'm not sure what the impact of characters vs bytes would be

soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source.py
```diff
@@ -69,6 +69,12 @@ def escape_string(self, value: str):
         string_literal = string_literal.replace("'", "''")
         return string_literal
 
+    def get_max_sql_statement_length(self) -> int:
+        # Redshift's documented maximum SQL statement size is 16 MB — far
+        # below the 63 MB generic default. Reserve 1 MB because we count
+        # characters while the server limit is bytes (multi-byte UTF-8).
+        return 15 * 1024 * 1024
```
paulteehan, line 76: again are we sure 1 MiB is enough buffer?

paulteehan (approved):
very nice intricate in-the-weeds work!  a few minor questions, not blocking

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
Niels-b: It's a bit of a guess. We can't convert it to use the correct bytes, as that will duplicate the memory footprint (`encode` creates a copy with the encoded bytes). It depends on the data, so we have to look for a bit of a middle-ground here.
For example: if you have some emoji in your text field it may add some extra bytes, but not the full string. This one can easily be revisited later imo.

soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source.py
Niels-b: See other comment

soda-core/src/soda_core/common/data_source_connection.py
Niels-b: As pointed out in the extensions PR, we can still change/improve this later. Thank you so much for pointing this out!

PR #2758, opened 2026-06-16
ci: add publish-from-tag workflow for PyPI

paulteehan:
## Why

Releases are now driven by **release-buddy** (bump → PR → merge → tag → GitHub Release). But the only workflow that publishes to **public PyPI** is `release.yaml` ("one-button release"), which is `workflow_dispatch`-only and **refuses to run against an already-existing tag** (its `validate` job hard-fails on `Check tag does not exist`). Since the tag already exists by the time we'd publish, `release.yaml` can't be used.

Result: nothing reached public pypi.org between **v4.7.0 (2026-04-17)** and **v4.14.0** — only dev PyPI kept flowing. Internal consumers pull core from private/enterprise PyPI so it went unnoticed; **only OSS users were affected** (stuck on 4.7.0). Affects all 13 release-matrix packages (soda-core + soda-postgres, soda-bigquery, …).

## What

New `.github/workflows/publish.yaml` that **only builds a given ref and uploads** — it never tags, bumps, or creates releases.

- **`release: published`** → auto-publishes new releases. This is the permanent fix: release-buddy creates the Release, this publishes it.
- **`workflow_dispatch` (input: `tag`, `dry_run`)** → backfill/recovery for a specific existing tag (e.g. `v4.14.0`, and the 4.8.0–4.13.0 backlog).
- Uploads to **both** Soda PyPI and public PyPI, reusing the exact secret wiring from `release.yaml` (`PYPI_API_TOKEN`, AWS-secrets Soda creds, `production-release` environment).
- `twine upload --skip-existing` → idempotent; safe re-runs and tolerates a registry already having the version.
- Verifies built artifacts are the clean tag version (not a `.devN`) before any upload — PyPI uploads are irreversible.

## Immediate use

After merge: **Actions → Publish to PyPI → Run workflow → tag `v4.14.0`** (run once with `dry_run: true` first to confirm builds). This backfills the current release for OSS users.

## Follow-ups (not in this PR)

- Backfill v4.8.0–v4.13.0 on demand (dispatch per tag).
- `release.yaml`'s inline build-and-publish becomes redundant once `release: published` is live — consider trimming it to avoid double work (`--skip-existing` already makes the overlap harmless).
- Add a post-release **public-PyPI verification gate** in release-buddy (the existing `checkReleaseArtifact()` already hits pypi.org) so a stalled publish can never go unnoticed again.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -16,18 +16,25 @@ name: Publish to PyPI
 # creates releases.
 #
 # Triggers:
-#   release: published   -> standard path. Fires automatically when a GitHub
-#                           Release is created (e.g. by release-buddy), so new
-#                           releases reach public PyPI without a manual step.
-#   workflow_dispatch    -> backfill / recovery. Publish a specific existing
-#                           tag (e.g. v4.14.0, or the 4.8.0-4.13.0 backlog).
+#   push: tags (v*)      -> standard path. The release tag is the canonical
+#                           "this version exists" signal and is always pushed
+#                           by release-buddy, so we key off the tag directly.
+#                           (A GitHub Release is best-effort — release-buddy
+#                           tolerates `gh release create` failing — so keying
+#                           off `release: published` could silently skip a
+#                           publish, the very gap this closes. A tag can't be
+#                           missed.)
+#   workflow_dispatch    -> backfill / recovery for an already-existing tag
+#                           (e.g. v4.14.0, or the 4.8.0-4.13.0 backlog); a tag
+#                           pushed before this workflow existed won't re-fire.
 #
 # twine runs with --skip-existing, so re-runs are safe and the automatic path
 # does not fail when a registry already has the version (e.g. Soda PyPI).
 
 on:
-  release:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,200 @@
...
+#   release: published   -> standard path. Fires automatically when a GitHub
+#                           Release is created (e.g. by release-buddy), so new
+#                           releases reach public PyPI without a manual step.
+#   workflow_dispatch    -> backfill / recovery. Publish a specific existing
+#                           tag (e.g. v4.14.0, or the 4.8.0-4.13.0 backlog).
+#
+# twine runs with --skip-existing, so re-runs are safe and the automatic path
+# does not fail when a registry already has the version (e.g. Soda PyPI).
+
+on:
```
m1n0, line 34: shouldn't this run by default on tags?

paulteehan: Yes good catch, fixed

m1n0 (approved):
thanks for addressing this!

PR #2760, opened 2026-06-17
Emit "time" measure and millisecond values for freshness check diagnostics

m1n0:
## Problem
V4 contract-scan freshness checks render as raw floats (e.g. `11.155277…`) in Soda Cloud's check charts instead of a human-readable duration — V3 showed clean unit-bearing values like "X days and Y hours". Two gaps in the freshness diagnostics soda-core sends:
1. No measure/type marker on the value, so Soda Cloud formats it as a plain number.
2. The value is in the check's configured unit (hours/days), but the duration formatter expects **milliseconds**.

## Fix
Align the freshness diagnostics wire payload with the proven V3 `soda-library` `FreshnessCheck` contract:
- emit `diagnostics.measure = "time"` so the value is formatted as a duration
- express `diagnostics.value` and the `fail` threshold bounds in **milliseconds** (scaled from the configured unit)

Implemented via two hooks on `CheckResult`:
- `get_soda_cloud_measure()` — `None` by default, `"time"` for freshness
- `to_soda_cloud_measure_value()` — identity by default, unit → ms for freshness

Both default to no-ops, so **non-time check payloads are byte-for-byte unchanged**.

## Tests (written test-first, red → green)
- **Unit** (`test_soda_cloud.py`): freshness diagnostics carry `measure == "time"` and value `3_600_000` ms; non-time checks omit `measure`.
- **Integration** (`test_freshness_check.py`): end-to-end upload payload asserts `measure`, value in ms, and `fail.greaterThanOrEqual` in ms.
- **Regression**: missing / duplicate / row-count diagnostics unchanged.

Formatted with black 23.3.0 (repo-pinned).

## Notes
- Freshness is floored to whole-second precision (`freshness_in_seconds`) before scaling to ms; V3 kept sub-second precision. Negligible for freshness.
- V3 also emitted a `warn` threshold in ms; V4's diagnostics builder has no `warn` block today — out of scope here.

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1774,14 +1774,25 @@ def _build_diagnostics_json_dict(check_result: CheckResult) -> Optional[dict]:
     # if check_result.outcome == CheckOutcome.EXCLUDED:
     # return None
 
-    return {
+    raw_value = (
         #  TODO: this default 0 value is here only because check.diagnostics.value is a required non-nullable field in the api.
-        "value": int(check_result.threshold_value)
+        int(check_result.threshold_value)
         if isinstance(check_result.threshold_value, bool)
-        else (check_result.threshold_value or 0),
+        else (check_result.threshold_value or 0)
+    )
+    diagnostics: dict = {
+        # ``value`` (and the fail thresholds below) are converted into the measure's
+        # wire unit — identity for plain numbers, milliseconds for "time".
+        "value": check_result.to_soda_cloud_measure_value(raw_value),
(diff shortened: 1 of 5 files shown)


m1n0:
Superseded by #2770, which carries the same fix (emit `measure="time"` + millisecond values for freshness diagnostics) plus two improvements: a guard so an unknown/`None` unit falls back to the default instead of `KeyError`-ing mid-upload (the raw `_UNIT_SECONDS[self.unit]` subscript here would abort the scan's result POST), and additional `test_soda_cloud.py` coverage. The other three changed files are identical between the two. Closing in favour of #2770.

PR #2762, opened 2026-06-18
Redact SQL Server connection parameter values in logs

wunianze666-netizen:
## Summary
- stop logging raw SQL Server `connection_parameters` values when building the connection string
- keep the parameter names in the log output so debugging still shows which keys were applied
- add a unit test that verifies secrets stay out of the logs while the connection string still includes the configured parameters

## Why
`connection_parameters` can carry security-sensitive values such as access tokens. Logging those values verbatim makes it easy to leak secrets through CI logs or shared runtime logs.


mivds (approved):

mivds:
Hi @wunianze666-netizen, thank you for the contribution! This looks good to me. I've taken to opportunity to make a tiny refactor to simplify the test you added, and fixed a latent bug.

PR #2769, opened 2026-06-24
feat: add failed_rows keys_query check form (r-601)

Niels-b:
## Description

Adds a third `failed_rows` check form — `keys_query` (+ optional `key_column_mapping`) — alongside the existing `expression` and `query` forms.

- **What problem:** custom-SQL failed-rows checks each get their own per-check diagnostics table (`frq_`), which sprawls (the majority of checks are custom SQL). The new `keys_query` form returns the dataset's diagnostics-warehouse key columns, which Soda hashes and routes into the shared per-dataset failed-keys/rows tables so custom-SQL failures consolidate and become scoreable. This PR is the soda-core half: parsing, validation, and running the form as a complete failed-rows check.
- Parse + validate the new form (the three forms are mutually exclusive); `keys_query` runs with no diagnostics-warehouse config required.
- JSON schema declares `keys_query` / `key_column_mapping` and forbids `key_column_mapping` unless `keys_query` is set.
- **Downstream impact:** key routing into the shared diagnostics-warehouse tables is implemented in soda-extensions on the matching `r-601-implement-failed-keys-check` branch. No behavior change to existing `expression` / `query` checks.

Ticket: r-601.

## Checklist

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
@@ -78,12 +78,19 @@ class FailedRowsCheckImpl(CheckImpl):
                 RowCountMetricImpl(contract_impl=contract_impl, check_impl=self)
             )
 
-        elif self.is_query_check():
+        elif self.is_query_check() or self.is_keys_query_check():
+            # The `query` and `keys_query` forms both count failing rows from a user query. They differ
+            # only downstream: the `query` form's failures land in a per-check table, while the
+            # `keys_query` form's keys are hashed and routed into the shared diagnostics-warehouse
+            # fk_/fr_ tables (handled in soda-extensions). Counting is identical here.
+            active_query: str = self.failed_rows_check_yaml.query or self.failed_rows_check_yaml.keys_query
             self.failed_rows_count_metric_impl = self._resolve_metric(
-                FailedRowsQueryMetricImpl(contract_impl=contract_impl, column_impl=column_impl, check_impl=self)
+                FailedRowsQueryMetricImpl(
+                    contract_impl=contract_impl, column_impl=column_impl, check_impl=self, query=active_query
+                )
             )
 
-            sql = self.failed_rows_check_yaml.query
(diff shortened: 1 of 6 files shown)


m1n0:
**`key_column_mapping` without `keys_query`: JSON schema rejects it, the parser only warns**

Minor consistency point between the two validation layers this PR touches:

- **JSON schema** (`soda_data_contract_json_schema_1_0_0.json`): the new `if {required: [key_column_mapping]} → then {required: [keys_query]}` sits as a sibling of the form `oneOf`. So a check with `query` + `key_column_mapping` is a **hard schema failure** — the `then` forces `keys_query`, which contradicts the query-only `oneOf` branch.
- **Parser** (`failed_rows_check_yaml.py`): the same combination only **warns** (`'key_column_mapping' is only used with 'keys_query' mode`) and otherwise parses fine.

Net effect: the Cloud/editor (schema-validated) rejects a contract that the engine accepts with a warning. Not a correctness bug — both nudge the user the same direction — but worth a deliberate call on which layer is authoritative. If the schema's hard-fail is the intended behavior, consider upgrading the parser warning to an error so local runs and Cloud agree.

m1n0:
**Follow-up: public docs for `keys_query` / `key_column_mapping`**

This PR adds two user-facing `failed_rows` contract fields — `keys_query` and `key_column_mapping`. Before the feature is user-visible, the public docs (docs repo / GitBook `failed_rows` reference) should cover:

- what `keys_query` is for — consolidating custom-SQL failures into the shared diagnostics-warehouse `fk_`/`fr_` tables instead of a per-check `frq_` table, so they become scoreable;
- the requirement that the query output supply the dataset's diagnostics-warehouse key column(s);
- `key_column_mapping` — maps each DWH key column to the query output column that supplies it; identity (same names) when omitted;
- mutual exclusivity with `expression` / `query`.

Not a merge blocker — flagging so the docs work is tracked alongside the soda-extensions half (r-601). Happy to open a Linear ticket or a docs PR for this.

m1n0 (approved):

PR #2773, opened 2026-06-30
OBSL-1013: ship table discovery in soda-core (soda data-source discover)

mivds:
## PR chain (soda-core)
1. #2773 — ship table/column discovery  ← this PR

---

## Summary

First-party table discovery in soda-core (OBSL-1013, part of the v3→v4 observability port OBSL-1003). Adds a `soda data-source discover` command that posts a **DQN-only** `sodaCoreInsertScanResults` payload (`version: "4"`) for the BE's v4 `DiscoveryScanHandlerModule`.

- `DatasetIdentifier.from_object(...)` — builds a **dialect-correct** DQN from a discovered `FullyQualifiedObjectName` using the dialect's `get_database_prefix_index()`/`get_schema_prefix_index()` hooks (verified across all dialects: postgres `[db,schema]`, duckdb `[schema]` (catalog dropped), BigQuery `[project,dataset]`, oracle/db2/sparkdf `[schema]`). No hand-rolled prefixes.
- `DiscoveryRun.execute(...)` — discovers within a scope, filters `__soda_temp` + include/exclude (pushed down to `MetadataTablesQuery`'s server-side SQL `LIKE` filters), maps each object to its DQN. Covers **base tables, views and materialized views** (explicit `object_types`, matching v3 — v3's `information_schema.tables` query had no type filter; the metadata query's `None` default is tables-only). Unknown table types coerce to `TABLE` in `convert_table_type_to_enum` and are included, as in v3. System schemas are excluded via `SqlDialect.is_system_schema` (base: `information_schema`; dialects extend — postgres adds `pg_catalog`/`pg_toast`, Fabric adds `sys`/`queryinsights`).
- `discovery_payload` — builds the DQN-only v4 envelope as a `SodaCoreInsertScanResultsDTO` (TypedDict in `soda_cloud_dto`, mirroring the backend `SodaCoreInsertScanResultsCommand` field set: required keys = the @NotNull/@NotEmpty fields + the `Literal` `type` discriminator). Posted via `SodaCloud.insert_scan_results(payload) -> bool` (same accepted-bool contract as `mark_scan_as_failed`) — the transport lives on `SodaCloud`, not in the discovery module. DTO + method are the intended shared shape for other result-inserting flows: profiling adopts them in the soda-extensions PR; the contract flow keeps its richer `send_check_collection_results` (response post-processing) and is converged separately.
- `soda data-source discover` CLI command + `handle_discover_data_source` handler (failure-reporting flow below).

> **Merged `main`** (incl. #2779 "report runner scan FAILED with logs on abort re-raise"): the failure-reporting test file was reconciled as a union — #2779's abort-path mark-with-logs tests and this branch's rejected-mark tests both live and pass. Audited: the abort helper (`_report_runner_scan_failed_before_reraise`, fires only on the `abort_on_first_error` re-raise) and the phase-3 combined-upload mark are mutually exclusive per scan (uncaught raise → helper marks once and skips phase 3; error caught inside `verify()` → phase 3 marks once), so there is no double-report or contradiction.

### Handler structure

`handle_discover_data_source` receives fully resolved dependencies — a `DataSourceImpl`, a required `SodaCloud` and the scan definition name — instead of file paths. The three resolvers live in `soda_core.cli.handlers.dependencies` and share one exception contract: `resolve_soda_cloud(path) -> SodaCloud`, `resolve_data_source(path) -> DataSourceImpl` and `resolve_scan_definition_name(arg) -> str` (non-Optional) raise `ScanExecutionFailedException` carrying a user-facing message for expected shapes (missing flag, missing or syntactically invalid YAML file, unparseable/invalid config, validation rejection). The catch tuples are `(Invalid<X>ConfigurationException, YamlParserException, ValueError)` — deliberately broad on `ValueError`: pydantic `ValidationError` is a `ValueError` subclass, and plain value errors during config parsing are user-config shapes; the breadth is an accepted trade-off for symmetric resolvers. Genuinely unexpected failures (e.g. `ImportError` from a missing plugin) propagate raw so the caller logs them with the traceback — the deliberate traceback-for-DS-creation asymmetry lives on there. Nothing is logged at raise sites; `ScanExecutionFailedException` extends `SodaCoreException` (audited: the only `except SodaCoreException` site wraps contract-verification query execution, which this path never crosses). `SodaCloud.from_yaml_source` raises `InvalidSodaCloudConfigurationException` for a config missing the top-level `soda_cloud` key, same contract as its api_key checks.

`run_with_failure_reporting(soda_cloud, command)` takes the **already-constructed reporting channel** and a command — construction is injected at the wiring layer, which decides what resolves inside the command. The wrapper now hands the command **its own `Logs` collector** (`command(logs)`): a downstream flow that runs a check-collection session can thread it through (each impl built with `logs=logs`) so its records reach the failure report — without this a command that builds its own inner `Logs` would displace the wrapper's collector and lose its failure logs. Discovery constructs no inner `Logs` (DiscoveryRun emits via `soda_logger`, already captured by the active wrapper collector), so its wiring accepts and ignores the argument. It owns only the `Logs` lifecycle and the **single logging site** with two arms picking the log form: SEFE → the clean message, no traceback; any other exception → `logger.exception` with the traceback. Both are logged before the records are captured, so the failure line is part of the `mark_scan_as_failed` report. The discover wiring resolves `soda_cloud` first in a small guard (unusable → log + exit 4 `RESULTS_NOT_SENT_TO_CLOUD` — nothing can be reported; raw resolution failures get the same exit with a traceback log), then runs the wrapper with the data source and scan definition name resolving **inside** the wrapped command so their failures take the standard mark-with-logs mapping. Cloud-resolution failure records are not part of a `mark_scan_as_failed` report (the guard runs before the wrapper's `Logs`) — they are never deliverable in that state, the exit code carries the signal. A rejected results upload is not a failure raise: the handler returns `RESULTS_NOT_SENT_TO_CLOUD` directly. Like `failure_reporting`, `dependencies` is a stable import point for the soda-extensions CLIs.

### Local flow (no `-sc`)

`-sc` is optional: without it, discovery runs locally via `handle_discover_data_source_locally` — connect → `DiscoveryRun.execute` → print the discovered DQNs to the console (one per line, plus a summary count) → 0 `OK`. No scan lifecycle and no Cloud calls. The handler propagates failures raw; the local wiring is the single logging site (SEFE → clean message, anything else → traceback) and maps every failure to 3 `LOG_ERRORS` (ad-hoc semantics — there is no scan to mark). One guard in the wiring, before choosing the flow: `SODA_SCAN_ID` set without `-sc` is a launcher misconfiguration, NOT a local run — error logged, exit 4 `RESULTS_NOT_SENT_TO_CLOUD` so the launcher fallback marks the scan failed. `--scan-definition-name` is Cloud-only and ignored locally: the local wiring logs a warning when it is provided without `-sc`.

### Mandatory scan definition name (Cloud flow)

The Cloud flow's scan definition name is **mandatory**: `--scan-definition-name` CLI arg > `SODA_SCAN_DEFINITION` env, with no default — an implicit per-data-source name would silently register a new scan definition on Soda Cloud on typos or missing config. `resolve_scan_definition_name` is called inside the command wrapped by `run_with_failure_reporting`, so a missing name raises `ScanExecutionFailedException` (carrying the user-facing message; the wrapper logs it) into the standard failure mapping: managed scans get marked failed with the captured log records and exit 3; ad-hoc runs exit 3 (see the stage table). The handler takes the resolved name as a required parameter. The launcher always passes `--scan-definition-name`, so managed runs are unaffected.

### Failure reporting & exit codes

The CLI wiring resolves the Soda Cloud config **first** and captures the run's log records. On engine failures in a managed scan (`SODA_SCAN_ID` set), it reports the failure to Soda Cloud via `mark_scan_as_failed` with the full log records — mirroring contract verification — instead of relying on the launcher's generic fallback. The decision+send step is `soda_core.cli.handlers.failure_reporting.report_scan_execution_failure(soda_cloud, log_records)` so the soda-extensions flows (profiling, metric monitoring) can reuse it.

| Stage | Sent to Soda Cloud | Exit code |
|---|---|---|
| Missing `-sc` | nothing — ad-hoc: local run (DQNs printed to the console); managed (`SODA_SCAN_ID` set): launcher misconfiguration | ad-hoc: local flow's own code (0 `OK` / 3 `LOG_ERRORS`); managed: 4 `RESULTS_NOT_SENT_TO_CLOUD` — launcher fallback marks the scan failed |
| Unparseable cloud config (`-sc` given) | nothing (nothing can reach Cloud) | 4 `RESULTS_NOT_SENT_TO_CLOUD` — launcher fallback marks the scan failed |
| Missing scan definition name (`-sc` given, no `--scan-definition-name`/`SODA_SCAN_DEFINITION`) | `mark_scan_as_failed` + full log records (if `SODA_SCAN_ID`) | 3 `LOG_ERRORS` on success; 4 if the mark call fails; 3 without a scan id (ad-hoc, nothing sent) |
| Data source YAML unparseable | `mark_scan_as_failed` + full log records (if `SODA_SCAN_ID`) | 3 `LOG_ERRORS` on success; 4 if the mark call fails; 3 without a scan id (ad-hoc, nothing sent) |
| Connection failure | same as above | same as above |
| Discovery query failure | same as above | same as above |
| Any other unexpected engine exception (payload build, connection close, exception escaping the send) | same as above | same as above |
| Results upload rejected/failed | payload attempted, not accepted | 4 `RESULTS_NOT_SENT_TO_CLOUD` |
| Success | discovery results | 0 `OK` |

`SodaCloud.mark_scan_as_failed` returns a success bool. The contract-verification mark sites (per-file `verify()` and the session-level combined-upload path) use it: a rejected mark sets `sending_results_to_soda_cloud_failed`, so the run exits 4 and the launcher fallback marks the scan failed instead of a silent `LOG_ERRORS`. The `verify_api.py` mark site is unchanged: it re-raises and has no result object to flag.

### Notes for reviewers
- DQN-only is intentional and matches the v3-against-v4 behaviour (`discover_v4_data_source`) and the OBSL-1026 golden.
- Query scope is currently `prefixes=[]` (discover all visible); this PR ships system-schema filtering for **`information_schema` on every dialect (base `is_system_schema`) plus per-dialect extensions on postgres, redshift and fabric only**. The other in-repo dialects (snowflake, databricks, trino, sqlserver/synapse, bigquery, athena, duckdb) inherit the base and so filter `information_schema` only — their non-`information_schema` system namespaces (e.g. Snowflake `ACCOUNT_USAGE`, Databricks/Trino `system`, SQL Server/Synapse `sys`) aren't excluded yet. That's harmless under `prefixes=[]` (discovery doesn't reach them), but it means the remaining follow-up is **both prefix scoping and the per-dialect system-schema overrides** (the latter carried by the extensions PRs), not prefix scoping alone.
- Deliberate behavior choice: `FabricSqlDialect.is_system_schema` extends (rather than replaces) the base `information_schema` exclusion. Fabric's `INFORMATION_SCHEMA` contains views, and discovery includes views — without this, Fabric would be the one dialect surfacing catalog metadata as discoverable datasets.
- Envelope completeness vs the real BE DTO is confirmed at parity time (OBSL-1021), not in this PR.

## Test Plan
- [x] Unit: `from_object` across dialect shapes; `DiscoveryRun` filtering/mapping; payload shape; CLI arg mapping (7 tests)
- [x] Unit: `report_scan_execution_failure` branches (no scan id, no usable cloud, mark accepted/rejected/raising) + per-stage discovery handler exit-code tests; rejected-mark tests for both contract-flow mark sites
- [x] Integration: `soda-tests/tests/integration/test_discovery.py` against postgres via `MockSodaCloud` — DQN-only metadata, `version == "4"`, test table's DQN present; views discovered alongside the base table (v3 parity)
- [x] Unit: local flow — handler prints DQNs + connection lifecycle + failures propagate raw; wiring: no `-sc` → local (SodaCloud never constructed), unusable ds → 3, raising handler → 3 (single traceback log), `SODA_SCAN_ID`+no `-sc` guard → 4, guard scoped to missing `-sc` only, `--scan-definition-name` without `-sc` warns
- [x] Unit: views/materialized views pinned into the DQN list with explicit `object_types`; system-schema exclusion per dialect (base, postgres, Fabric incl. INFORMATION_SCHEMA); mandatory scan definition name — arg > env precedence, missing name raises; wiring: managed → mark-failed + 3, ad-hoc → 3, env fallback reaches the handler
- [x] Unit: resolver exception contract (raise with message match, cause chained, nothing logged at raise site, `ImportError` propagates raw); wrapper (channel injected, command receives the wrapper's own `Logs` — identity + downstream-records-reach-report pinned, single log line with/without traceback pinned); wiring guard (unusable cloud → 4 before any ds resolution — reorder proof); SEFE-extends-SodaCoreException pin
- [x] Unit: YAML syntax error / missing file / missing `soda_cloud` key resolve to clean messages (real files); DTO conformance pin (emitted keys ⊆ DTO, required keys emitted); `SodaCloud.insert_scan_results` (payload passthrough, accepted/rejected/no-response mapping)
- [x] Regression: existing `test_cli.py` (21) pass; full soda-tests unit suite (1060 passed); pre-commit (isort/black/autoflake) clean
- [x] Reviewer: confirm payload accepted by BE v4 `DiscoveryScanHandlerModule` (parity-time)


m1n0:
Reviewed with focus on flow correctness, codebase practices, and template quality (this is the seam the profiling/MM PRs will copy). **No blocking findings.**

Verified against the actual sources rather than trusting the PR body: DTO required/optional keys match `SodaCoreInsertScanResultsCommand` 1:1 (incl. `version: "4"` ↔ `@SerializedName("4")`, DQN-only `MetadataCoreDTO`); postgres/redshift `is_system_schema` are verbatim v3 ports and Fabric's base-exclusion is the one (flagged) divergence; the exit-code table is pinned by tests; `Logs.close()` / `close_connection()` are safe on the failure paths. Unit suite green locally; local flow semantics (print DQNs, warn on `--scan-definition-name` without `-sc`, exit 4 on `SODA_SCAN_ID` without `-sc`) make sense to me.

Non-blocking observations:

- **`sys` on SQL Server/Synapse**: Fabric excludes `sys`/`queryinsights` because its INFORMATION_SCHEMA lists them as views and discovery now includes views; plain sqlserver/synapse have no exclusion (matches v3). Worth one manual glance at discovery output on a real SQL Server to confirm nothing from `sys` leaks. Same, lower stakes, for DuckDB system catalogs.
- **Bare `--include`** (`nargs="*"`) parses to `[]` → falsy at `metadata_tables_query.py:96` → silently means "include everything". `nargs="+"` would make it a parse error.
- **Style**: `PostgresSqlDialect.is_system_schema` hardcodes `information_schema` instead of `super()` like redshift/fabric — behavior is right (v3 verbatim), but extensions dialects will copy one of these as the template; suggest aligning.
- **Chain follow-up**: v3's db2 dialect has its own `is_system_schema` (`SYS*` prefixes) and other extensions-owned dialects may need equivalents — make sure the extensions PRs carry their overrides; the green extensions run only proves existing tests.

m1n0 (approved):
LGTM, only a few NITs picked up by claude!

I just realized - this is happening in OSS, how is the split going to work?

soda-core/src/soda_core/cli/handlers/data_source.py
```diff
@@ -126,3 +133,88 @@ def build_test_connection_log_uploader(
         dataset="",
     )
     return Logs(gatherer=logs_queue)
+
+
+def handle_discover_data_source(
```
Niels-b, line 163: NIT: Extract common parts of this function and the one below (`handle_discover_data_source_locally`).

soda-core/src/soda_core/common/data_source_impl.py
Niels-b, line 412: IIRC this gives all objects (can be view, can be table,...) which are currently supported. This support varies per datasource.

soda-core/src/soda_core/discovery/discovery_run.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Optional
+
+from soda_core.common.data_source_impl import DataSourceImpl
+from soda_core.common.dataset_identifier import DatasetIdentifier
+from soda_core.common.statements.table_types import TableType
+
+SODA_TEMP_PREFIX = "__soda_temp"
+
+
+class DiscoveryRun:
```
Niels-b: Not 100% convinced on the name because of "run". This seems like a v3 concept still. 
Don't really have a better alternative to be honest. We can definitely keep this, just wanted to write my thoughts 😉 .

Niels-b (approved):
LGTM, thanks for the work on this.

Went through it manually to understand the flows. And also had Claude do a review, it still has some things that I'll leave up to you if you want to address.

Niels-b (commented):
Thorough review — fanned a few agents out over the discovery logic, the failure/exit-code wiring, the Cloud transport/DTO, cross-dialect correctness, and the tests.

Overall this is a strong, well-documented PR. The failure-handling and exit-code path in particular matches the stage table exactly, and the tests are meaningful (correct mock targets, real assertions, `monkeypatch` env isolation). **No blockers.**

Inline notes below: one small concrete fix (#1, the Postgres case-sensitivity), a couple of things worth confirming against the BE (system-schema coverage framing, DQN casing), and a few test-coverage gaps. None of these need to block merge.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
```diff
@@ -86,6 +86,9 @@ class PostgresSqlDialect(SqlDialect, sqlglot_dialect="postgres"):
     def supports_materialized_views(self) -> bool:
         return True
 
+    def is_system_schema(self, schema_name: str) -> bool:
+        return schema_name.startswith("pg_") or schema_name == "information_schema"
```
Niels-b: `is_system_schema` here is case-sensitive on both arms (`startswith("pg_")` and `== "information_schema"`), whereas the base (`schema_name.lower() == "information_schema"`), Fabric, and Redshift all normalize with `.lower()` + `super()`. It's safe in practice since Postgres folds unquoted identifiers to lowercase, but it's inconsistent with the sibling dialects, and the unit test only exercises lowercase inputs — so an uppercase `PG_CATALOG`/`INFORMATION_SCHEMA` wouldn't be caught. Suggest aligning:

```python
return schema_name.lower().startswith("pg_") or super().is_system_schema(schema_name)
```

and adding `("PG_CATALOG", True)` / `("INFORMATION_SCHEMA", True)` to the parametrized test (Redshift's already checks `PG_AUTOMV`, Fabric checks `SYS`).

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -1394,6 +1394,14 @@ def get_database_prefix_index(self) -> int | None:
     def get_schema_prefix_index(self) -> int | None:
         return 1
 
+    def is_system_schema(self, schema_name: str) -> bool:
+        """Check if the schema is a data source internal/system schema.
+
+        Objects in system schemas are excluded from discovery. Dialects
+        override this to add their data source specific system schemas.
+        """
+        return schema_name.lower() == "information_schema"
```
Niels-b, line 1403: Only postgres, fabric and redshift override `is_system_schema`; the other in-repo dialects (snowflake, databricks, trino, sqlserver/synapse, bigquery, athena, duckdb) inherit this base and so filter `information_schema` only. That's fine for the current `prefixes=[]`, DQN-only behavior, but several of those expose non-`information_schema` system namespaces that would surface as discoverable datasets once discovery is scoped to reach them — e.g. Snowflake `ACCOUNT_USAGE`/`SNOWFLAKE`, Databricks/Trino `system`, SQL Server/Synapse `sys`. Might be worth either adding overrides for those or narrowing the PR note ("system-schema filtering ships in this PR") to "3 dialects + `information_schema` for the rest", so the remaining coverage isn't quietly lost as a follow-up.

soda-core/src/soda_core/common/dataset_identifier.py
```diff
@@ -32,6 +36,33 @@ def parse(cls, dataset_qualified_name: Optional[str]) -> DatasetIdentifier:
 
         return cls(data_source_name, prefixes, dataset_name)
 
+    @classmethod
+    def from_object(
+        cls,
+        data_source_name: str,
+        sql_dialect: "SqlDialect",
+        fully_qualified_object_name: "FullyQualifiedObjectName",
```
Niels-b, line 44: The discovered DQN takes its casing from `information_schema` (actual stored casing), while a contract-authored DQN comes from user YAML via `parse()` (whatever the user typed). On case-folding warehouses (Postgres → lower, Snowflake → upper) the same physical table can yield two case-different DQNs, and `to_hash()` hashes `to_string()` verbatim — so discovery and contract verification could land on two different dataset identities on Cloud. The PR notes this matches v3; could we confirm the BE resolves dataset identity case-insensitively so the two paths converge? If not, we may want to normalize casing consistently on both sides.

soda-core/src/soda_core/discovery/discovery_run.py
```diff
@@ -0,0 +1,38 @@
...
+    ) -> list[str]:
+        # include/exclude are pushed down as SQL LIKE filters. The __soda_temp prefix is
+        # filtered in Python instead: in a LIKE pattern its leading underscores would be
+        # single-character wildcards.
+        objects = data_source_impl.discover_qualified_objects(
+            prefixes=prefixes,
+            # Every object type, explicitly: discovery matches v3 (no type filter — base
+            # tables, views and materialized views alike), while the metadata query's
+            # None default is tables-only for backward compatibility of other callers.
+            object_types=[TableType.TABLE, TableType.VIEW, TableType.MATERIALIZED_VIEW],
```
Niels-b: The "unknown table types coerce to `TABLE` and are included (v3 parity)" behavior lives in `convert_table_type_to_enum` / `MetadataTablesQuery.get_results`, but every `DiscoveryRun` unit test feeds a fake data source that returns pre-built `FullyQualified*Name` objects and bypasses that coercion — so the claimed parity path isn't actually covered. Consider a small `MetadataTablesQuery.get_results` test with a row like `table_type="FOREIGN TABLE"` asserting it becomes a `FullyQualifiedTableName`.

soda-tests/tests/integration/test_discovery.py
```diff
@@ -0,0 +1,145 @@
...
+    assert posted["type"] == "sodaCoreInsertScanResults"
+    assert posted["version"] == "4"
+    assert all(set(entry.keys()) == {"datasetQualifiedName"} for entry in posted["metadata"])
+    expected_suffix = test_table.unique_name.lower()
+    assert any(
+        entry["datasetQualifiedName"].lower().endswith(expected_suffix) for entry in posted["metadata"]
+    ), f"{test_table.unique_name} not found in {posted['metadata']}"
+
+
+def test_discovery_includes_views(data_source_test_helper: DataSourceTestHelper):
```
Niels-b, line 56: `test_discovery_includes_views` pins plain views end-to-end (nice — with an internal skip where unsupported), but materialized views are a distinct `object_types` entry and a distinct `FullyQualified*Name` branch, and they're only covered by the fake-data-source unit test. Postgres supports MVs (`supports_materialized_views()` is True) and the helper has `create_materialized_view_from_test_table`, so it'd be cheap to extend this test (or add a sibling) to assert an MV shows up too.

soda-core/src/soda_core/discovery/discovery_run.py
```diff
@@ -0,0 +1,38 @@
...
+            include_table_name_like_filters=include,
+            exclude_table_name_like_filters=exclude,
+        )
+        objects = [o for o in objects if not o.get_object_name().lower().startswith(SODA_TEMP_PREFIX)]
```
Niels-b: The `__soda_temp` filter is `.lower().startswith(...)`, but the unit coverage only feeds a lowercase `__soda_temp_*` name, so the case-insensitivity here isn't actually asserted. A quick uppercase case (`__SODA_TEMP_X`) would lock it in.

soda-core/src/soda_core/discovery/discovery_payload.py
```diff
@@ -0,0 +1,63 @@
...
+        "type": "sodaCoreInsertScanResults",
+        "version": "4",
+        "scanType": None,
+        "definitionName": scan_definition_name,
+        "defaultDataSource": data_source_name,
+        "dataTimestamp": convert_datetime_to_str(data_timestamp),
+        "scanStartTimestamp": convert_datetime_to_str(scan_start_timestamp),
+        "scanEndTimestamp": convert_datetime_to_str(scan_end_timestamp),
+        "hasErrors": False,
+        "metadata": [{"datasetQualifiedName": dqn} for dqn in dqns],
```
Niels-b, line 58: Worth a tiny test that `build_discovery_payload(dqns=[])` yields `metadata == []` (and still a valid `version`/timestamps envelope) — the empty-discovery path isn't currently asserted, and it's the shape the BE will see for a data source with nothing to discover.

soda-core/src/soda_core/common/dataset_identifier.py
```diff
@@ -32,6 +36,33 @@ def parse(cls, dataset_qualified_name: Optional[str]) -> DatasetIdentifier:
 
         return cls(data_source_name, prefixes, dataset_name)
```
Niels-b, line 38: Minor/edge: `from_object` appends `database_name` raw, and `parse()` splits DQNs on `/`. For dialects where a prefix component can itself contain `/` — e.g. an Athena catalog — the built DQN wouldn't round-trip cleanly back through `parse()`. It's build-only and an edge case (and Athena has custom `extract_*_from_prefix` handling that this path bypasses), so flagging for awareness rather than asking for a change here.

soda-core/src/soda_core/common/dataset_identifier.py
mivds: We intentionally don't normalize casing here. Case-different quoted identifiers are *distinct physical tables* — e.g. on Postgres `"Test"` and `"test"` coexist side-by-side when quoted at creation time — so folding case would collapse two real datasets into one identity. Discovery takes its casing from `information_schema` (actual stored casing) and contract DQNs come from user YAML verbatim; preserving both as-is is the correct behavior, and matches v3. Keeping as-is.

soda-core/src/soda_core/cli/handlers/data_source.py
mivds: Done — extracted the shared open→discover→close block into `_discover_dqns(...)`; both `handle_discover_data_source` and `handle_discover_data_source_locally` now call it.

soda-core/src/soda_core/common/data_source_impl.py
mivds: Right — `object_types=[TABLE, VIEW, MATERIALIZED_VIEW]` is passed explicitly and each dialect's metadata query only returns the types it supports (unsupported ones simply don't appear). Matches v3, no change needed.

soda-core/src/soda_core/discovery/discovery_run.py
mivds: Renamed. `DiscoveryRun` was a stateless class with a single `@staticmethod`, so it's now a module-level function `discover_dataset_dqns(...)` in `discovery.py` — drops the v3 "run" framing and is the honest shape for what it does. Easy to promote back to a class if scoping later needs state.

soda-postgres/src/soda_postgres/common/data_sources/postgres_data_source.py
mivds: Fixed — now `schema_name.lower().startswith("pg_") or super().is_system_schema(schema_name)`, aligned with redshift/fabric. Added `("PG_CATALOG", True)` / `("INFORMATION_SCHEMA", True)` to the parametrized test.

soda-core/src/soda_core/common/sql_dialect.py
mivds: Narrowed the PR note. It now spells out that this PR ships `information_schema` on every dialect plus per-dialect overrides on **postgres/redshift/fabric only**; the other dialects' non-`information_schema` system namespaces (Snowflake `ACCOUNT_USAGE`, Databricks/Trino `system`, SQL Server/Synapse `sys`, …) remain a follow-up carried by the extensions PRs — harmless under `prefixes=[]`. Not adding those overrides here (YAGNI until scoping reaches them).

soda-core/src/soda_core/discovery/discovery_run.py
mivds: Added `test_metadata_tables_query_get_results.py` — pins the `convert_table_type_to_enum` coercion directly, including a `"FOREIGN TABLE"` row coercing to `FullyQualifiedTableName` (plus known-type mapping and type filtering).

soda-tests/tests/integration/test_discovery.py
mivds: Added `test_discovery_includes_materialized_views` (integration), mirroring the views test via `create_materialized_view_from_test_table` with the same internal skip where unsupported. Postgres exercises the MV branch end-to-end now.

soda-core/src/soda_core/discovery/discovery_run.py
mivds: Added an uppercase `__SODA_TEMP_Y` entry to the filter test to lock in the case-insensitive `.lower().startswith`.

soda-core/src/soda_core/discovery/discovery_payload.py
mivds: Added `test_payload_empty_discovery_yields_empty_metadata` — asserts `metadata == []` with a still-valid `version` + @NotNull timestamps envelope.

soda-core/src/soda_core/common/dataset_identifier.py
mivds: Noted — build-only edge, and Athena's custom `extract_*_from_prefix` handling bypasses this path, so there's no round-trip through `parse()` in practice today. Leaving as-is per your flag; worth revisiting if/when Athena discovery scoping lands.

PR #2774, opened 2026-06-30
OBSL-1025: emit metrics[] in scan-results payload when measurements present

mivds:
## PR chain (soda-core)
2. #2774 — emit `metrics[]` in scan-results payload  ← this PR

---

## Summary

Extends v4's `_build_check_collection_results_json_dict` to emit a top-level `metrics: [...]` array — the prerequisite for metric monitoring (OBSL-1008 / OBSL-1024) to send measurements.

- Adds an optional `measurement_dicts: list[dict]` field to `CheckCollectionResult` (defaults empty; distinct from the engine-internal `measurements: list[Measurement]`). MM callers populate it; contracts leave it empty.
- The builder collects measurement dicts across results and adds `payload["metrics"]` **only when at least one is present**.
- **Byte-identical for contracts:** when there are no measurements the `"metrics"` key is absent entirely (not null, not `[]`), so existing contract payloads are unchanged.

> Stacked on #2773 (OBSL-1013). Review/merge that first; this branch targets it.

## Test Plan
- [x] Unit: contracts-only payload omits the `metrics` key; measurements present → `metrics: [...]` populated
- [x] Regression: full unit suite (986) passes; pre-commit (isort/black/autoflake) clean
- [ ] MM tickets (OBSL-1008/1024) will populate `measurement_dicts` with v3 `metric.get_cloud_dict()`-shaped dicts


m1n0:
Reviewed as a small incremental change to the shared payload builder. **No blocking findings.** Verified: no pre-existing `metrics` key in the builder dict, the dataclass field addition is safe (default after defaults), the byte-identical-for-contracts claim holds, and both branches are pinned by tests. CI green.

One robustness point worth a look before the MM PRs build on this:

- **`payload["metrics"]` bypasses `to_jsonnable`** (added after the builder's call). Today that's rescued by an accident: `_execute_cqrs_request` runs `to_jsonnable(request_body)` **for debug logging** (`soda_cloud.py:1177`), which mutates the body in place before `requests` serializes it with stock `json.dumps`. In v3 the metrics dicts were only JSON-safe because the whole body went through `JsonHelper.to_jsonnable` (`cloud/soda_cloud.py:184→223`) — `get_cloud_dict()` values can be `Decimal`/`timedelta`/`datetime`. So an MM caller attaching raw v3-shaped dicts works today only via the logging side effect; anyone making that log line lazy breaks sends. Suggest making it explicit — `payload["metrics"] = to_jsonnable(all_measurement_dicts)` (and consciously pick `remove_null_values_in_dicts`, since the accidental pass currently strips nulls inside measurement dicts too) — or pin the "pre-serialised" contract with a test that `json.dumps(payload)` succeeds with realistic (Decimal/datetime) values.

Nit: `getattr(r, "measurement_dicts", None)` — the sibling accumulations (`post_processing_stages`, `token_usage`) use direct attribute access; the field lives on the base dataclass, so `r.measurement_dicts` works and won't hide typos.

mivds:
Both points addressed in 2057a7c2:

- **`metrics` serialisation** — made explicit: `payload["metrics"] = to_jsonnable(all_measurement_dicts)`, so it no longer relies on the debug-log `to_jsonnable` side effect mutating the shared list at the send site. Kept `remove_null_values_in_dicts` at its default `True` to match the null-stripping the payload dict already gets, and pinned the pre-serialised contract with a `json.dumps(payload)` test over `Decimal`/`datetime` values.
- **`getattr` nit** — confirmed `measurement_dicts` is a declared field on `CheckCollectionResult` (`default_factory=list`), so switched to direct `r.measurement_dicts` like the sibling accumulations.

paulteehan (approved):

PR #2775, opened 2026-06-30
feat(check-collections): global impl extensions — reconciliation in all check-collection kinds (DES-433)

m1n0:
## What & why (DES-433)

Reconciliation checks (`row_count_diff`, `rows_diff`, …) only worked inside **contracts**. A data
standard with a `reconciliation:` block parsed fine but the checks were silently dropped. Root cause:
the impl-extension registry was per-subtype isolated and reconciliation registered only on
`ContractImpl`, while the YAML-extension layer was already global. This PR makes the impl layer
symmetric.

## Changes
- **Global-by-default impl extensions.** Extensions registered on the `CheckCollectionImpl` base now
  apply to every collection kind; per-subtype registrations stay isolated. New
  `_resolve_impl_extensions()` merges the global (base) dict with the per-kind dict. The extension
  Protocol `ContractImplExtension` was renamed `CheckCollectionImplExtension` and its types broadened
  to `CheckCollectionImpl`. No back-compat alias — there are no importers outside this repo.
- **Base-level primary data-source resolution.** Moved into `CheckCollectionImpl.__init__`, so every
  kind resolves its primary data source by dataset name when the caller passes `data_source_impl=None`
  (the universal `soda contract verify` dispatch path). Previously only `ContractImpl` did this, so a
  `kind: data-standard` file run through that path failed at publish with "Data source not found".
- **`columns:` optional for data standards only.** A `columns_required` flag on the YAML class:
  `True` on `ContractYaml` (contracts still require `columns:`, matching the backend contract schema's
  `required: [dataset, columns]`), `False` on `DataStandardYaml` (a checks-only or reconciliation-only
  standard may omit it). Contract behaviour is unchanged.

## Out of scope (reverted)
An earlier revision also relaxed empty dataset-level `checks: []`. That is not needed for DES-433 — a
reconciliation-only collection *omits* the `checks:` key (reconciliation checks live under the
`reconciliation:` block), it never sends `checks: []` — and it has contract check-archival
implications. Reverted here and in the backend schema (`checks` stays `minItems: 1`).

## Tests
`soda-tests` unit + feature + integration green: new global/isolated registry tests, `columns_required`
parse tests (contract raises on missing `columns:`, data standard allows it), and a base
data-source-resolution regression test.

## Cross-repo / merge order
Part of a 3-repo DES-433 set. **soda-extensions** `#452` (reconciliation registers globally;
data-standard data-source handling; e2e) depends on this and resolves `soda-core` from PyPI, so merge +
release this first. **soda** `#12621` adds backend reconciliation-source provisioning for
`DATA_STANDARDS` scans.

Spec + plan: `soda-punch/docs/superpowers/{specs,plans}/…reconciliation-all-collection-types…`


soda-core/src/soda_core/check_collections/base.py
```diff
@@ -364,23 +364,64 @@ def for_kind(cls, kind: str) -> type["CheckCollectionImpl"]:
...
+        apply to every kind; per-subtype extensions (registered on a subclass)
+        apply only to that kind. A subtype registration overrides a global of
+        the same name. Mirrors the already-global YAML-extension layer.
+        """
+        global_exts: dict = CheckCollectionImpl.impl_extensions
+        own_exts: dict = cls.__dict__.get("impl_extensions", {})
+        return {**global_exts, **own_exts}
+
+    @staticmethod
+    def _resolve_primary_data_source_impl(
```
paulteehan, line 401: okay I think I see, this method used to exist in contract impl, we're moving into check collection impl to make it generally available for all check collection types, not just contracts

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -473,14 +473,14 @@ def _execute_on_agent(cls, *args, **kwargs) -> list[ContractVerificationResult]:
         return cls._execute_on_runner(*args, **kwargs)
 
 
-class ContractImplExtension(Protocol):
-    def __init__(self, contract_impl: ContractImpl):
-        self.contract_impl: ContractImpl = contract_impl
+class CheckCollectionImplExtension(Protocol):
```
paulteehan, line 476: so we're generalizing the extension mechanism from ContractImpl to CheckCollectionImpl?

Makes sense but in that case I would say this code should be in a file/folder with 'contracts' in the name since that's a specific subclass 

also do you maybe want to keep ContractImplExtension around as a subclass of ContractImplExtension in case you need to override it?

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -50,6 +50,11 @@ class ContractYaml(CheckCollectionYaml):
 
     contract_yaml_extensions: dict[str, type[ContractYamlExtension]] = {}
 
+    # Whether a 'columns' property is required. Contracts require it (backend
+    # contract schema: required [dataset, columns]); subtypes whose schema allows
+    # a columns-less document (e.g. data standards) override this to False.
+    columns_required: bool = True
```
paulteehan, line 56: It looks like this is not defined in the base CheckCollectionsYaml, I think it should be?  if I'm not mistaken

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -243,10 +248,15 @@ def _parse_columns(self, contract_yaml_object: YamlObject) -> list[ColumnYaml]:
...
             column_yaml_objects: Optional[YamlList] = contract_yaml_object.read_list_of_objects_opt("columns")
-            if not column_yaml_objects:
-                raise ContractParserException(
-                    "The contract is missing the required 'columns' property", str(contract_yaml_object.location)
-                )
+            if column_yaml_objects is None:
+                # Contracts require a 'columns' property; data standards (which set
+                # columns_required = False) allow it absent — their schema permits
+                # checks-only or reconciliation-only collections.
+                if self.columns_required:
```
paulteehan, line 255: I'm confused now, this appears to be a base implementation, as if data standards are a subclass of contracts, is that correct?  I thought both contracts and standards were subclasses of check collections

soda-core/src/soda_core/contracts/impl/contract_yaml.py
paulteehan: I'm mistaken :)
I checked the other PR and I see that standards inherit from contracts, not check collections
I think the code is fine, just want to confirm this is the way you want the inheritance to work

soda-core/src/soda_core/contracts/impl/contract_yaml.py
paulteehan: No I'm wrong, see previous comment

paulteehan (approved):
A few comments and questions but not blocking, looks good!

soda-core/src/soda_core/contracts/impl/contract_yaml.py
Niels-b, line 56: Will this not break contracts in production that don't have the `columns:` attribute? Like your example in the test with the recon use-case

Niels-b (commented):
LGTM. I like moving things into the base check collection to make it available in multiple locations.

Claude flagged an inconsistency between PR and the description, update at your discretion :).

One comment/question about the `columns:` attribute 🤔

soda-core/src/soda_core/contracts/impl/contract_yaml.py
m1n0: Good check. Contracts still require `columns:` — `ContractYaml.columns_required` stays `True`, which restores the original pre-PR behaviour (parse error on missing `columns:`, matching the backend contract schema's `required: [dataset, columns]`). Only `DataStandardYaml` sets it `False`, so nothing changes for contracts in production.

The old description's "columns optional for all kinds" was an earlier revision — it's been scoped down to data-standards-only, and the description is now updated.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Agree the generalized `CheckCollectionImplExtension` reads oddly in a `contract_verification_impl.py`-named file. I looked at moving it next to `CheckCollectionImpl` in `check_collections/base.py`, but its method signatures reference `CheckImpl`/`Query` (both defined in `contract_verification_impl.py`), so the move creates a base→impl circular import. The clean fix is splitting the general impl machinery out of that file — bigger than this PR, and exactly the contracts/standards structural dedup Lepo00 captured in **DES-447**; I'll fold the engine-side split in there.

On keeping `ContractImplExtension` as a subclass: skipping for now (YAGNI — nothing overrides it, and there are no importers outside the repo). Trivial to add a `ContractImplExtension(CheckCollectionImplExtension)` the moment a contract-specific override is actually needed.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
m1n0: Confirmed — `DataStandardYaml(ContractYaml)` is intentional for the POC: a data standard is a contract-shaped document plus a `name:`, so it reuses contract parsing. The cleaner long-term model (a shared `CheckCollectionYaml` base rather than standards inheriting contracts) is the DES-447 cleanup.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -473,14 +473,14 @@ def _execute_on_agent(cls, *args, **kwargs) -> list[ContractVerificationResult]:
...
+    def __init__(self, contract_impl: CheckCollectionImpl):
+        self.contract_impl: CheckCollectionImpl = contract_impl
 
-    def parse_checks(self, contract_impl: ContractImpl) -> list[CheckImpl]:
+    def parse_checks(self, contract_impl: CheckCollectionImpl) -> list[CheckImpl]:
         return []
 
-    def build_queries(self, contract_impl: ContractImpl) -> list[Query]:
+    def build_queries(self, contract_impl: CheckCollectionImpl) -> list[Query]:
         return []
```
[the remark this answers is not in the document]
m1n0: Intentional — no back-compat alias. There are no importers of `ContractImplExtension` outside this repo; the one internal user (soda-reconciliation) is updated in the paired PR to import `CheckCollectionImplExtension`. The PR description previously said "alias kept" — that was stale and is now corrected.

soda-tests/src/helpers/data_source_test_helper.py
```diff
@@ -1506,8 +1506,9 @@ def _dedent_strip_and_prepend_dataset(self, contract_yaml_str: str, test_table:
         checks_contract_yaml_str = dedent(contract_yaml_str).strip()
         if test_table:
             header_contract_yaml_str: str = f"dataset: {self.build_dqn(test_table)}\n"
-            # This asserts that any "columns" statement in a check is single line.
-            columns_contract_yaml_str: str = "columns: []\n" if not "columns:\n" in checks_contract_yaml_str else ""
+            # Contracts require a 'columns' property; inject an empty one when the
+            # test's contract doesn't declare columns (e.g. reconciliation-only).
+            columns_contract_yaml_str: str = "columns: []\n" if "columns:\n" not in checks_contract_yaml_str else ""
             checks_contract_yaml_str = header_contract_yaml_str + columns_contract_yaml_str + checks_contract_yaml_str
```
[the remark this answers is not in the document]
m1n0: Fixed in db95c3f5 — the injector now detects any top-level `columns:` (block or inline) via `re.search(r"(?m)^columns:", ...)` before injecting an empty one, so an inline `columns: []` no longer produces a duplicate key.

soda-core/src/soda_core/check_collections/base.py
m1n0: yes exactly

PR #2776, opened 2026-06-30
feat(sql-ast): window functions and UNION-in-CTE

mivds:
Ticket: OBSL-1027

## PR chain (soda-core)
1. #2773 — ship table/column discovery
2. #2774 — emit `metrics[]` in scan-results payload
3. #2776 — SQL AST window functions + UNION-in-CTE  ← this PR
4. #2777 — raw feature-config passthrough (config DTO)
5. #2787 — all-datasource dialect groundwork

---

## Summary

Adds the SQL-AST primitives the v3 numeric profiler needs, unblocking the profiling SQL port (OBSL-1017). Scope was determined by the OBSL-1014 spike.

- **`WINDOW_FUNCTION` AST node** (`sql_ast.py`) — `name`, `args`, optional `partition_by: list[COLUMN]`, optional `order_by` (reuses `ORDER_BY_ASC`/`ORDER_BY_DESC`). Renders `NAME(args) OVER ([PARTITION BY ...] [ORDER BY ...])`.
- **UNION-in-CTE** — widened `CTE.cte_query` to accept `UNION` (covers `UNION_ALL`); `_build_cte` renders it via `build_union_sql`.
- Rendering reuses existing function-arg + ORDER BY logic; additive dispatch branch in `build_expression_sql`.

Rendered output examples (asserted in tests): `ROW_NUMBER() OVER (ORDER BY "value_" DESC)`, `SUM("amount") OVER (PARTITION BY "category" ORDER BY "id" ASC)`, CTE bodies containing `UNION` / `UNION ALL`.

> Stacked on #2774 (OBSL-1025) → #2773 (OBSL-1013). Closes the AST gap from the OBSL-1014 spike.

## Test Plan
- [x] 7 new unit tests asserting rendered SQL (window fns w/ partition+order, no-clause, multi-col partition, CTE UNION/UNION ALL)
- [x] Regression: `soda-tests/tests/unit` → 993 passed, 2 skipped, 0 failures
- [x] pre-commit (isort/black/autoflake) clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2777, opened 2026-06-30
feat(soda-core): raw feature-config passthroughs on the dataset-config DTO + MM/dialect seams

mivds:
Ticket: OBSL-1019, OBSL-1007

## PR chain (soda-core)
4. #2777 — raw feature-config passthrough (config DTO)  ← this PR

---

## Summary

Cloud/dialect seams for the observability extensions (profiling, metric monitoring), stacked on #2776 (OBSL-1027).

### Dataset-config DTO: raw feature-config passthroughs
- `DatasetConfigurationDTO.profiling_configuration` (alias `profilingConfiguration`) is a raw `Optional[dict]` passthrough: the soda-profiling extension (`soda_profiling.profiling_config`, soda-extensions) owns the parsing, so soda-core stays free of profiling semantics — the same convention as `metricMonitoringConfiguration` (metric-monitoring's `monitor_config`) and `timePartitionConfiguration` (the partition detector). Wire JSON unchanged; the dict arrives and round-trips verbatim. The earlier typed `ProfilingConfigurationDTO` + sampling-strategy union from this branch was removed again in favor of the passthrough; the typed models now live in soda-extensions.
- `metric_monitoring_configuration` raw `list[dict]` field + historic-data queries + identity helpers (MM transport seams, OBSL-1007) and `get_last_measurement` single-identity historic query (OBSL-1028).
- Per-`CheckResult` cloud-dict override hook for non-contract check shapes (OBSL-1008).

### SQL AST + dialect seams
- `TIME_DELTA` + `ADD_INTERVAL` time-bucket nodes with snowflake/bigquery renderers (OBSL-1028).
- `PERCENTILE_WITHIN_GROUP` ordered-set aggregate node + bigquery `APPROX_QUANTILES` renderer (OBSL-1022).
- `sql_expr_timestamp_coerce` dialect hook for partition-window comparisons and `get_large_numeric_cast_type_name` capability for high-precision aggregate math (OBSL-1005).
- Fixes: BigQuery renders `UNION` as `UNION ALL`; BigQuery `DATETIME`/`BIGNUMERIC` + parameterized types in the soda-type reverse map; defensive Snowflake string-type aliases (OBSL-1005).

> soda-core part of OBSL-1019 and the MM/dialect seams it stacks with. The profiling orchestrator + `soda profile` CLI (and the typed profiling-config models) are the soda-extensions part (sodadata/soda-extensions#457).

## Test Plan
- [x] Unit: `profilingConfiguration` passes through as a raw dict via alias, round-trips by alias unchanged, omitted key parses to `None` (no regression); typed parsing is covered in soda-extensions (`soda-profiling/tests/test_profiling_config.py`)
- [x] `pytest soda-tests/tests/unit -q`: 1059 passed, 1 skipped; pre-commit clean


m1n0:
Reviewed (aware it's still draft) as a mixed PR, per group. Most of it is clean: DTO passthroughs round-trip by alias and the ownership rationale is documented; `observability_identity` digests are pinned and the add-order is called out as load-bearing; the historic-query trio (XOR validation, 500-identity batching, grouping) is well-tested; and both override hooks (`build_soda_cloud_check_dict`, summary-table extra columns/headers) provably leave the default contract paths unchanged. CI green.

One finding I'd fix before undrafting:

- **BigQuery `build_union_sql` renders plain `UNION` as `UNION ALL`** (`bigquery_data_source.py:174`). The docstring's justification is profiling-specific ("disjoint constant metric_ labels"), but this is a general dialect method: any future caller building `UNION(...)` on BigQuery expecting set semantics silently gets duplicates. BigQuery supports `UNION DISTINCT` — rendering `UNION` → `UNION DISTINCT` and `UNION_ALL` → `UNION ALL` is both valid BQ SQL and semantics-preserving, and callers that want the cheaper form can say `UNION_ALL` explicitly.

Smaller notes:

- `get_historic_measurements`/`get_historic_check_results` with `limit` + >500 identities applies the limit **per batch** — merged results can exceed it. Current callers won't hit it (limit is used with single identities), but worth a line in the docstring or a guard.
- Snowflake `_build_time_delta_sql` counts crossed second-boundaries (`TIMESTAMPDIFF(second, ...)`) while the base form measures true duration — sub-second window bounds can bucket off-by-one vs postgres. Immaterial if MM scan times are whole-second aligned; flagging the boundary.
- `snowflake_data_source.py:185`: the new `get_large_numeric_cast_type_name` landed between the pre-existing "TODO: test this thorough / AI-generated" comment and the type-map function it describes — the TODO now reads as if it's about the new method.
- `TIME_BUCKET_UNITS` and BigQuery's `_TIME_BUCKET_UNIT_NAMES` are two sources of truth; adding a unit to the former silently KeyErrors at BQ render time.

mivds:
Addressed in 23a9ae9d:

- **BigQuery `build_union_sql`** — plain `UNION` now renders `UNION DISTINCT` (was `UNION ALL`), preserving set semantics for any caller; `UNION_ALL` still renders `UNION ALL`. Tests updated to pin both.
- **Snowflake TODO placement** — moved the "AI-generated / test thorough" TODO back above the type-map function it describes (the new `get_large_numeric_cast_type_name` had been inserted between them).
- **Per-batch limit** — documented on both `get_historic_measurements`/`get_historic_check_results` that `limit` is applied per identity-batch, so a multi-batch merged result can exceed it (current callers pass a limit only with single identities).

Deferred (noted, not changed): the Snowflake `_build_time_delta_sql` second-boundary note is immaterial for whole-second MM scan times; the `TIME_BUCKET_UNITS` vs `_TIME_BUCKET_UNIT_NAMES` two-sources-of-truth is the same cross-dialect consolidation you flagged on #2787 — tracking it there as one follow-up rather than touching ~6 dialects in this PR.

Niels-b (changes requested):
**Verdict: Request changes.**

**Blocker (HIGH):** BigQuery `_build_add_interval_sql` maps `weeks → WEEK` and emits `TIMESTAMP_ADD(ts, INTERVAL n WEEK)`, but `WEEK` is **not** a valid part for BigQuery `TIMESTAMP_ADD` (only `MICROSECOND`..`DAY`). `weeks` is a validated, reachable unit with no guard, so weekly metric-monitoring on BigQuery will error at runtime.
*Fix:* translate `weeks → INTERVAL 7*count DAY`, or use `DATE_ADD`/`DATETIME_ADD`. (The `TIMESTAMP_DIFF` path is fine — BQ accepts `WEEK` there.)

Non-blocking: `_group_results_by_key` buckets key-less rows under `None`; `get_last_measurement` skips the drop-empty-value filter that `get_historic_measurements` applies.

---
*AI-assisted static review (read-only; not executed against a live warehouse). Assessed only this PR's incremental diff; blocking findings were adversarially re-verified.*

mivds:
@Niels-b thanks — good catch on the BigQuery blocker. Fixed in 82275df3.

`weeks` in the `ADD_INTERVAL` path now renders as `INTERVAL <count> * 7 DAY` (BigQuery `TIMESTAMP_ADD` only accepts `MICROSECOND`..`DAY`). The `TIME_DELTA`/`TIMESTAMP_DIFF` path keeps `WEEK`, which is valid there — added a test pinning that asymmetry plus one for the new weekly-interval rendering.

Niels-b (approved):
**Re-review after fix — verdict: Approve.** ✅ Blocker resolved.

`82275df3` special-cases `weeks` in BigQuery `_build_add_interval_sql` to emit `TIMESTAMP_ADD(ts, INTERVAL <count> * 7 DAY)`, avoiding the invalid `WEEK` part (BigQuery `TIMESTAMP_ADD` accepts only `MICROSECOND`..`DAY`). Verified correct **and** complete: weeks are fixed-length (7 days); the multiplier is precedence-safe (count renders parenthesized); `days`/`hours`/`seconds` are unchanged; and the `TIMESTAMP_DIFF` path correctly keeps `WEEK` (valid there). New locking tests cover both cases, and CI is green.

Remaining items are the pre-existing non-blocking nits only: `_group_results_by_key` buckets key-less rows under `None`, and `get_last_measurement` skips the drop-empty-value filter that `get_historic_measurements` applies.

---
*AI-assisted re-review after the author's fix (read-only static analysis; blocking findings independently re-verified against the code).*

PR #2779, opened 2026-07-02
fix(check-collections): report runner scan FAILED with logs on abort re-raise (SAS-13001)

m1n0:
## What & why

[SAS-13001](https://sodadata.atlassian.net/browse/SAS-13001): a post-migration contract scan failed but produced **no logs** in Cloud, making it undiagnosable.

Root cause is a diagnosability hole in the engine. A single-contract (runner) scan runs `execute_check_collections` with `abort_on_first_error=True`, so the **first** exception during contract construction or verify re-raises immediately — *before* phase 3's combined upload, which is the only place that ships the engine's captured logs to Cloud and calls `mark_scan_as_failed`. The exception then reaches the CLI (exit 3) and the launcher only writes it to pod logs, so the Cloud scan record shows FAILED with an empty log payload. This turns **any** pre-upload crash into a silent, undiagnosable scan.

## Fix
Before the `abort_on_first_error` re-raise, best-effort mark the still-PENDING runner scan FAILED with the captured logs (`_report_runner_scan_failed_before_reraise`). The shared `Logs` gatherer holds every construction/verify record (each impl is built with `logs=logs`), so its records are exactly what should reach Cloud. The historical re-raise contract is preserved and reporting never masks the original exception. No-op for ad-hoc runs (no scan id).

## Tests
- New `test_uncaught_exception_during_verify_marks_scan_failed_with_logs`: a scan that raises during verify now marks the scan FAILED with a non-empty log payload (fails on `main`, passes here).
- Full soda-core unit suite: 978 passed, 2 skipped; pre-commit clean.

## Note
This is the diagnosability fix. The *underlying* crash for users  (most likely a migrated contract with an unsupported check type) is tracked separately — this change makes that (and any future pre-upload crash) self-diagnosing in Cloud.

[SAS-13001]: https://sodadata.atlassian.net/browse/SAS-13001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/soda-core/src/soda_core/check_collections/session.py
+++ b/soda-core/src/soda_core/check_collections/session.py
@@ -201,6 +201,7 @@ def execute_check_collections(
             constructed.append((impl, impl_class, None, yaml_source))
         except Exception as exc:
             if abort_on_first_error:
+                _report_runner_scan_failed_before_reraise(soda_cloud_impl, publish_results, logs, exc)
                 raise
             constructed.append((None, impl_class, exc, yaml_source))
 
@@ -238,6 +239,7 @@ def execute_check_collections(
                 results.append(impl.verify())
             except Exception as exc:
                 if abort_on_first_error:
+                    _report_runner_scan_failed_before_reraise(soda_cloud_impl, publish_results, logs, exc)
                     raise
                 results.append(impl_class.build_error_result(yaml_source, exc))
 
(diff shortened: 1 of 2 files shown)


soda-core/src/soda_core/check_collections/session.py
```diff
@@ -398,6 +400,43 @@ def execute_check_collections(
...
+        return
+    try:
+        log_records = logs.get_log_records() if logs is not None else None
+        soda_cloud_impl.mark_scan_as_failed(
+            scan_id=soda_scan_id,
+            logs=log_records,
+            exc=exc if isinstance(exc, Exception) else None,
+        )
+    except Exception:
+        logger.debug("Could not report runner scan as failed before re-raising", exc_info=True)
```
Niels-b: Should this be a debug log? That doesn't show up in cloud 🤔 .

soda-tests/tests/unit/test_contract_marks_scan_failed_on_connection_error.py
```diff
@@ -145,3 +147,44 @@ def test_ad_hoc_run_without_scan_id_still_uploads_results(monkeypatch):
...
+
+
+def test_uncaught_exception_during_verify_marks_scan_failed_with_logs(monkeypatch):
+    """A single-contract (runner) scan that raises an *uncaught* exception during verify aborts
+    and re-raises before phase 3's combined upload runs. The engine must still report the runner
+    scan as FAILED with the captured logs, otherwise the Cloud scan record shows no logs and the
+    failure is undiagnosable (SAS-13001)."""
+    monkeypatch.setenv("SODA_SCAN_ID", "scan-under-test")
+
+    def _boom(self, *args, **kwargs):
```
Niels-b, line 159: Love the name 💥

Niels-b (approved):
LGTM. One minor comment

soda-tests/tests/unit/test_contract_marks_scan_failed_on_connection_error.py
```diff
@@ -145,3 +147,44 @@ def test_ad_hoc_run_without_scan_id_still_uploads_results(monkeypatch):
         "Ad-hoc run has no scan id to mark failed; mark_scan_as_failed must not be used. "
         f"Requests seen: {request_types}"
     )
+
```
paulteehan, line 150: it seems there are two failure pathways here, construction and verification, but only verification is tested, did you think about testing construction as well?

paulteehan (approved):

soda-core/src/soda_core/check_collections/session.py
m1n0: Agreed — bumped to `warning` in 60306898. You're right that debug is the wrong level here: this only fires if the mark-scan-failed call *itself* fails, in which case the scan stays FAILED-with-no-logs in Cloud (exactly what this helper is trying to prevent), so it needs to be visible in the pod logs / Datadog. It can't go to Cloud by nature (Cloud reporting is what just failed), and warning keeps it from being drowned out while still not masking the original exception, which is re-raised regardless.

soda-tests/tests/unit/test_contract_marks_scan_failed_on_connection_error.py
m1n0: Good call — added `test_uncaught_exception_during_construction_marks_scan_failed_with_logs` in 60306898 (raises inside `ContractImpl.__init__` so the exception escapes phase-1 construction rather than verify). Both abort points call the same helper, so this locks in that the construction pathway reports FAILED-with-logs too. 979 unit tests green.

PR #2781, opened 2026-07-06
feat(databricks): OAuth (M2M + Azure service principal) auth [PLATL-820]

paulteehan:
## Summary

Adds OAuth authentication to the `soda-databricks` (v4) connector alongside the
existing personal-access-token auth.

- New `auth_type` discriminator on the Databricks connection: `personal-access-token`
  (default / backward-compatible when absent), `databricks-oauth-m2m` (Databricks-managed
  service principal), `azure-service-principal` (Entra ID).
- OAuth modes build a Databricks SDK `credentials_provider` (`oauth_service_principal` /
  `azure_service_principal`); credential fields are stripped from `sql.connect` kwargs and
  the workspace host is normalized (scheme re-added for the SDK `Config`).
- `databricks-sdk>=0.19` added as a dependency.

## Tests

- `tests/unit/test_databricks_oauth.py` (new) — auth dispatch, credential-field stripping,
  credentials-provider construction, host-scheme handling, PAT regression. **13 pass.**
- `tests/data_sources/test_databricks.py` — added `TestConnection` config-validation cases
  (unknown `auth_type`, missing credentials) that run without live creds, plus live OAuth
  M2M / Azure-SP cases (env-guarded). The module is now import-safe without creds.

## Cross-repo

Part of PLATL-820. Pairs with the backend passthrough in `sodadata/soda` and the spark
bridge in `sodadata/soda-library`. The `auth_type` literals match across all three.

### DIFF
--- a/soda-databricks/pyproject.toml
+++ b/soda-databricks/pyproject.toml
@@ -10,6 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.16.0",
     "databricks-sql-connector",
+    "databricks-sdk>=0.19",
 ]
 
 [project.entry-points."soda.plugins.data_source.databricks"]
--- a/soda-databricks/pyproject.toml
+++ b/soda-databricks/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.16.0",
     "databricks-sql-connector",
-    "databricks-sdk>=0.19",
+    "databricks-sdk>=0.19,<1",
 ]
 
 [project.entry-points."soda.plugins.data_source.databricks"]
--- a/soda-databricks/pyproject.toml
+++ b/soda-databricks/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.16.0",
     "databricks-sql-connector",
-    "databricks-sdk>=0.117,<1",
+    "databricks-sdk>=0.19,<1",
 ]
 
 [project.entry-points."soda.plugins.data_source.databricks"]
--- a/soda-databricks/pyproject.toml
+++ b/soda-databricks/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.16.0",
     "databricks-sql-connector",
-    "databricks-sdk>=0.19,<1",
+    "databricks-sdk>=0.117,<1",
 ]
 
 [project.entry-points."soda.plugins.data_source.databricks"]
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
@@ -13,7 +13,9 @@ from soda_core.model.data_source.data_source_connection_properties import (
     DataSourceConnectionProperties,
 )
 from soda_databricks.model.data_source.databricks_connection_properties import (
+    DatabricksAzureServicePrincipal,
     DatabricksConnectionProperties,
+    DatabricksOAuthM2M,
 )
 
 logger: logging.Logger = soda_logger
@@ -27,10 +29,60 @@ class DatabricksDataSourceConnection(DataSourceConnection):
         self,
         config: DatabricksConnectionProperties,
     ):
+        connection_kwargs = config.to_connection_kwargs()
+
+        # OAuth modes: the SDK builds a credentials_provider callable (handling token
+        # acquisition + refresh). The credential fields are stripped from connection_kwargs
+        # by the properties class, so they never reach sql.connect as plain kwargs.
+        credentials_provider = self._build_credentials_provider(config, connection_kwargs)
+        if credentials_provider is not None:
+            return sql.connect(
+                user_agent_entry="Soda Core",
+                credentials_provider=credentials_provider,
+                **connection_kwargs,
+            )
+
+        # Token (PAT) auth: access_token flows through connection_kwargs unchanged.
         return sql.connect(
             user_agent_entry="Soda Core",
-            **config.to_connection_kwargs(),
+            **connection_kwargs,
         )
+
+    @staticmethod
+    def _build_credentials_provider(config: DatabricksConnectionProperties, connection_kwargs: dict):
+        """Return an SDK credentials_provider callable for OAuth modes, else None.
+
+        The Databricks SDK derives the token endpoint from the workspace host and handles
+        refresh. ``connection_kwargs['server_hostname']`` has had any scheme stripped by the
+        properties layer, so re-add ``https://`` for the SDK ``Config``.
+        """
+        if not isinstance(config, (DatabricksOAuthM2M, DatabricksAzureServicePrincipal)):
+            return None
+
+        from databricks.sdk.core import Config
+        from databricks.sdk.credentials_provider import (
+            azure_service_principal,
+            oauth_service_principal,
+        )
+
+        host = f"https://{connection_kwargs['server_hostname']}"
+
+        if isinstance(config, DatabricksOAuthM2M):
+            sdk_config = Config(
+                host=host,
+                client_id=config.client_id,
+                client_secret=config.client_secret.get_secret_value(),
+            )
+            return oauth_service_principal(sdk_config)
+
+        # DatabricksAzureServicePrincipal — Entra ID (Azure AD) service principal.
+        sdk_config = Config(
+            host=host,
+            azure_client_id=config.azure_client_id,
+            azure_client_secret=config.azure_client_secret.get_secret_value(),
+            azure_tenant_id=config.azure_tenant_id,
+        )
+        return azure_service_principal(sdk_config)
 
     def _fetch_session_timezone(self) -> tzinfo:
         with self.connection.cursor() as cursor:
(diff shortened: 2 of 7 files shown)


soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
```diff
@@ -27,10 +29,60 @@ def _create_connection(
...
+
+        host = f"https://{connection_kwargs['server_hostname']}"
+
+        if isinstance(config, DatabricksOAuthM2M):
+            sdk_config = Config(
+                host=host,
+                client_id=config.client_id,
+                client_secret=config.client_secret.get_secret_value(),
+            )
+            return oauth_service_principal(sdk_config)
```
m1n0: 🔴 **Blocker — no OAuth connection can be established: `credentials_provider` is one level too shallow.**

`oauth_service_principal(cfg)` (here) and `azure_service_principal(cfg)` (line 85) return the *header factory itself* — a `Callable[[], dict]`. But `databricks-sql-connector` expects `credentials_provider` to be a callable that **returns** a header factory. In `databricks/sql/auth/authenticators.py` (v4.2.4), `ExternalAuthProvider.__init__` does `self._header_factory = credentials_provider()` and `add_headers` then calls `self._header_factory()` per request. So passing the SDK factory directly makes `credentials_provider()` return the headers **dict**, and the per-request `dict()` raises `TypeError: 'dict' object is not callable` on the first query — neither M2M nor Azure SP can connect. (Verified against the pinned connector 4.2.4 + databricks-sdk 0.117.)

The unit tests mock `oauth_service_principal` to a sentinel string and assert only that it's forwarded, so they pass while the path is broken.

**Fix** — wrap in a zero-arg callable (the documented Databricks pattern):
```python
return lambda: oauth_service_principal(sdk_config)      # and: lambda: azure_service_principal(sdk_config)
```
and add a test that drives the returned provider through the real `ExternalAuthProvider(...).add_headers({})` (or asserts `provider()()` is a dict).

Note: soda-library#759 wires it identically and needs the same fix.

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
```diff
@@ -27,10 +29,60 @@ def _create_connection(
         self,
         config: DatabricksConnectionProperties,
     ):
+        connection_kwargs = config.to_connection_kwargs()
+
+        # OAuth modes: the SDK builds a credentials_provider callable (handling token
+        # acquisition + refresh). The credential fields are stripped from connection_kwargs
+        # by the properties class, so they never reach sql.connect as plain kwargs.
+        credentials_provider = self._build_credentials_provider(config, connection_kwargs)
+        if credentials_provider is not None:
```
m1n0, line 38: 🟠 **Silent fallthrough to a credential-less PAT connect.** `oauth_service_principal()` returns `None` when OIDC discovery yields nothing; `_build_credentials_provider` returns that `None`, so this `is not None` check fails and we fall through to `sql.connect(**connection_kwargs)` with no `access_token` — a cryptic failure for a config the user explicitly marked OAuth. Once inside the OAuth branch, raise a clear error if the built provider is falsy instead of degrading to PAT. (After the blocker fix wraps it in a lambda, the `None` just defers to a `None()` TypeError — still not a clear message.)

m1n0 (commented):
Reviewed the Databricks OAuth engine change. One **blocker** and one major inline.

The `auth_type` discriminator literals (`personal-access-token` / `databricks-oauth-m2m` / `azure-service-principal`) and the OAuth key names (`client_id`/`client_secret`/`azure_client_id`/`azure_client_secret`/`azure_tenant_id`) **match** soda-library#759 and the Cloud form (soda#12685), so that cross-repo contract is good — the "provisional literals" note can be resolved. Credential safety is sound (`SecretStr`, stripped from `sql.connect` kwargs, `.get_secret_value()` only into the SDK `Config`). The dependency addition is in the right place.

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
paulteehan: Fixed in ac86ad09 — wrapped the SDK factory in `lambda: header_factory` so `ExternalAuthProvider` gets a callable that *returns* the factory. Verified against the real connector 4.2.4: `add_headers({})` now resolves to a headers dict, while the old direct-factory shape raises `TypeError: 'dict' object is not callable` as you described. Added a test driving `provider()()` to a dict.

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
paulteehan: Fixed in ac86ad09 — the OAuth branch now raises a clear `ValueError` when the SDK yields no header factory, instead of degrading to a credential-less PAT connect. Added a None→error test.

soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
```diff
@@ -39,3 +39,47 @@ def to_connection_kwargs(self) -> dict:
...
+    client_secret: SecretStr = Field(..., description="Databricks OAuth service-principal client secret")
+
+    # Consumed by the connection layer to build the credentials_provider, never passed to sql.connect.
+    _credential_fields: ClassVar[tuple] = ("client_id", "client_secret")
+
+    def to_connection_kwargs(self) -> dict:
+        connection_kwargs = super().to_connection_kwargs()
+        for field_name in self._credential_fields:
+            connection_kwargs.pop(field_name, None)
+        return connection_kwargs
```
Niels-b, line 63: NIT: we can also do this in the connection itself before passing it to the databricks connector itself. Then we don't have to overwrite the `to_connection_kwargs`.

Really low prio NIT though 😉 .

m1n0 (approved):

soda-databricks/pyproject.toml
```diff
@@ -10,6 +10,7 @@ authors = [
 dependencies = [
     "soda-core==4.16.0",
     "databricks-sql-connector",
+    "databricks-sdk>=0.19,<1",
```
Niels-b: **Bump the `databricks-sdk` floor to a version that CI actually exercises.** `>=0.19` is far below what runs — `uv.lock` resolves `0.119.0` and the credentials-provider fix was verified against `0.117`. The OAuth path imports `databricks.sdk.credentials_provider.oauth_service_principal` / `azure_service_principal` and passes specific `Config(...)` kwargs, none of which are guaranteed at `0.19` (that release predates much of that surface). Since no CI job exercises the floor, a resolver that picks an old version could import-fail or misbehave at runtime. Suggest `databricks-sdk>=0.117,<1` (or the earliest version you've verified).

soda-databricks/src/soda_databricks/model/data_source/databricks_data_source.py
```diff
@@ -4,10 +4,25 @@
...
 from soda_databricks.model.data_source.databricks_connection_properties import (
+    DatabricksAzureServicePrincipal,
     DatabricksConnectionProperties,
+    DatabricksOAuthM2M,
     DatabricksTokenAuth,
 )
 
+# Explicit auth_type discriminator values (Trino/Snowflake/BigQuery style).
+# NOTE (provisional): these literals must match what the Backend emits in the generated
+# YAML — confirm with BE before merge. See implementation-plan.md.
```
Niels-b: **This "provisional" note is resolved — please drop it before merge.** The cross-repo contract was confirmed in review: the `auth_type` literals and key names match soda-library#759 and the Cloud form (soda#12685), so the "confirm with BE before merge" caveat no longer holds. The `implementation-plan.md` reference is also dangling (not in the repo), so this would ship as a misleading TODO. Either delete these two lines or reduce them to a plain statement of the cross-repo contract.

Niels-b (approved):
LGTM.

Noticed one small NIT when going through it. And Claude also had some minor things that seem relevant (databricks-sdk one doesn't need to be fixed in this PR, there may also be a wider reason for it).

soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
paulteehan: Fair point 👍. Keeping the strip in `to_connection_kwargs` for now — that method already owns kwarg-shaping for these props (host-scheme stripping lives there too), and the unit tests assert credentials never reach `sql.connect` at that layer. Low-prio either way; happy to move it into the connection in a follow-up if preferred. Resolving.

soda-databricks/pyproject.toml
paulteehan: Done in 0169f809 — bumped to `databricks-sdk>=0.117,<1` and regenerated `uv.lock` (`uv lock --check` passes; resolves 0.119.0).

soda-databricks/src/soda_databricks/model/data_source/databricks_data_source.py
paulteehan: Done in 0169f809 — dropped the provisional/'confirm with BE' caveat and the dangling `implementation-plan.md` reference; reduced to a plain statement of the cross-repo contract.

soda-databricks/pyproject.toml
paulteehan: Correction after digging in: reverted this to `databricks-sdk>=0.19,<1` (soda-core 44e24d99; soda-library follow-up). Checked the SDK source at tag `v0.19.0` — the OAuth surface we import (`oauth_service_principal`, `azure_service_principal`, and `Config(client_id/client_secret/azure_*)`) is **already present at 0.19.0**. So `0.117` (a 2026-06 release) would exclude ~2 years of compatible releases with no benefit and risk resolver conflicts. Kept the `<1` major cap — that's the useful guard. Happy to set a higher floor if you'd want behavioral (not just API) coverage of a specific version.

PR #2782, opened 2026-07-07
feat(cloud): emit dataset schema metadata when a schema check runs

m1n0:
## Summary

v4 contracts / data standards now push the discovered dataset schema to Soda Cloud when a **schema check** runs. `_build_check_collection_results_json_dict` gains a top-level `metadata[]` block on the `sodaCoreInsertScanResults` payload:

```json
{ "datasetQualifiedName": "<datasource>/<prefix…>/<dataset>",
  "schema": [ { "columnName": "id", "sourceDataType": "integer" } ] }
```

- Reuses the **v3 discover** payload shape: raw, un-parameterized data type (`SqlDataType.name`, e.g. `character varying` — not `character varying(255)`); `isPrimaryKey` omitted (the v4 schema check doesn't discover PKs; omitting preserves the backend's existing PK flags).
- `datasetQualifiedName` = the collection's `soda_qualified_dataset_name`, which resolves to the exact dataset the contract checks registered — so the backend runs its existing `syncDatasetsColumns` reconciliation against the right dataset.
- One entry **per distinct dataset** that ran a schema check; **omitted entirely** when no schema check discovered columns (an empty schema is never sent — the backend full-reconciles and would otherwise soft-delete every column).

Without this, a schema check's discovered columns only ever reach Cloud as per-check-result diagnostics (`diagnostics.v4.actual`); nothing refreshes the dataset's own column list.

## Requires (end-to-end)

Backend counterpart: **sodadata/soda#12702** — the contract scan handler must ingest this `metadata[]` block. This engine PR is inert until that lands. (Ties into DTL-1807.)

## Known limitation

`SqlDataType` lowercases type names; the backend compares `sourceDataType` case-sensitively. **Postgres is identical to v3.** For warehouses whose catalog returns uppercase types (e.g. Snowflake `TEXT`/`NUMBER`), the first contract scan of a previously-onboarded dataset surfaces a one-time, self-converging "type changed" event. This is pre-existing shared v3/v4 comparison behavior, not introduced here.

## Rebased 2026-08-25

Rebased onto `main` (v4.22.1rc0), 45 commits of drift. One conflict: `main` restructured this builder from `return to_jsonnable({...})` into `payload: dict = {...}` + `to_jsonnable(payload)` + a conditional `payload["metrics"]`. Resolved onto `main`'s structure, with `"metadata"` re-inserted as a payload key. Semantics unchanged — `to_jsonnable` defaults to `remove_null_values_in_dicts=True` and mutates in place, so `… or None` still strips the key off the wire.

## Test plan

- [x] `soda-tests/tests/unit/test_dataset_metadata_wire.py` — metadata present with the right shape; un-parameterized type; absent with no schema check; absent when the schema check found no columns; one entry per dataset in a combined multi-file session. 5 passed.
- [x] Full unit suite after the rebase: **1277 passed, 2 skipped** (no regressions in the surrounding wire tests).

DTL-1807

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1588,6 +1588,56 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
     return []
 
 
+def _build_dataset_metadata_json_dicts(results: list[ContractVerificationResult]) -> list[dict]:
+    """Top-level ``metadata`` array that refreshes each dataset's column schema
+    in Soda Cloud.
+
+    One entry per distinct dataset that ran a schema check which discovered a
+    non-empty column set. Reuses v3 discover's ``{columnName, sourceDataType}``
+    shape so the backend runs the same full column reconciliation (columns
+    absent from ``schema`` are soft-deleted) — so we send every discovered
+    column, in discovery order, and never an empty schema.
+
+    ``datasetQualifiedName`` is the collection's ``soda_qualified_dataset_name``
+    (the raw ``yaml.dataset`` dqn), identical to the identifier the backend
+    builds from each check, so the block resolves to the same dataset the checks
+    registered under. ``sourceDataType`` is the raw, un-parameterized DB type
+    (``SqlDataType.name``, e.g. ``character varying``) — not the parameterized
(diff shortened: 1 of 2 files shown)


m1n0:
Closing in favor of #2859, which does the same thing with a cleaner seam and real e2e coverage.

PR #2787, opened 2026-07-12
feat(dialect): all-datasource groundwork — time/interval, percentile, typed timestamp literal, NaN filter

mivds:
Ticket: OBSL-1036

## PR chain (soda-core)
5. #2787 — all-datasource dialect groundwork  ← this PR

---

## What & why ([OBSL-1036](https://linear.app/sodadata/issue/OBSL-1036))

Extends the v4 profiling and metric-monitoring engines to **all** data sources, porting the per-warehouse SQL forms from the v3 engine (soda-library) into the v4 `SqlDialect` layer: time-bucket arithmetic (`TIME_DELTA` / `ADD_INTERVAL`), the percentile aggregate (`PERCENTILE_WITHIN_GROUP`), a typed timestamp literal, and a NaN-exclusion predicate. Every override cites its v3 source (`file:line`) in its docstring, per the existing convention.

## New base seams (soda-core `SqlDialect`)

- **`literal_timestamp_typed(dt)`** — typed timestamp literal for use *inside* timestamp arithmetic. Base `TIMESTAMP 'YYYY-MM-DD HH:MM:SS'` = v3 `sql_time_filter_to_timestamp` (data_source.py:1524-1529), incl. sub-second truncation; byte-parity with the MM bulk builder's `_timestamp_anchor_literal`.
- **`sql_expr_is_not_nan(expr)`** — NaN-exclusion predicate for float aggregates; base `None` = no filter needed (v3 spark_data_source.py:427-488).
- **`supports_percentile_within_group()`** — base `True`; consumers skip the Q1/median/Q3 metrics when `False` instead of rendering invalid SQL. SQL Server version-gates this at runtime (the `APPROX_PERCENTILE_DISC` aggregate is 2022+/Azure only — see deviations).

These base seams are consumed by the downstream profiling and metric-monitoring engines.

## Per-dialect forms

| Dialect | TIME_DELTA | ADD_INTERVAL | PERCENTILE_WITHIN_GROUP | Other |
|---|---|---|---|---|
| SQL Server | `DATEDIFF(second, s, e) / n` (v3 :710-716) | `DATEADD(UNIT, c, ts)` (v3 :725-727) | `APPROX_PERCENTILE_DISC(p) WITHIN GROUP (...)` (v3 :336-337); `supports_percentile_within_group()` version-gated ⚠️ | `literal_timestamp_typed` → `CAST('...' AS DATETIME2)` ⚠️ |
| Synapse | inherited | inherited | `supports_percentile_within_group()` → `False` ⚠️ | typed literal inherited |
| Fabric | inherited (pinned) | inherited (pinned) | `supports_percentile_within_group()` → `True` (pinned; Fabric engine is always current) | |
| Databricks | `FLOOR(DATEDIFF(SECOND, s, e) / n)` (v3 spark :1158-1161) | `TIMESTAMPADD(UNIT, c, ts)` ⚠️ | no override — base `PERCENTILE_DISC(p) WITHIN GROUP` valid since DBR 11.3 (pinned) | `sql_expr_is_not_nan` → `NOT ISNAN(expr)` |
| SparkDf / DatabricksHive | inherited (pinned) | inherited (pinned) | inherited | inherited |
| Trino | `cast(floor(date_diff('second', s, e) / n.0) as int)` (v3 :269-271) | `date_add('unit', c, ts)` (v3 :280-281) ⚠️ | `approx_percentile(expr, p)` (v3 :245-246) | |
| Athena | same as Trino ⚠️ | same as Trino ⚠️ | `approx_percentile(expr, p)` (v3 :256-259) | |
| Redshift | no override — base epoch-floor IS v3's redshift form (v3 :287-290, pinned) | no override — v3 shipped the base interval-multiply on Redshift (pinned) | `APPROXIMATE PERCENTILE_DISC(p) WITHIN GROUP (...)` (v3 :224-225) | |

## Deviations from v3 (⚠️ above), with doc citations

- **T-SQL TIME_DELTA truncates toward zero** (v3-verbatim `DATEDIFF(second,...)/n` int division) while every other dialect FLOORs toward -inf. Equal only for non-negative deltas — guaranteed by the MM bulk query's `ts >= anchor` filter; a consumer relying on negative bucket indices would misbucket on the T-SQL family only. Also noted: `DATEDIFF(second,...)` overflows at ~68-year spans (`DATEDIFF_BIG` is the escape hatch; not switched, v3 parity).
- **SQL Server `literal_timestamp_typed`**: v3 rendered a bare string (:699-703). T-SQL has no `TIMESTAMP '...'` literal (`TIMESTAMP` = deprecated rowversion), so the seam renders `CAST('...' AS DATETIME2)` ([DATETIME2 docs](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime2-transact-sql)).
- **SQL Server percentile version gate**: v3 emitted `APPROX_PERCENTILE_DISC` unconditionally (:336-337), which errors on SQL Server ≤2019 (the aggregate is [2022+/Azure only](https://learn.microsoft.com/en-us/sql/t-sql/functions/approx-percentile-disc-transact-sql)). Instead of inheriting that assumption, the data source probes `SERVERPROPERTY('ProductMajorVersion')`/`('EngineEdition')` on connect and injects the result into `supports_percentile_within_group()` (`major >= 16`, or Azure DB/MI editions 5/8) — strictly safer than v3 without regressing 2022+/Azure. Unprobed dialects (pure SQL rendering, no connection) assume support; a probe failure falls back to assuming support and never fails the scan. Fabric pins the flag `True` (engine is always current); Synapse pins `False`.
- **Synapse percentile**: v3 warned and returned a dummy `1` (:56-58). v4 skips the metric via the support flag; Synapse is absent from [APPROX_PERCENTILE_DISC's applies-to list](https://learn.microsoft.com/en-us/sql/t-sql/functions/approx-percentile-disc-transact-sql).
- **Databricks ADD_INTERVAL**: v3 inherited the base `ts + INTERVAL '1 <unit>' * n`; Spark parses the plural-unit string as a legacy CalendarInterval, so v4 uses `TIMESTAMPADD`, documented on [Databricks DBR 10.4+](https://docs.databricks.com/en/sql/language-manual/functions/timestampadd.html) and OSS Spark 3.3+ (SPARK-38195). The 3-arg `DATEDIFF` used for TIME_DELTA is the documented [TIMESTAMPDIFF synonym](https://docs.databricks.com/en/sql/language-manual/functions/datediff3.html) (OSS: SPARK-38389).
- **Trino/Athena units lowercased**: v3 trino passed uppercase enum names; the [Trino datetime docs](https://trino.io/docs/current/functions/datetime.html) list lowercase units (v3 athena already used them).
- **Athena TIME_DELTA int cast**: v3 athena (:269-273) lacked v3-trino's `as int` cast; it is required because `floor()` returns `double` on the Trino engine and `date_add`'s value argument is integer-typed. [Athena engine v3 functions are Trino's](https://docs.aws.amazon.com/athena/latest/ug/functions-env3.html).
- **Redshift base forms kept**: interval-literal arithmetic incl. plural units and `interval * numeric` are documented ([interval types](https://docs.aws.amazon.com/redshift/latest/dg/r_interval_data_types.html), [interval literals](https://docs.aws.amazon.com/redshift/latest/dg/r_interval_literals.html)); `APPROXIMATE PERCENTILE_DISC` per [its docs](https://docs.aws.amazon.com/redshift/latest/dg/r_APPROXIMATE_PERCENTILE_DISC.html).

## Tests

Per-package snapshot tests pin every rendered form (incl. inheritance-pinning tests for Synapse/Fabric/SparkDf and the deliberate no-override cases on Databricks/Redshift), plus base-seam unit tests in `soda-tests/tests/unit/test_sql_dialect_seams.py`.

`soda-tests` unit (1191 passed, 2 skipped), feature (13 passed) and integration (164 passed, 14 skipped) green, plus all touched-package unit suites.

### DIFF
--- a/docs/superpowers/specs/2026-07-24-sqlserver-version-guard-design.md
+++ b/docs/superpowers/specs/2026-07-24-sqlserver-version-guard-design.md
@@ -1,0 +1,145 @@
+# SQL Server version-based capability derivation — design
+
+**Date:** 2026-07-24
+**Branch:** `obsl-1036-dialect-groundwork-all-datasources` (PR #2787)
+**Status:** approved
+
+## Problem
+
+The current `APPROX_PERCENTILE_DISC` version guard (commit `055176ce`) inverts the
+desired responsibility split: `SqlServerDataSourceImpl` probes the engine
+(`SERVERPROPERTY('ProductMajorVersion')` / `EngineEdition`), derives a capability
+boolean in `_engine_supports_approx_percentile_disc`, and injects that pre-chewed
+boolean into the dialect via `set_approx_percentile_disc_supported()`. The dialect
+never sees the version, so every future version-dependent capability needs its own
+probe/setter pair in the connector.
+
+## Goal
+
+Mirror the Oracle pattern from soda-extensions #2795: the connection detects raw
+server facts, the data source syncs them onto the dialect, and the dialect derives
+capabilities itself. The connector computes no capabilities; the dialect touches no
+connection.
+
+## Decisions (with rationale)
+
+1. **The dialect receives two raw facts** — `server_major_version` and
+   `engine_edition` — not a derived boolean and not a normalized version. A version
+   number alone cannot decide this capability: Azure SQL Database (EngineEdition 5)
+   and Managed Instance (EngineEdition 8) report legacy `ProductMajorVersion` 12 but
+   support `APPROX_PERCENTILE_DISC`; on-prem needs 2022+ (major ≥ 16).
+2. **Detection is eager, at connection open, in the connection class.**
+   `server_major_version` comes free from the pyodbc login handshake
+   (`connection.getinfo(pyodbc.SQL_DBMS_VER)`); `engine_edition` costs one
+   `SELECT CAST(SERVERPROPERTY('EngineEdition') AS INT)` query. Lazy detection was
+   rejected: reading the connection during SQL generation breaks snapshot replay
+   (see the explicit warning in `oracle_data_source.py`).
+3. **Facts are synced onto the dialect after creation (Oracle-style setter/sync).**
+   Constructor injection was explored and rejected: the base `DataSourceImpl.__init__`
+   creates the dialect before any connection exists, and contract parsing /
+   `only_validate_without_execute` read `sql_dialect` without ever opening a
+   connection — so constructor injection would force replacing the dialect instance
+   at connection open, which we don't want. A lazy `server_info_provider` callable
+   was also explored and rejected as over-engineered.
+4. **SQL-Server-local pattern.** No generalization into the `SqlDialect` base in
+   soda-core. Oracle and SQL Server are two local precedents; unify in core when a
+   third dialect needs it (rule of three).
+
+## Design
+
+### 1. Detection — `SqlServerDataSourceConnection`
+
+After the pyodbc connection is created in `_create_connection`:
+
+- `self.server_major_version: Optional[int]` — parsed from
+  `connection.getinfo(pyodbc.SQL_DBMS_VER)` (e.g. `"15.00.4123"` → `15`) via a
+  static `_parse_server_major_version` helper (same shape as Oracle's).
+- `self.engine_edition: Optional[int]` — one `SERVERPROPERTY('EngineEdition')` query.
+
+Any detection failure logs a warning and leaves the attribute `None`; detection never
+fails an otherwise healthy connect. Fabric/Synapse connections inherit this; the one
+extra query at their connect is harmless (their dialects pin the capability, so a
+`None` there is inert).
+
+### 2. Sync — `SqlServerDataSourceImpl`
+
+The current `open_connection` override (probe + boolean push) is replaced by
+`_sync_dialect_server_info()`, called from:
+
+- `open_connection()`, after `super().open_connection()`, and
+- `__init__` when a live connection is passed in (the
+  `create_copy_with_different_connection` path — Oracle has the same two call sites).
+
+It copies both attributes onto the dialect when
+`isinstance(conn, SqlServerDataSourceConnection)` and
+`isinstance(self.sql_dialect, SqlServerSqlDialect)`. The current
+`type(...) is SqlServerSqlDialect` exact-type guard disappears: syncing facts onto
+Fabric/Synapse dialects is inert because they override the capability method.
+
+### 3. Derivation — `SqlServerSqlDialect`
+
+`__init__` sets `self.server_major_version: Optional[int] = None` and
+`self.engine_edition: Optional[int] = None`. The old seam
+(`_approx_percentile_disc_supported`, `set_approx_percentile_disc_supported`,
+`_probe_approx_percentile_disc_support`, `_engine_supports_approx_percentile_disc`)
+is deleted. The decision matrix moves into the dialect:
+
+```python
+def supports_percentile_within_group(self) -> bool:
+    if self.server_major_version is None and self.engine_edition is None:
+        return True  # no live server facts (offline rendering, unit tests,
+                     # snapshot replay): assume newest engine, like Oracle
+    return (
+        self.server_major_version is not None
+        and self.server_major_version >= SQLSERVER_2022_MAJOR_VERSION
+    ) or self.engine_edition in (
+        AZURE_SQL_DATABASE_ENGINE_EDITION,      # 5
+        AZURE_SQL_MANAGED_INSTANCE_ENGINE_EDITION,  # 8
+    )
+```
+
+Named module constants replace the magic numbers (`SQLSERVER_2022_MAJOR_VERSION = 16`,
+editions 5 and 8).
+
+One deliberate semantic change: the old code distinguished a *failed probe*
+(→ assume support) from a *live server reporting NULL for both properties*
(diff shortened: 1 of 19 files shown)


m1n0:
Reviewed (aware it's draft). **No blocking findings** — this is the best-documented PR of the chain: every override cites its v3 source, every deviation carries a vendor-doc link, and the claimed pins all exist (checked: fabric-inherits-sqlserver, sparkdf-inherits-databricks, synapse `False` flag, redshift deliberate no-override). The SQL forms check out against engine semantics — boundary-counting DATEDIFF/TIMESTAMPDIFF handled via seconds everywhere, Trino/Athena `as int` cast justified, T-SQL truncation-vs-FLOOR deviation honestly scoped to non-negative deltas. CI green.

Minor notes:

- **`literal_timestamp_typed` silently drops tzinfo** (`sql_dialect.py` base + the SQL Server DATETIME2 override): `dt.strftime(...)` renders wall-clock time, so a tz-aware non-UTC datetime renders its local time with the zone discarded — wrong against UTC-typed columns. Suggest normalizing tz-aware input (`dt.astimezone(timezone.utc)`, like `_convert_scan_time_to_str` does) or asserting/documenting naive-UTC-only.
- **T-SQL `_build_time_delta_sql` is the only form that isn't self-contained**: `DATEDIFF(second, s, e) / m` unparenthesized is precedence-fragile if a caller embeds TIME_DELTA in larger arithmetic; every other dialect's form is wrapped in `FLOOR()`/`cast()`. Parens are free.
- **`_TIME_BUCKET_UNIT_NAMES` now has ~6 near-identical copies** across dialects. Follow-up simplification: one shared plural→singular map next to `TIME_BUCKET_UNITS` in `sql_ast`, with dialects choosing only casing/quoting — also collapses the drift risk noted on #2777.
- SQL Server's `APPROX_PERCENTILE_DISC` needs 2022+/Azure — v3 parity, so fine, but the Synapse-style `supports_percentile_within_group()` flag is the natural gate if older-version support ever comes up.

mivds:
Addressed in cefd991e:

- **`literal_timestamp_typed` dropping tzinfo** — tz-aware datetimes are now normalized to UTC before rendering (`dt.astimezone(timezone.utc)`), via a shared `_typed_timestamp_str` helper on the base dialect that the SQL Server DATETIME2 override reuses. Naive datetimes are unchanged (assumed UTC). Added base + SQL Server tests for the tz-aware case.
- **T-SQL `_build_time_delta_sql` precedence** — wrapped the `DATEDIFF(second, …) / m` form in parens so it stays self-contained when embedded in larger arithmetic, matching every other dialect's `FLOOR`/`cast` wrapping.

Deferred as follow-ups: the ~6 `_TIME_BUCKET_UNIT_NAMES` copies → one shared plural→singular map (also resolves the drift risk noted on #2777); `APPROX_PERCENTILE_DISC` needing 2022+/Azure is v3 parity, with `supports_percentile_within_group()` already the natural gate if older-version support comes up.

Niels-b (changes requested):
**Verdict: Request changes.**

**Blocker (MEDIUM):** SqlServer `_build_percentile_within_group_sql` emits `APPROX_PERCENTILE_DISC(...) WITHIN GROUP (...)`, which is **SQL Server 2022+ only**, while `supports_percentile_within_group()` returns `True` unconditionally for SqlServer. That capability flag is the seam downstream consumers rely on, so it must not advertise support it can't honor on older servers — the fix belongs here, in the PR that introduces the seam, not in the consumer.
*Fix:* version-gate the SqlServer flag (or don't emit the unsupported form).

Note: the `observability_identity.py` deletion is self-consistent (no remaining references in soda-core) — just coordinate the module move with the extensions PRs (#488) at merge time.

---
*AI-assisted static review (read-only; not executed against a live warehouse). Assessed only this PR's incremental diff; blocking findings were adversarially re-verified.*

mivds:
@Niels-b addressed the SqlServer percentile blocker in 1177442e.

Rather than dropping the capability, I version-gated it (confirmed v3 emitted `APPROX_PERCENTILE_DISC` unconditionally, so this is strictly safer without regressing parity on 2022+/Azure): the data source probes `SERVERPROPERTY('ProductMajorVersion')`/`('EngineEdition')` on connect and injects the result into `supports_percentile_within_group()` (major >= 16, or Azure DB/MI editions 5/8). Fabric pins the capability `True` (engine is always current); Synapse stays `False`. Unprobed dialects (pure SQL rendering) assume support so rendering paths are unaffected, and a probe failure falls back to assuming support rather than failing the scan. Tests cover the version matrix + the Fabric pin.

Niels-b (approved):
**Re-review after fix — verdict: Approve (with nits).** ✅ Blocker resolved.

`1177442e` makes SqlServer `supports_percentile_within_group()` return a *probed* capability rather than unconditional `True`, so it no longer advertises `APPROX_PERCENTILE_DISC` (SQL Server 2022+ only) as supported on older servers.

Two non-blocking items before merge:
- **CI is red** — a pre-commit/formatting check is failing on the newly added code; please get it green.
- The probe-failure fallback defaults to `True` (the less-safe direction). Consider defaulting to `False` when the version probe can't determine support, so an unknown server degrades gracefully rather than emitting unsupported SQL.

---
*AI-assisted re-review after the author's fix (read-only static analysis; blocking findings independently re-verified against the code).*

PR #2788, opened 2026-07-12
Redact SQL Server connection parameter values in logs

mivds:
Continuation of #2762. CI fails on that PR because it comes from an external contributor


mivds:
Merged in #2762

PR #2789, opened 2026-07-13
fix(redshift): detect existing DWH schema via svv_all_schemas

Niels-b:
## Summary

On Redshift, setting up the Diagnostics Warehouse (DWH) could fail with `permission denied for database <db>` when the target DWH schema already exists but the connecting user is **not its owner**.

## Root cause

The DWH schema-existence check queried `information_schema.schemata`. On Redshift (Postgres-8 lineage) that view only lists schemas the current user **owns** — `USAGE`/`CREATE` grants do **not** make a schema appear there. So a pre-created schema the service user could access came back as "missing", Soda issued `CREATE SCHEMA`, which needs `CREATE` on the **database**, and failed even though the schema already existed.

Because visibility depends on ownership rather than grants, the failure can look inconsistent across otherwise-identical setups: a user who owns the schema succeeds, while a user with only `USAGE`/`CREATE` on the same schema does not.

## Fix

Override the schemas metadata query on Redshift to use `pg_catalog.svv_all_schemas` (visible by **access**, not ownership), consistent with the existing `svv_tables` / `svv_columns` overrides. The base dialect is unchanged (still `information_schema.schemata`) via a new `build_schemas_metadata_from_clause()` hook — no other data source is affected.

Result: the pre-existing schema is detected, `CREATE SCHEMA` is skipped, and no database-level `CREATE` privilege is required.

## Verification

- **Unit** — new test asserts Redshift emits `svv_all_schemas`; Postgres/base output unchanged.
- **Live Redshift** — `test_schema_exists` returns correct results; reproduced the visibility gap directly (`information_schema.schemata` listed fewer schemas than `svv_all_schemas`, missing ones the user could access, including `public`).
- **`soda-redshift/tests`** — 13 passed. **Core integration suite** — 168 passed, 8 skipped.
- **soda-extensions DWH suite** against live Redshift (this branch) — DWH tests passed (config, check_results table, views, rows-diff, failed-rows, failed-keys).

## CI

soda-extensions `main` workflow dispatched against this branch to confirm the DWH tests pass on Redshift:

- **Run:** https://github.com/sodadata/soda-extensions/actions/runs/29255799995
- Inputs: `dataSource=redshift`, `snapshotMode=off` (replay off / real DB), cross-source disabled.

### DIFF
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -1576,6 +1576,17 @@ class SqlDialect:
             ]
         )
 
+    def build_schemas_metadata_from_clause(self, table_namespace: Optional[DataSourceNamespace] = None) -> FROM:
+        """
+        FROM clause for the schemas metadata query. Defaults to information_schema.schemata.
+
+        Override in dialects whose schema listing must not come from information_schema.
+        (e.g. Redshift, where information_schema.schemata is visibility-scoped to the schema
+        OWNER and therefore misses schemas the current user only has USAGE/CREATE on.)
+        """
+        information_schema_namespace_elements = self.information_schema_namespace_elements(table_namespace)
+        return FROM(self.table_schemata()).IN(information_schema_namespace_elements)
+
     def build_schemas_metadata_query_str(
         self, table_namespace: Optional[DataSourceNamespace] = None, filter_on_schema_name: Optional[str] = None
     ) -> str:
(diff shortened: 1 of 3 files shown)


Niels-b:
CI: soda-extensions `main` workflow dispatched against this branch (Redshift, `snapshotMode=off`, cross-source disabled) to verify the DWH tests pass — https://github.com/sodadata/soda-extensions/actions/runs/29255799995

nielsn (approved):

PR #2793, opened 2026-07-15
chore: bump pyasn1 to 0.6.4 (CVE-2026-59884/59885/59886)

m1n0:
## What

`uv lock --upgrade-package pyasn1`: 0.6.3 → 0.6.4.

## Why

pyasn1 0.6.3 is affected by three DoS advisories published this week — CVE-2026-59884 (unbounded long-form tag IDs), CVE-2026-59885 (quadratic OID processing), CVE-2026-59886 (REAL value resource consumption) — all fixed in 0.6.4. Closes the Aikido finding on this repo (AIKIDO-2026-27979 bundle, flagged "needs human review" in the 2026-07-13 repo vulnerability assessment because the advisories were unpublished at the time).

Wanted on main before the `release_20260715_1` core bump so it rides 4.17.1 (tracking issue sodadata/tools-release-buddy#93). Both launchers already lock 0.6.4 on their mains.

The `exceptiongroup` marker hunk is uv 0.10.4 normalizing the entry on re-lock; no resolution change.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -506,7 +506,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/50/79/66800aadf48771f6b62f7eb014e352e5d06856655206165d775e675a02c9/exceptiongroup-1.3.1.tar.gz", hash = "sha256:8b412432c6055b0b7d14c310000ae93352ed6754f70fa8f7c34141f91c4e3219", size = 30371, upload-time = "2025-11-21T23:01:54.787Z" }
 wheels = [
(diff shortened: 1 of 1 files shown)


uv.lock
```diff
@@ -506,7 +506,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
```
Niels-b, line 509: So package only gets updated on python 3.11 and higher? 

Just to verify; which version do we use in runner?

m1n0: No behavior change here — this hunk is a uv-version side effect, not something we added. `exceptiongroup` is only in the lock as pytest's backport dependency for Python <3.11 (3.11+ has `ExceptionGroup` in the stdlib, so it's never installed there). Upstream declares `typing-extensions; python_version < "3.13"`, and the newer uv the upgrade ran with normalizes that edge to the intersection (`< 3.11`) instead of leaving it unconditional like the previous lock. Resolution is identical, and it's a test-only dep anyway.

Re the runner: it doesn't install from this uv.lock at all — this file only drives soda-core dev/CI. The launcher images compile their own requirements and already pin pyasn1 0.6.4 on main; this PR just brings soda-core's own lockfile in line so the Aikido repo finding closes.

Niels-b (approved):

PR #2795, opened 2026-07-15
feat: Oracle 18c/19c support (soda-core dialect capability + test infra)

Niels-b:
## Oracle 18c/19c support — soda-core side

Part of cross-repo Oracle 18c/19c support. This PR carries the soda-core-side dialect capability and test infrastructure; the Oracle dialect/adapter changes live in the paired **soda-extensions** PR on the matching `r-606-oracle-19-support` branch.

### Why
Oracle < 23ai lacks several SQL features that 23ai added (native `BOOLEAN`, multi-row `VALUES`, `IF [NOT] EXISTS`, FROM-less `SELECT`, ...). Supporting Oracle 18c/19c requires a small, additive capability hook in the core dialect plus version-aware test helpers.

### Changes (additive; non-Oracle behavior unchanged)
- `SqlDialect.supports_native_boolean()` — default `True`. Data sources whose `BOOLEAN` cannot round-trip (Oracle < 23ai stores it as `NUMBER`, indistinguishable from other numerics on read-back) override to `False`.
- `DataSourceTestHelper.select_literal_query()` — portable "select a literal" query for tests (Oracle appends `FROM DUAL`); `get_column_mappings()` expects `NUMERIC` for the boolean column when the source has no native boolean.
- `TestConnection` health-check probe moved into the helper (`_connection_test_query`) instead of a raw FROM-less `SELECT 1`.
- Tests: data-type and rows_tested tests made capability/version-aware; added an Oracle-gated schema-check test asserting a `data_type: boolean` contract passes.

### Validation
Validated against **Postgres, Oracle 23ai, and Oracle 18c** (18c is a proxy for 19c — same 12.2 kernel family): integration suite, DWH (both transfer modes), and connection/schema/rows_tested tests — all green, no regressions on non-Oracle sources.

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
@@ -726,7 +726,11 @@ class MissingAndValidity:
                     AND([NOT(IN(column_expression, literal_values)), IS_NOT_NULL(column_expression)])
                 )
             elif not self.valid_values:
-                invalid_clauses.append(AND([LITERAL(True)]))
+                # No value is valid -> every row is invalid. Use a portable always-true predicate
+                # (1 = 1) rather than a bare boolean literal: a standalone TRUE/FALSE is not a valid
+                # condition on data sources without a native boolean type in predicate position
+                # (e.g. pre-23ai Oracle, where it would render as `1` -> ORA-00920).
+                invalid_clauses.append(EQ(LITERAL(1), LITERAL(1)))
             else:
                 invalid_clauses.append(NOT(IN(column_expression, literal_values)))
         if isinstance(self.invalid_values, list):
@@ -734,7 +738,9 @@ class MissingAndValidity:
             if None in self.invalid_values:
                 invalid_clauses.append(AND([IN(column_expression, literal_values), IS_NULL(column_expression)]))
             elif not self.invalid_values:
-                invalid_clauses.append(AND([LITERAL(False)]))
+                # No value is explicitly invalid -> always-false predicate (1 = 0); see the
+                # valid_values note above on why a bare boolean literal is not portable.
+                invalid_clauses.append(EQ(LITERAL(1), LITERAL(0)))
             else:
                 invalid_clauses.append(IN(column_expression, literal_values))
         if isinstance(self.valid_format, RegexFormat) and isinstance(self.valid_format.regex, str):
--- a/soda-tests/src/helpers/test_connection.py
+++ b/soda-tests/src/helpers/test_connection.py
@@ -84,14 +84,15 @@ class TestConnection:
                 # do not try to query if connection failed
                 return
 
+            test_query = self._connection_test_query(data_source_impl)
             if self.query_should_succeed:
-                data_source_impl.execute_query(self._connection_test_query(data_source_impl))
+                data_source_impl.execute_query(test_query)
                 if logs.has_errors:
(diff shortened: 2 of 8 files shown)


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -726,15 +726,21 @@ def is_invalid_expr(self, column_expression: str | COLUMN | SqlExpressionStr) ->
                     AND([NOT(IN(column_expression, literal_values)), IS_NOT_NULL(column_expression)])
                 )
             elif not self.valid_values:
-                invalid_clauses.append(AND([LITERAL(True)]))
+                # No value is valid -> every row is invalid. Use a portable always-true predicate
```
paulteehan, line 729: Seems like this could gate on the new `supports_native_boolean` to use either, and then the comment is unnecessary

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -726,15 +726,21 @@ def is_invalid_expr(self, column_expression: str | COLUMN | SqlExpressionStr) ->
...
+                invalid_clauses.append(EQ(LITERAL(1), LITERAL(1)))
             else:
                 invalid_clauses.append(NOT(IN(column_expression, literal_values)))
         if isinstance(self.invalid_values, list):
             literal_values = [LITERAL(value) for value in self.invalid_values if value is not None]
             if None in self.invalid_values:
                 invalid_clauses.append(AND([IN(column_expression, literal_values), IS_NULL(column_expression)]))
             elif not self.invalid_values:
-                invalid_clauses.append(AND([LITERAL(False)]))
+                # No value is explicitly invalid -> always-false predicate (1 = 0); see the
```
paulteehan, line 741: Also here to use `supports_native_boolean`

soda-tests/src/helpers/test_connection.py
```diff
@@ -44,6 +44,13 @@ def create_data_source_yaml(self) -> DataSourceYamlSource:
     def create_data_source_impl(self, data_source_yaml_source: DataSourceYamlSource) -> DataSourceImpl:
         return DataSourceImpl.from_yaml_source(data_source_yaml_source)
 
+    def _connection_test_query(self, data_source_impl: DataSourceImpl) -> str:
+        # A trivial query to verify the live connection. Oracle requires a FROM clause
+        # (FROM DUAL is valid on every Oracle version); other data sources accept a bare SELECT.
+        if data_source_impl.type_name == "oracle":
```
paulteehan: Isn't this meant to use the new `select_literal_query` in data_source_test_helper?

paulteehan (approved):
Minor thing that confused me a bit, up to you

see also the requested changes in the other PR, I didn't check if they have interactions here at all, I don't think so?   maybe double-check

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
Niels-b: This makes it dialect independent. So the `supports_native_boolean` is not required here.
Comment is just for clarification/reasoning behind this.

soda-tests/src/helpers/test_connection.py
Niels-b: Nice catch indeed! Resolved 🙌

paulteehan (approved):

PR #2797, opened 2026-07-16
fix: zero-pad years below 1000 in SQL date and datetime literals

m1n0:
## Description

Failed-row-keys inserts into the Databricks diagnostics warehouse fail with `[INVALID_TYPED_LITERAL] The value of the typed literal "DATE" is invalid: '200-12-17'` when a source date has a year below 1000 (customer report: SCS-1320).

Root cause: `SqlDialect.literal_date` formatted dates with `strftime("%Y-%m-%d")`, and C `strftime("%Y")` does **not** zero-pad years < 1000 on glibc (where scan pods run) — it does pad on macOS, so the bug never reproduces on a dev Mac. `date.isoformat()` always emits a 4-digit year on every platform.

Fixed the same pattern at every `%Y` site that feeds SQL literals or wire payloads:
- base `SqlDialect.literal_date` (inherited by Databricks, Spark, Postgres, Snowflake, …)
- `SqlServerSqlDialect.literal_date` (same bug in its override)
- Trino `literal_datetime` / `literal_datetime_with_tz` (same bug for timestamps)
- `to_jsonnable` date serialization in `soda_cloud.py`

**Any expected impact on downstream packages/services?**

For years ≥ 1000 all outputs are byte-identical except Trino's tz-aware timestamp offset, which now renders `+02:00` (the format Trino documents) instead of `+0200`. Grepped soda-extensions: nothing pins the old formats.

**Reference issue:** SCS-1320

## Checklist

- [x] I verified this PR does not break [soda-extensions][1] (no extensions code pins the changed literal formats; only edge-case years < 1000 and the Trino offset rendering change).

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1516,7 +1516,8 @@ def to_jsonnable(o: Any, remove_null_values_in_dicts: bool = True) -> object:
     if isinstance(o, datetime):
         return convert_datetime_to_str(o)
     if isinstance(o, date):
-        return o.strftime("%Y-%m-%d")
+        # Not strftime("%Y-%m-%d"): C strftime does not zero-pad years < 1000 on glibc.
+        return o.isoformat()
     if isinstance(o, time):
         return o.strftime("%H:%M:%S")
     if isinstance(o, timedelta):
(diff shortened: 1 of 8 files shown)


Niels-b (approved):
Pre-approving, assuming all CI passes in extensions

m1n0:
@Niels-b https://github.com/sodadata/soda-extensions/actions/runs/29500308986

m1n0:
@Niels-b double checking before I merge (I cannot meaningfully verify that the changed code is right) - the extensions CI passed with replay off, so good to merge?

Niels-b:
@m1n0, if all of this runs with replay=off, then it should be okay to merge from my pov. The tests there should cover this change AFAIK.
Thanks for doing this 🙌

PR #2800, opened 2026-07-22
fix(bigquery): qualify CREATE SCHEMA with the project (SAS-13229)

Niels-b:
## Description

In a split BigQuery setup (separate compute/execution and storage projects), the diagnostics warehouse failed to write. `BigQuerySqlDialect` inherited the base `create_schema_if_not_exists_sql`, which qualifies with the dataset name only. BigQuery datasets are project-scoped, so `CREATE SCHEMA` created the dataset in the client's default (compute) project while the tables were addressed in the storage project — producing `404 Not found: Dataset <storage-project>:<dataset>` on every write.

**Fix**

Override `create_schema_if_not_exists_sql` in `BigQuerySqlDialect` to qualify the dataset with the project (`prefixes[0]`), consistent with the schema-existence check which already uses it.

```sql
-- before
CREATE SCHEMA IF NOT EXISTS `soda_diagnostics`;
-- after
CREATE SCHEMA IF NOT EXISTS `mm-storage-project`.`soda_diagnostics`;
```

Single-project setups are unchanged (the project equals the client's default, so the dataset lands in the same place as before).

**Expected impact on downstream packages/services**

- This is one half of the SAS-13229 fix. The companion change lives in soda-extensions (`build_dwh_prefixes` → storage project) and is required for the full fix. Verified end-to-end against a live split compute/storage BigQuery setup.
- Snapshots: this changes the emitted BigQuery `CREATE SCHEMA` SQL, so BigQuery snapshot replay falls back to live until the nightly re-record refreshes recordings. Fallback is enabled on PR/main replay, so CI stays green; only BigQuery recordings are affected.

**Reference:** SAS-13229

## Checklist

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -153,6 +153,15 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
     def information_schema_namespace_elements(self, data_source_namespace: BigQueryDataSourceNamespace) -> list[str]:
         return [data_source_namespace.project_id, data_source_namespace.dataset, "INFORMATION_SCHEMA"]
 
+    def create_schema_if_not_exists_sql(self, prefixes: list[str], add_semicolon: Optional[bool] = None) -> str:
+        # BigQuery datasets are project-scoped, so qualify the dataset with the project
+        # (prefixes[0]) — otherwise it lands in the client's default (compute) project
+        # instead of the storage project. The base implementation uses the dataset name alone.
+        add_semicolon = self.apply_default_add_semicolon(add_semicolon)
+        project_id, dataset = prefixes
+        qualified_schema_name: str = f"{self.quote_for_ddl(project_id)}.{self.quote_for_ddl(dataset)}"
+        return f"CREATE SCHEMA IF NOT EXISTS {qualified_schema_name}" + (";" if add_semicolon else "")
+
     def default_casify(self, identifier: str) -> str:
         return identifier.upper()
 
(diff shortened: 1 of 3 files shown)


soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -153,6 +153,15 @@ def _get_data_type_name_synonyms(self) -> list[list[str]]:
     def information_schema_namespace_elements(self, data_source_namespace: BigQueryDataSourceNamespace) -> list[str]:
         return [data_source_namespace.project_id, data_source_namespace.dataset, "INFORMATION_SCHEMA"]
 
+    def create_schema_if_not_exists_sql(self, prefixes: list[str], add_semicolon: Optional[bool] = None) -> str:
+        # BigQuery datasets are project-scoped, so qualify the dataset with the project
+        # (prefixes[0]) — otherwise it lands in the client's default (compute) project
+        # instead of the storage project. The base implementation uses the dataset name alone.
+        add_semicolon = self.apply_default_add_semicolon(add_semicolon)
+        project_id, dataset = prefixes
```
mivds, line 165: Do we need to validate the prefixes here? Fine as-is if that happens earlier already. Just thinking this may result in a hard to understand error if `prefixes` is invalid

mivds (approved):

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
Niels-b: Prefixes are usually generated by the BigQuery data source itself. So in that sense these should be okay already. 

But it won't hurt to have some extra validation there. I'll add it so we can more easily debug this.

PR #2802, opened 2026-07-27
feat(dialect): athena FLOAT casts + weekly BigQuery buckets, lenient athena timestamps, sample-stat nodes, redshift percentile capability

mivds:
## What
Dialect-level seams so soda-extensions observability code stops branching on dialect names (review direction from soda-extensions#516):

1. **Athena `_build_cast_sql` override** — canonical FLOAT casts render the Trino DML type `real` (Athena DDL is Hive where `float` is the float32 name; DML is Trino where it is `real`; the single canonical map serves DDL). The DDL map deliberately keeps `float`, pinned by a test.
2. **`STDDEV_SAMP` / `VAR_SAMP` AST nodes** — base renders the standard names; the T-SQL family (SqlServerSqlDialect + synapse/fabric subclasses) renders `STDEV`/`VAR`.
3. **BigQuery weekly time buckets** — `TIME_DELTA` renders weekly buckets day-denominated (`floor(days / 7·count)`): `TIMESTAMP_DIFF` only accepts MICROSECOND..DAY parts, and the previous `WEEK` rendering fails on the live engine (the test pinning it was authored from the renderer, not the engine — removed). `ADD_INTERVAL` already rendered weeks as `* 7 DAY`.
4. **Athena lenient timestamp converter** — pyathena's default converter requires fractional seconds and fails the whole fetch on `timestamp(0)` values (e.g. second-precision literals fed through date arithmetic); the connection now registers a converter accepting both forms.
5. **`supports_percentiles_with_other_distinct_aggregates()`** capability (base True, Redshift False) — Redshift refuses percentile aggregates combined with other DISTINCT aggregates in one SELECT; consumers batch them separately without naming the dialect.

## Consumers
soda-extensions#516 pins this branch (`SODA_CORE_BRANCH=athena-cast-real-not-float`) and removes all corresponding dialect-name branches from profiling/metric-monitoring; soda-extensions#515 reverts the FLOAT-cast workarounds once this merges.

## Tests
Unit pins in soda-athena / soda-bigquery / soda-sqlserver / soda-redshift dialect test files; live validation via soda-extensions#516's full data-source matrix against this branch.

### DIFF
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
@@ -18,6 +18,7 @@ from soda_core.common.metadata_types import SodaDataTypeName, SqlDataType
 from soda_core.common.sql_ast import (
     ADD_INTERVAL,
     ALTER_TABLE_ADD_COLUMN,
+    CAST,
     COLUMN,
     CREATE_TABLE,
     CREATE_TABLE_COLUMN,
@@ -354,6 +355,17 @@ class AthenaSqlDialect(SqlDialect, sqlglot_dialect="athena"):
         form; approx_percentile is its percentile aggregate."""
         expression_sql: str = self.build_expression_sql(percentile_within_group.expression)
         return f"approx_percentile({expression_sql}, {percentile_within_group.percentile})"
+
+    def _build_cast_sql(self, cast: CAST) -> str:
+        """Athena splits its SQL across two parsers: DDL is Hive (FLOAT is the
+        float32 type; REAL does not exist) while DML/expressions are Trino
+        (REAL is the float32 type; FLOAT does not exist). The canonical type
+        map serves DDL (see get_data_source_data_type_name_by_soda_data_type_names
+        and its create-table consumers), so canonical FLOAT casts must render
+        the Trino name here instead."""
+        if cast.to_type == SodaDataTypeName.FLOAT:
+            return f"CAST({self.build_expression_sql(cast.expression)} AS real)"
+        return super()._build_cast_sql(cast)
 
     def build_create_table_sql(
         self, create_table: CREATE_TABLE | CREATE_TABLE_IF_NOT_EXISTS, add_semicolon: bool = True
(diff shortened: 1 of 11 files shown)


Niels-b (approved):
**Review — verdict: Approve (with nits).** CI green; no blocking issues.

Nice set of seams — these let the observability code stop branching on dialect names, and they correctly supersede the earlier per-dialect workarounds (#2777 / #488 / #2787). Verified:

- **BigQuery weekly `TIME_DELTA`** renders day-denominated `FLOOR(TIMESTAMP_DIFF(…, DAY) / (7·count))` — no invalid `WEEK` part, and the math matches the base epoch-floor semantics + the `ADD_INTERVAL … * 7 DAY` counterpart (nested-floor identity checked).
- **`STDDEV_SAMP`/`VAR_SAMP` nodes**: base emits the standard names; the T-SQL family (SqlServer + inherited Synapse/Fabric) emits `STDEV`/`VAR` — the *sample* estimators, not `STDEVP`/`VARP`. New nodes are direct `SqlExpression` subclasses, so dispatch isn't shadowed.
- **Athena FLOAT cast** → Trino DML `real`; the DDL map keeps Hive `float`; the reverse map accepts both. Only the canonical FLOAT path is retargeted.
- **Redshift percentile capability** seam (base default True, Redshift False) and **lenient Athena timestamp parsing** (per-instance converter copy, fractional→plain fallback, `None`→`None`, tz variant untouched) both check out.

Non-blocking nits:
1. **Worth calling out:** the BigQuery weekly change also changes the *numeric bucketing*, not only the invalid `WEEK` part. The old `TIMESTAMP_DIFF(…, WEEK)` counted Sunday-based week *boundaries*, whereas the new form is start-anchored 7-day buckets (e.g. a 3-day span crossing a Sunday: old → 1, new → 0). The new form is the correct one (it matches the base epoch-floor + `ADD_INTERVAL` pairing), but since it shifts weekly-partition bucket numbers on BigQuery, worth noting explicitly in the PR description.
2. The Athena `real` cast narrows to 32-bit (float32) by design — consistent with the Hive `float` DDL column; just make sure downstream float aggregations tolerate float32.
3. Test hygiene: the retained `ADD_INTERVAL` weekly test is now a weaker duplicate of an existing exact-string pin and carries a stale comment that contradicts the corrected weekly understanding; and the raw-string `'float'` cast edge isn't pinned (behavior is fine via str-enum equality).

---
*AI-assisted review across correctness / SQL / tests lenses (read-only static analysis).*

PR #2803, opened 2026-07-27
fix(sqlserver): retry deadlock victims with jittered exponential backoff

mivds:
## Problem

SQL Server CI jobs fail intermittently with:

```
pyodbc.Error: ('40001', '... Transaction (Process ID N) was deadlocked on lock resources
with another process and has been chosen as the deadlock victim. Rerun the transaction. (1205) ...')
```

Root cause: under pytest-xdist all workers share one database. Per-worker schemas isolate the *user tables*, but not the **system catalog** — the test harness's frequent metadata scans (`information_schema.tables` lookups in `ensure_test_table` and schema teardown) take shared locks across catalog rows of *all* schemas, while other workers run `CREATE TABLE` / `DROP TABLE` / `DROP SCHEMA` DDL holding exclusive locks on the same system base tables. SQL Server resolves the resulting deadlock by killing the (cheaper) reader.

## Fix

`SqlServerDataSourceConnection.execute_query` / `execute_update` now retry statements killed as deadlock victims (SQLSTATE `40001`), which is Microsoft's documented remedy for error 1205:

- up to 3 attempts, then the error propagates
- `rollback()` between attempts to clear the aborted transaction
- full-jitter exponential backoff (`uniform(0, 0.1s)`, then `uniform(0, 0.2s)`) so both parties of a killed deadlock don't retry in lockstep
- all other `pyodbc` errors propagate unchanged

Fabric and Synapse connections subclass this connection and inherit the behavior.

## Testing

- 4 new unit tests in `soda-sqlserver/tests/unit/test_sqlserver_deadlock_retry.py` (retry-then-succeed, non-deadlock errors not retried, re-raise after persistent deadlocks, `execute_update` path, backoff shape) — written test-first
- Full `soda-sqlserver`, `soda-fabric`, `soda-synapse` suites pass (63 tests)

### DIFF
--- a/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
+++ b/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
@@ -1,15 +1,18 @@
 from __future__ import annotations
 
 import logging
+import random
 import struct
+import time
 from abc import ABC
 from datetime import datetime, timedelta, timezone, tzinfo
-from typing import Any, Literal, Optional, Union
+from typing import Any, Callable, Literal, Optional, Union
 
 import pyodbc
 from pydantic import Field, SecretStr
 from soda_core.__version__ import SODA_CORE_VERSION
 from soda_core.common.data_source_connection import DataSourceConnection
+from soda_core.common.data_source_results import QueryResult
 from soda_core.common.exceptions import DataSourceConnectionException
 from soda_core.common.logging_constants import soda_logger
 from soda_core.model.data_source.data_source import DataSourceBase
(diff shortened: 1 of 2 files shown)


Niels-b (approved):
**Approve with nits** ✅

Nice, focused change — the retry logic itself is correct (no off-by-one, correct re-raise semantics, safe SQLSTATE detection) and the inheritance to Fabric/Synapse checks out. A few things worth addressing; none block merge.

### Correctness

**[MAJOR — correctness, low real-world risk] Rollback is transaction-scoped, not statement-scoped, and SQL Server proper runs `autocommit=False`.**
`_get_autocommit_setting()` returns `False` for SQL Server proper, and `execute_query` never commits — so SELECTs accumulate in one open transaction. On a deadlock the wrapper issues a blanket `self.connection.rollback()` and then retries only the failed statement, discarding any prior uncommitted work on that connection. Real-world risk is low (writes go through `execute_update`, which commits per call; the harness statements are effectively single-unit; Fabric/Synapse are `autocommit=True` and unaffected), but the base class this lives in isn't uniformly safe for a multi-statement transaction. Worth a one-line docstring caveat that the wrapper assumes single-statement units, or explicitly scoping it.

**[MINOR→MAJOR depending on intent] Coverage is uneven across execution paths.**
Only `execute_query` + `execute_update` are wrapped. That fully covers the harness deadlock path (metadata scans → `execute_query`; CREATE/DROP TABLE/SCHEMA → `execute_update`). But `execute_query_one_by_one`, `execute_query_one_by_one_prefer_streaming`, and `execute_query_iterate` are not wrapped — so a production streaming read (e.g. the diagnostics-warehouse failed-rows flow) gets no retry while a buffered one does. Fine if the scope is intentionally "harness only" (a docstring note would help); inconsistent if the goal is general production deadlock resilience.

### Style / API

**[MAJOR — style only] Method ordering.**
The new `execute_query` / `execute_update` / helpers are placed before `__init__`, which diverges from the sibling `PostgresDataSourceConnection` (and the base class), both of which lead with `__init__`. Suggest moving them below.

**[MINOR] Knobs are class attributes, not connection properties.**
`DEADLOCK_MAX_ATTEMPTS` / `DEADLOCK_RETRY_BACKOFF_SECONDS` can't be tuned without subclassing, despite the existing `connection_max_retries` connection property being the natural sibling. If this is meant to be operator-tunable, consider exposing them the same way.

### Tests

**[MAJOR — test gap] The rollback-failure branch is untested.**
The `except Exception as rollback_error` path (rollback itself throwing during recovery) has no coverage — and it's arguably the riskiest branch, since it must not mask the original deadlock error. `FakeConnection` already centralizes rollback, so a variant that raises is cheap. Highest-value addition.

**[MAJOR — test rigor] The backoff test doesn't actually validate the jitter.**
`test_deadlock_backoff_is_exponential_with_jitter` stubs `random.uniform` to return `high`, so `sleeps == [base, base*2]` only proves the cap schedule is exponential — it never exercises full-jitter's defining property (`0 <= sleep <= cap`). Consider an additional test that leaves `uniform` un-stubbed (seed `random`) and asserts each recorded sleep falls within `[0, cap]`. That's the test that would catch someone dropping the jitter.

**[MINOR — test gaps] Missing a mixed `[deadlock, other_error]` sequence.**
i.e. the first attempt deadlocks and retries, the second hits a different error that must propagate unchanged. Realistic and currently uncovered. (Also nice-to-have: a `caplog` assertion that the retry warning is emitted.)

**[MINOR — test fidelity] Deadlock fixture uses base `pyodbc.Error`; real pyodbc raises `OperationalError`.**
The tests pass as written because the predicate catches `pyodbc.Error` (the base) and keys off `e.args[0] == "40001"`, so this is purely fidelity — using `pyodbc.OperationalError` in `deadlock_error()` would mirror what the driver actually raises for 1205.

---

Thanks for the test-first approach and the clear root-cause writeup in the description 👍

PR #2805, opened 2026-07-28
fix(logging): don't mutate frame.f_locals while iterating in _mask_exception

Niels-b:
## Problem

After #2803 merged, the `(3.10, sqlserver)` CI lane started failing on `soda-groupby`'s `test_groupby_check_single_group_cloud_please_crash`:

```
AssertionError: Expected contract verification failed
```

The test deliberately injects an invalid column (`please_crash`) to force a query error and asserts the contract reports a failure. Instead the check came back `NOT_EVALUATED` with `hasErrors: false`.

## Root cause

`_mask_exception` (`logging_configuration.py`) masks secrets in a logged exception by walking its traceback frames and reassigning string locals **while iterating the same mapping**:

```python
for var_name, var_value in frame.f_locals.items():
    if isinstance(var_value, str):
        frame.f_locals[var_name] = _mask_message(var_value)   # re-accesses frame.f_locals
```

Each `frame.f_locals` access re-syncs the frame's fast locals into that dict. Mutating it mid-iteration raises `RuntimeError: dictionary keys changed during iteration` on **Python 3.10** for certain frame shapes — specifically a bound-method local plus a closure frame.

That is exactly the traceback shape #2803 introduced on the SQL Server path:

```python
def execute_query(self, sql, log_query=True):
    execute = super().execute_query                       # bound-method local
    return self._execute_with_deadlock_retry(lambda: execute(sql, log_query))  # closure frame
```

So the pre-existing latent bug in `_mask_exception` became reachable. And because it raised **inside the logging handler** (`logger.error(..., exc_info=True)`), Python dropped the log record — silently swallowing the query error, which is why the check ended up `NOT_EVALUATED` rather than failed.

Only 3.10 is affected for this shape (3.11/3.12/3.13 don't crash here), matching the single failing CI lane.

## Fix

- Iterate over a `list(...)` **snapshot** of the frame locals, so the in-loop reassignment can't invalidate the iterator.
- Guard the per-frame masking with `try/except` — masking is best-effort and cosmetic, so it must never break logging (that swallowing is exactly what turned a real error into a green-but-wrong check).

## Testing

New unit tests in `soda-tests/tests/unit/test_logging_masking.py`:
- **regression**: masking an exception raised through the retry-wrapper traceback shape must not raise;
- masking still redacts registered secrets in the exception args;
- `None` handling.

Verified against the real fixed module on CPython **3.10 / 3.11 / 3.12 / 3.13**: the pre-fix code raises `dictionary keys changed during iteration` on 3.10; the fixed code passes all three tests on every version.

### DIFF
--- a/soda-core/src/soda_core/common/logging_configuration.py
+++ b/soda-core/src/soda_core/common/logging_configuration.py
@@ -136,12 +136,21 @@ def _mask_exception(e: Optional[BaseException]) -> Optional[BaseException]:
     tb = e.__traceback__
     while tb is not None:
         frame = tb.tb_frame
-        # Mask local variables in the frame that are strings
+        # Mask local variables in the frame that are strings.
+        # Iterate over a snapshot (``list(...)``): assigning ``frame.f_locals[...]``
+        # re-syncs the frame's fast locals, mutating the very mapping being iterated,
+        # which raises "dictionary keys changed during iteration" on Python < 3.13 for
+        # some frames (e.g. the closure frames added by the SQL Server deadlock-retry
+        # wrapper). Masking is best-effort and cosmetic, so it must never break logging:
+        # a failure here previously propagated out of the logging handler and silently
+        # dropped the record, swallowing the very error being logged.
         if frame.f_locals:
-            for var_name, var_value in frame.f_locals.items():
-                if isinstance(var_value, str):
-                    masked_value = _mask_message(var_value)
-                    frame.f_locals[var_name] = masked_value
+            try:
+                for var_name, var_value in list(frame.f_locals.items()):
+                    if isinstance(var_value, str):
+                        frame.f_locals[var_name] = _mask_message(var_value)
+            except Exception:
+                pass
         tb = tb.tb_next
 
     # Recursively mask chained exceptions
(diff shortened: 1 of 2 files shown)


soda-core/src/soda_core/common/logging_configuration.py
```diff
@@ -136,12 +136,21 @@ def _mask_exception(e: Optional[BaseException]) -> Optional[BaseException]:
...
-            for var_name, var_value in frame.f_locals.items():
-                if isinstance(var_value, str):
-                    masked_value = _mask_message(var_value)
-                    frame.f_locals[var_name] = masked_value
+            try:
+                for var_name, var_value in list(frame.f_locals.items()):
+                    if isinstance(var_value, str):
+                        frame.f_locals[var_name] = _mask_message(var_value)
+            except Exception:
+                pass
```
mivds: The blanket pass is an issue here. If we can't mask, we should prevent logging. The current approach will happily log secrets

The comment above mentions masking is best-effort. That's wrong. Masking is critical for security
> Masking is best-effort and cosmetic, so it must never break logging

mivds (changes requested):

soda-core/src/soda_core/common/logging_configuration.py
m1n0: Agreed. Pushed 83c418e1: a masking failure now raises MaskingError and _mask_record drops the exception and traceback from the record entirely, appending [exception redacted: secret masking failed] to the (already masked) message. Kept the record itself at ERROR rather than dropping it — dropping the record is what turned the failed check into NOT_EVALUATED in the first place. Tests cover the raise, the redaction and the rendered output.

mivds (approved):

PR #2809, opened 2026-07-28
fix(cloud): rename hasWarns to hasWarnings in scan results payload

mivds:
## Problem

Every v4 `sodaCoreInsertScanResults` payload emits the warning flag as `hasWarns`, but the backend command DTO only maps `hasWarnings` and silently drops unknown keys (GSON). The scan-state derivation (`getScanState`: hasErrors > hasFailures > hasWarnings) therefore never sees warnings from v4 clients — a scan whose checks warned reports **SUCCESS** instead of **COMPLETED_WITH_WARNINGS** in Soda Cloud. This affects all v4 flows that ship check results (contract verification, metric monitoring).

Found during the v3↔v4 parity crosscheck against the backend on the soda-extensions parity gate; v3 (soda-library) sends `hasWarnings`, and soda-core's own payload model (`soda_cloud_dto.py`) already declares `hasWarnings` — only the emitter in `_build_check_collection_results_json_dict` drifted.

## Change

Rename the emitted key to `hasWarnings`. No compatibility shim needed: the old spelling was never read by the backend, so nothing can regress by dropping it.

A companion PR on soda-extensions updates the MM parity contract, which currently pins the drifted spelling.

## Testing

`test_combined_results_wire.py` + `test_single_result_wire_equivalence.py` updated and passing (21/21); soda-extensions MM parity suite run against this branch (companion PR).

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1869,9 +1869,6 @@ def _build_check_collection_results_json_dict(
         "scanStartTimestamp": min(started_timestamps) if started_timestamps else head.started_timestamp,
         "scanEndTimestamp": max(ended_timestamps) if ended_timestamps else head.ended_timestamp,
         "hasErrors": any(r.has_errors for r in results),
-        # "hasWarnings" is the field the backend maps for scan-state
-        # derivation (COMPLETED_WITH_WARNINGS); unknown keys are silently
-        # dropped, so any other spelling loses the warning severity.
         "hasWarnings": any(r.is_warned for r in results),
         "hasFailures": any(r.is_failed for r in results),
         # Empty checks list serialises as ``None`` to match the legacy
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1792,7 +1792,7 @@ def _build_check_collection_results_json_dict(
 
     - ``scanStartTimestamp`` = min of per-result starts
     - ``scanEndTimestamp`` = max of per-result ends
-    - ``hasErrors``/``hasWarns``/``hasFailures`` = ORs over per-result flags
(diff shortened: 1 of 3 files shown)


Niels-b (approved):
Approving. This is a genuine wire-format bug fix, not a breaking rename. Verified across repos: the Cloud backend command DTO, converter, and parser round-trip test all read `hasWarnings` (zero occurrences of `hasWarns` in the backend), the published API spec uses `hasWarnings`, and v3 (soda-library) already sends `hasWarnings`. Because the backend deserializes with GSON (which silently ignores unknown keys), the v4 engine's `hasWarns` was never read — so warning-only scans were reported as SUCCESS instead of COMPLETED_WITH_WARNINGS. This also aligns the emitter with soda-core's own outbound DTO, which already declared `hasWarnings`. Complete within the repo (no stray `hasWarns` left), and the trio is now consistently spelled (`hasErrors`/`hasWarnings`/`hasFailures`).

Two small non-blocking notes:

- The updated wire tests assert the new key's value but not that the old key is gone. Since the payload is a dict literal, adding `assert "hasWarns" not in payload` alongside each assertion would catch an accidental reintroduction.
- Coordinate the merge with the soda-extensions v3↔v4 parity companion PR — the parity harness (`compare.py`) currently allow-lists the old `hasWarns` spelling, so landing this without that change would break the parity gate.

m1n0 (approved):
Checked: zero `hasWarns` left anywhere at this head; the backend only maps `hasWarnings` (SodaCoreInsertScanResultsCommand.java:82, no aliasing anywhere) and derives COMPLETED_WITH_WARNINGS from it; v3 always sent `hasWarnings`. So "no shim needed" holds.

Merge-order caveat: the companion soda-extensions#523 is a hard flip of the parity pin (compare.py:125/:137), and the parity gate runs on release tags against core git main — this and #523 both need to land before the next extensions release tag; either one alone goes red exactly at tag time.

PR #2811, opened 2026-07-31
feat(data-source): add primary-key metadata introspection

mivds:
## Problem

The v4 engine exposes column schema (`get_columns_metadata`) but has no way to read a table's primary keys.

## Change

Adds a neutral bulk `DataSourceImpl.get_primary_keys(dataset_prefixes, dataset_names) -> dict[str, list[str]]` primitive and an `is_primary_key` flag on `ColumnMetadata`. One query returns every requested table's primary key columns, keyed by table name, each list ordered by the column's position within the key so composite keys keep their declared order. Introspection is implemented for Postgres via a dedicated `MetadataPrimaryKeysQuery` — the same per-data-source Query-object pattern as `MetadataTablesQuery`, so query logic stays out of `SqlDialect` and non-standard sources extend it by subclassing. Postgres reads `pg_catalog` (`pg_constraint`/`pg_attribute`) rather than `information_schema`, which is privilege-filtered and silently returns zero primary keys for a read-only monitoring user — the same reason `PostgresMetadataTablesQuery` bypasses it. Other data sources inherit a safe empty-dict default and opt in per data source. The addition is additive and backward-compatible.

## Testing

Unit tests for `MetadataPrimaryKeysQuery` (query construction incl. the cross-schema JOIN guard, result parsing/ordering) and the Postgres `pg_catalog` query shape. Integration tests cover single + composite keys and a cross-schema constraint-name collision. Verified manually that a read-only role gets zero PKs via `information_schema` and the full ordered key via `pg_catalog`. Green.


Niels-b (approved):
LGTM, Only relevant thing Claude pointed out is:

get_primary_keys returns dict[str, set[str]] with no ORDER BY ordinal_position — composite-PK column order is discarded, while the write path preserves order. Not a functional bug for capture-schema (which only needs a per-column boolean), but this is a frozen contract the whole chain inherits; changing it later is expensive. Decide now: is an ordered list the right return type?

and

PK-DDL logic lives only in the base build_create_table_sql, but Synapse/Athena/Oracle override that method and silently skip the new PK clause. Fine today (they declare no PKs) but a broken extension point the moment #533 wants Oracle/DB2 PK fixtures.

m1n0 (changes requested):
The postgres implementation reads `information_schema.table_constraints`/`key_column_usage`, which postgres filters to tables the current user owns or has a non-SELECT privilege on — a read-only monitoring user (the common production grant) gets zero PKs with no error. The sibling metadata queries already bypass information_schema for pg_catalog for exactly this reason (postgres_metadata_tables_query.py:35). CI connects as the table owner, so the integration tests can't catch it. Needs a pg_catalog override (pg_index/pg_constraint), same pattern as PostgresMetadataTablesQuery.

- Schema/catalog filters compare exact case via metadata_casify (metadata_primary_keys_query.py:169-182), where MetadataTablesQuery does `LOWER(col) = lower(literal)` — a second silent-empty path when the contract spells the schema in a different case.
- `dict[str, set[str]]` can't represent composite-PK column order, and switching to a list later is a breaking change once extensions has subclassed. Either return ordered lists now or document order as out of scope in the docstring.
- No unit test asserts the built SQL — the join the "do not simplify" comment protects (metadata_primary_keys_query.py:124) is only covered by the postgres integration test; a unit regression dropping a join condition would pass.
- PR description says `get_primary_keys(prefixes, table) -> set[str]`; the actual API is bulk: `(dataset_prefixes, dataset_names) -> dict[str, set[str]]`. Worth fixing before extensions codes against the description.
- data_source_impl.py:270 docstring should say result keys carry the info-schema casing (can differ from the requested names) and that no-PK/missing tables are simply absent — the integration test already compensates with a case-insensitive lookup.

The cross-schema constraint-name collision regression test is a nice touch.

mivds:
Refactored to use a `list` instead of a `set` for primary keys. I originally went with `set`, because we drop the order anyway when communicating to Cloud. Points about it being a breaking change if we later change it are valid, so I refactored to preserve order here.

I've kept the PK-DDL logic as-is. Proposed solution is to add more hooks so we can have more specialized overrides for athena/synapse/oracle, and the primary key flow is then inherited. My argument here is that we can't continue adding hooks for everything. If a dialect does something different, override it, simple as that. The codebase is becoming increasingly difficult to reason about because everything becomes increasingly fragmented with hooks.

@m1n0 Regarding postgres: I propose keeping it as-is. The reason we query `pg_catalog` and others is to detect special table types like materialized views, which are not exposed in `information_schema`. Postgres doesn't support primary keys for those special table types, so `information_schema` should be sufficient. The access concern also doesn't make sense to me. If the user doesn't have `SELECT` permission on the table, we can't run anything on that table anyway.

m1n0:
Fair point on the matviews — the tables query is on pg_catalog for relkinds, my attribution was off.

The access concern is the reverse case though: not a user without SELECT, a user with *only* SELECT. `information_schema.table_constraints` only shows constraints on tables you own or hold some privilege *other than* SELECT on — its WHERE clause checks `has_table_privilege(oid, 'INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER')`, SELECT deliberately excluded. Ran it on postgres:16:

```sql
CREATE TABLE orders (tenant_id int, id int, PRIMARY KEY (tenant_id, id));
CREATE ROLE soda_ro LOGIN; GRANT USAGE ON SCHEMA public TO soda_ro;
GRANT SELECT ON orders TO soda_ro;
```

As `soda_ro`: `SELECT count(*) FROM orders` works fine, but this PR's `table_constraints ⋈ key_column_usage` join returns 0 rows. Same join as owner: 2 rows. A `pg_index`/`pg_class` version returns both PK columns for `soda_ro`.

So a read-only monitoring user — the common production grant — can run every scan and still silently gets "no PKs", and CI can't catch it because it connects as the owner. v3 ran the same join, so it's parity rather than a regression, but PK capture is the point of this feature. I'd do a pg_index-based override on the postgres data source (about the same size as the current query). If you'd rather ship as-is, then at minimum a warning when tables were found but the constraint query came back empty, plus a known-limitation note — same caveat likely applies to Redshift, which #2813 opts into this same query.

mivds:
Addressed in 84147ffd + 0f38ce4f:

- **pg_catalog override**: `PostgresMetadataPrimaryKeysQuery` now reads `pg_constraint`/`pg_class`/`pg_namespace`/`pg_attribute` (same pattern as `PostgresMetadataTablesQuery`), with `array_position(conkey, attnum)` as the ordinal. Verified against a live read-only role: `information_schema` returns zero rows, `pg_catalog` returns the full composite key in order.
- **Schema case**: the postgres query filters `LOWER(nspname) = lower(schema)`, mirroring the tables query.
- **Ordered list**: `get_primary_keys` now returns `dict[str, list[str]]` ordered by key position (base: `ordinal_position`; Snowflake `key_sequence`, Oracle `POSITION`, DB2 `COLSEQ` in the stacked PRs).
- **SQL-shape unit tests**: added — one guards the cross-schema JOIN conditions (a dropped condition now fails in unit, not just live postgres), one asserts the postgres query reads pg_catalog and not information_schema.
- **PR description + docstring**: updated to the bulk `dict[str, list[str]]` API; docstring documents result-key casing and absent-table semantics.

One heads-up: Redshift (#2813) delegates to the base information_schema query — being postgres-derived it may share the privilege filtering. Left as-is there since its metadata queries already use information_schema; can follow up if you think it warrants the same treatment.

Niels-b:
> Same caveat likely applies to Redshift, which https://github.com/sodadata/soda-core/pull/2813 opts into this same query.

Yes, we indeed had something similar pop up with Redshift recently. Also worth taking a look at in #2813 .

mivds:
Follow-up on the Redshift heads-up: done in #2813 (2b775040). Redshift now reads `pg_catalog.pg_constraint` too — its information_schema constraint views share the Postgres-8 owner-only filter (same lineage issue as `table_schemata`/SCS-1193). Redshift has no `array_position()`, so the key columns are parsed from `pg_get_constraintdef(oid)` (the function AWS's `v_generate_tbl_ddl` uses), which renders them in declared key order.

Niels-b (approved):

m1n0 (approved):

PR #2812, opened 2026-07-31
feat(core): default-off engine hooks for cross-source reconciliation (R-635)

Niels-b:
## Description

Additive, **default-off** engine hooks in `soda-core` that let a data source plug into
cross-source (e.g. Salesforce ↔ warehouse) reconciliation without changing behaviour for any
existing data source. Part of the **R-635 Salesforce v4** workstream (vertical slice across
`soda-core`, `soda-extensions`, `soda`, `soda-contracts-launcher`).

- What problem are you solving? The Salesforce connector needs to (a) declare a cross-type value
  comparator, (b) reuse the engine's filtered-CTE name, and (c) opt into recon-source capabilities
  (transient-read retries, text-key normalization, row sampling). These hooks live in the base
  `DataSourceImpl` / dialect with conventional defaults.
- Changes:
  - `ValueComparatorProtocol` (`common/data_source_impl.py`) — a `@runtime_checkable` **type-only**
    contract; `get_value_comparator()` typed `Optional[ValueComparatorProtocol]` (was `Optional[Any]`).
  - `SODA_FILTERED_CTE_NAME` constant (`common/sql_ast.py`) — the existing filtered-CTE literal
    hoisted to one place (byte-identical value), used by `check_collections/base.py`.
  - Default-off recon-source capability hooks (transient reads, value comparator, row sampling,
    text-key normalization) on the base data source / dialect, incl. the `select_all_paginated_sql`
    key-normalization render for the base dialect, SQL Server (OFFSET/FETCH), and Synapse (the
    hand-rolled ROW_NUMBER fold).
- Expected impact on downstream: `soda-extensions` (`soda-salesforce` imports the protocol + CTE
  name; `soda-reconciliation` consumes the hooks). **Independently validated** that all six new hooks
  default to behaviour-preserving values and are overridden **only** by `soda-salesforce` — for a
  plain warehouse↔warehouse recon the happy path of every check is byte-for-byte unchanged.
- Reference: R-635.

## Merge order

Land/publish **`soda-core` first**, then `soda-extensions` (its `soda-salesforce` imports these
symbols and hard-fails at import without them; `soda-reconciliation` at check runtime), then the
launcher.

## Checklist

- [x] I added a test to verify the new functionality. (Recon capability-hook + text-key tests; new
  offline **paginated normalize-render** seam tests for the base `SqlDialect`, SQL Server, and the
  Synapse ROW_NUMBER fold; plus the `ValueComparatorProtocol` conformance + CTE-drift tests in
  `soda-salesforce` so a soda-core contract change trips loudly.)
- [x] I verified this PR does not break [soda-extensions][1]. (Salesforce offline suite green: **243
  passed**; the dedicated `test-salesforce` CI job editable-installs this branch.)

### DIFF
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -2,7 +2,16 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Protocol, Type, runtime_checkable
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Protocol,
+    Type,
+    runtime_checkable,
+)
 
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_results import QueryResult, QueryResultIterator
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Callable, Dict, Optional, Type
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Type
 
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_results import QueryResult, QueryResultIterator
@@ -216,17 +216,43 @@ class DataSourceImpl(ABC):
         """
         return True
 
+    @property
+    def reads_may_fail_transiently(self) -> bool:
+        """Whether a runtime error while streaming reads from this source should be treated as a
+        soft (recoverable) failure by a consumer rather than propagate.
+
+        Default False: conventional SQL sources keep their existing behavior — a read error
+        propagates (a genuine DB failure is surfaced/raised, unchanged). A remote/API source whose
+        streamed reads can fail transiently mid-scan (e.g. Salesforce REST queryMore / token
+        expiry) returns True, so a consumer can degrade gracefully instead of crashing on such a
+        transient failure.
+        """
+        return False
+
+    def get_value_comparator(self) -> Optional[Any]:
+        """A value comparator to use when comparing this source's values against another source's,
+        or ``None`` to let the caller use its own default.
+
+        Default ``None``: conventional SQL sources have no special needs — the caller compares
+        exactly (existing behavior, unchanged). A source that normalizes values to canonical Python
+        types which may not exactly-equal a peer driver's native types (e.g. Salesforce returns
+        numbers as ``Decimal`` and datetimes as tz-aware UTC) overrides this to return its own
+        comparator object (exposing ``equals(x, y) -> bool`` and a ``handles_cross_type: bool``),
+        which the caller then uses whenever this source participates.
+        """
+        return None
+
     def validate_orderable_key_columns(
         self, dataset_prefixes: list[str], dataset_name: str, key_columns: list[str]
     ) -> list[str]:
-        """Return human-readable problems if any key column cannot be used as an ordered/paged
-        reconciliation key on this data source; empty list means all keys are usable.
+        """Return human-readable problems if any of the given columns cannot be used as an
+        ordered/paged read key on this data source; empty list means all are usable.
 
-        A paginated reconciliation read (rows_diff/reference_diff) ORDER BYs the key columns.
-        Most SQL sources can order by any column, so the default returns no problems and existing
-        behavior is unchanged. A source whose query language forbids ordering by some field types
-        (e.g. Salesforce/SOQL cannot ORDER BY a long-text field) overrides this to fail early —
-        at check setup, before any cursor opens — instead of erroring mid-stream.
+        A paginated ordered read ORDER BYs these columns. Most SQL sources can order by any column,
+        so the default returns no problems and existing behavior is unchanged. A source whose query
+        language forbids ordering by some field types (e.g. Salesforce/SOQL cannot ORDER BY a
+        long-text field) overrides this to fail early — before any cursor opens — instead of
+        erroring mid-stream.
         """
         return []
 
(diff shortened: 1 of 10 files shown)


soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -30,6 +39,22 @@
     from soda_core.contracts.impl.contract_verification_impl import ContractImpl
 
 
+@runtime_checkable
```
mivds: Do we need `runtime_checkable`? Not sure if we check it anywhere, but from what I remember it isn't that great:

1. It only checks that it has matching names; function signatures are not validated. Meaning `def equals(self):` would also satisfy the protocol but fail at runtime
2. `isinstance` checking is slow for protocols

I would say 1 is main argument against it, as it may give a false sense of security

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -30,6 +39,22 @@
...
+class ValueComparatorProtocol(Protocol):
+    """Cross-source value-equality contract a DataSourceImpl may supply via get_value_comparator().
+
+    When this source participates in a cross-source comparison, a comparator matches mixed
+    representations of the same underlying value (e.g. Decimal vs float, tz-aware vs naive datetime).
+    ``handles_cross_type`` advertises whether it does such cross-type matching. Structural — any object
+    exposing these two members conforms; the concrete comparators live in the extensions that need
+    them (e.g. soda-salesforce, soda-reconciliation)."""
+
+    handles_cross_type: bool
```
mivds: Assuming this is a read-only field, I'd make it a property. An implementation that uses a regular attribute still complies, it just makes it explicit the field is read-only

```suggestion
    @property
    def handles_cross_type(self) -> bool:
        pass
```

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -450,13 +451,34 @@ def select_all_paginated_sql(
...
+
+    def order_by_key_expression(self, column_expression_sql: str) -> str:
+        """Case-fold a key column for case-insensitive ORDER BY.
+
+        Invoked only when the caller activates text-key normalization (default-off), so existing
+        ORDER BY rendering is untouched. Standard SQL case-folds with LOWER(); a dialect that
+        already orders text case-insensitively (e.g. Salesforce/SOQL) overrides this to identity —
+        and in practice such a side is never asked to normalize.
+        """
+        return f"LOWER({column_expression_sql})"
```
mivds: Why not use the AST `LOWER` here? That would increase compatibility and remove the need for wrapping with `SqlExpressionStr`

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -440,6 +440,7 @@ def select_all_paginated_sql(
         order_by: list[str],
         limit: int,
         offset: int,
+        normalize_key_columns: frozenset[str] = frozenset(),
```
mivds, line 444: Are there cases where we only want to normalize a subset of the columns? I would think it's always normalize everything or nothing. That would allow to simplify the logic a little bit

mivds (approved):

m1n0 (commented):
Default path checks out — byte-identical SQL with `normalize_key_columns` empty, the `base.py` change is a real no-op, and Synapse's paginator finally has tests. Three things:

- `sql_dialect.py:461-475`: with normalization active, the folded term is the only sort key, so `"ABC"` and `"abc"` tie — and since each page re-executes the query, tied rows can swap across page boundaries: silently skipped/duplicated rows in the recon stream. Cheap fix: raw column as tiebreaker, `ORDER BY LOWER("code"), "code"`.
- The fold isn't projected into the SELECT list, which breaks under DISTINCT on pg/oracle/tsql. Not reachable today, but my #2816 adds a `distinct` param to this same function and we conflict on these lines anyway — I'll reconcile when merging second, flagging so the resolution is deliberate rather than whoever-git-picks.
- `test_sql_dialect_seams.py:99` only asserts `'LOWER'` is absent from the default-path SQL — a golden-string assertion on one dialect would actually pin the equivalence the docstring promises.

Nits: `supports_row_sampling` is a `@property` while every sibling `supports_*` is a plain method — easy for a subclass to override wrong; and the six new base-class members have no consumers in core itself, a short docstring pointer to the consuming extension would help OSS readers.

soda-core/src/soda_core/common/sql_dialect.py
mivds: Never mind, realized we need it to deal with non-text columns

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: Thanks for flagging, I will remove it.

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Nice catch, thank you for pointing it out!

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: Done — made `handles_cross_type` a `@property`.

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Yep — keeping the per-column set exactly for that: it lets us fold only the flagged text keys and leave numeric/other key columns untouched. Left as-is.

Niels-b:
Thanks for the detailed pass @m1n0 — addressed the first three:

- **Tiebreaker:** with normalization active the folded term is no longer the only sort key — I append the raw column as a deterministic tiebreaker (`ORDER BY LOWER("code"), "code"`) across the base, SQL Server and Synapse paginators, so the paging order is total and tied case-collisions can't skip/duplicate across pages. The default-off path stays byte-identical.
- **Golden test:** the seams test now pins the full paginated SQL string (both default-off and normalized-with-tiebreaker) instead of just asserting `LOWER` is absent.
- **`supports_row_sampling`:** now a plain method like its `supports_*` siblings, with callers updated.

On the **SELECT-list projection under DISTINCT**: leaving it as-is on purpose — as you said it's not reachable today, and your #2816 adds a `distinct` param that conflicts on these exact lines, so I'd rather you reconcile it deliberately when merging #2816 second than have git pick for us.

PR #2813, opened 2026-07-31
feat(data-source): bulk primary-key introspection for more data sources

mivds:
## Problem

The v4 primary-key primitive (#2811) implemented `get_primary_keys` only for Postgres; every other data source returned an empty result.

## Change

Enables bulk primary-key introspection across the OSS data sources — Snowflake, Redshift, SQL Server, Databricks, BigQuery, DuckDB, Synapse — each opting in via `supports_primary_keys()` + a bulk `get_primary_keys` (delegate to the standard `information_schema` query, or a `MetadataPrimaryKeysQuery` subclass for BigQuery's dataset-qualified namespace and DuckDB's `constraint_column_usage`). BigQuery and Synapse add a create-table `PRIMARY KEY (...) NOT ENFORCED` DDL override so the integration test can create a keyed table. Trino, SparkDF, Fabric, and Athena stay **opt-out** — no `information_schema` constraint views and/or no inline PRIMARY KEY DDL (Fabric is ALTER-only).

Stacked on #2811. Part of OBSL-1043.

## Verification

`test_primary_keys_metadata` (from #2811) runs per data source in the CI matrix, gated on `supports_primary_keys()`. **Locally verified: Postgres and DuckDB** (embedded — DuckDB exercises the subclass path end-to-end). Container-backed sources (Snowflake, BigQuery, Redshift, SQL Server, Databricks, Synapse) are verified by their CI jobs; each assumes its `information_schema` / declared-PK metadata is populated and role-visible.

### DIFF
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
@@ -193,6 +193,10 @@ class AthenaDataSourceImpl(DataSourceImpl, model_class=AthenaDataSourceModel):
 class AthenaSqlDialect(SqlDialect, sqlglot_dialect="athena"):
     SUPPORTS_DROP_TABLE_CASCADE = False
 
+    # Primary-key introspection stays opt-out (supports_primary_keys inherits False from the
+    # base dialect): Athena runs over Glue/Hive-style catalogs that neither declare primary keys
+    # in CREATE TABLE nor expose them through information_schema constraint views.
+
     SODA_DATA_TYPE_SYNONYMS = (
         (SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR),
         (SodaDataTypeName.NUMERIC, SodaDataTypeName.DECIMAL),
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -110,13 +110,10 @@ class BigQueryDataSourceImpl(DataSourceImpl, model_class=BigQueryDataSourceModel
         return BigQueryDataSourceNamespace(project_id=prefixes[0], dataset=prefixes[1])
 
     def create_metadata_primary_keys_query(self) -> MetadataPrimaryKeysQuery:
+        # BigQuery exposes primary keys through the dataset-qualified INFORMATION_SCHEMA constraint views.
         return BigQueryMetadataPrimaryKeysQuery(
             sql_dialect=self.sql_dialect, data_source_connection=self.data_source_connection
         )
-
-    def get_primary_keys(self, dataset_prefixes: list[str], dataset_names: list[str]) -> dict[str, list[str]]:
-        # BigQuery exposes primary keys through the dataset-qualified INFORMATION_SCHEMA constraint views.
-        return self.create_metadata_primary_keys_query().execute(dataset_prefixes, dataset_names)
 
     def _build_table_namespace_for_schema_query(self, prefixes: list[str]) -> tuple[DataSourceNamespace, str]:
         table_namespace: DataSourceNamespace = BigQueryDataSourceNamespace(
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -114,7 +114,7 @@ class BigQueryDataSourceImpl(DataSourceImpl, model_class=BigQueryDataSourceModel
             sql_dialect=self.sql_dialect, data_source_connection=self.data_source_connection
         )
 
-    def get_primary_keys(self, dataset_prefixes: list[str], dataset_names: list[str]) -> dict[str, set[str]]:
+    def get_primary_keys(self, dataset_prefixes: list[str], dataset_names: list[str]) -> dict[str, list[str]]:
         # BigQuery exposes primary keys through the dataset-qualified INFORMATION_SCHEMA constraint views.
         return self.create_metadata_primary_keys_query().execute(dataset_prefixes, dataset_names)
 
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -17,6 +17,8 @@ from soda_core.common.sql_ast import (
     COLUMN,
     CONCAT_WS,
     COUNT,
+    CREATE_TABLE,
+    CREATE_TABLE_IF_NOT_EXISTS,
     DISTINCT,
     LITERAL,
     PERCENTILE_WITHIN_GROUP,
@@ -31,6 +33,9 @@ from soda_core.common.sql_ast import (
     WITH,
 )
 from soda_core.common.sql_dialect import SqlDialect
+from soda_core.common.statements.metadata_primary_keys_query import (
+    MetadataPrimaryKeysQuery,
+)
 from soda_core.common.statements.metadata_tables_query import MetadataTablesQuery
 
 logger: logging.Logger = soda_logger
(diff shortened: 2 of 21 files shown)


m1n0 (approved):
- Redshift opts into the same generic information_schema query flagged on #2811 — Redshift's catalog is PG-lineage, so the same "SELECT-only user sees no constraints" caveat likely applies there too. v3 ran this exact query but wrapped it in OptionalQuery and warned when it came back empty (soda-library data_source.py:680); v4 returns {} silently. Worth verifying with a read-only Redshift user, or carrying the v3 warning into the base query as part of the #2811 rework.
- Snowflake is a faithful port of v3's SHOW PRIMARY KEYS (same positional columns, matching the documented output). Two nits: _SCHEMA_NAME_COLUMN is unused, and it's the one new override without a unit test — a faked-connection test pinning the SHOW statement + row indices would be cheap.
- The leak-test fix (get_schema_prefix_index() instead of a hard default_prefix[1]) is right, and the BigQuery/DuckDB/Synapse SQL-shape tests are exactly the kind #2811's base query is still missing.

Opt-outs (Trino/Athena/sparkdf/Fabric/Hive) all check out against what those engines actually support.

Niels-b (commented):
Reviewed at head `09ec40a9`. This is a well-built PR — the per-dialect opt-in/opt-out decisions all check out against the vendor catalogs, the set→ordered-list refactor is consistent across core + the capture-schema consumer + the enterprise sources, the PK-DDL deepcopy fixes the object-mutation bug at the source, and the Postgres `pg_catalog` switch is a genuinely good fix. One headline gap and a couple of secondary issues below.

## Redshift is left on `information_schema` and will silently return no PKs for read-only users

The read-only-permission concern raised on #2811 was fixed for Postgres in this PR (the new `PostgresMetadataPrimaryKeysQuery` reads `pg_catalog` precisely because `information_schema` is privilege-filtered) — but **Redshift still delegates to the base `information_schema` query**:

- `RedshiftDataSourceImpl(DataSourceImpl, ...)` extends the base, not `PostgresDataSourceImpl`, and doesn't override `create_metadata_primary_keys_query`, so it never picks up the pg_catalog fix. Its comment still reads "Redshift exposes primary keys through the standard information_schema constraint views" — the exact reasoning the Postgres change just refuted.
- Redshift is a PostgreSQL 8.0.2 fork whose `information_schema.table_constraints` is **owner-scoped** (even stricter than modern Postgres' "owns or has a privilege other than SELECT"). A read-only monitoring role — the recommended/provisioned production grant — sees zero constraint rows, so `get_primary_keys` returns `{}` with no error. Onboarding then uploads empty PKs to Cloud, which defeats the purpose of capture-schema for the common grant.
- We've already hit and fixed this exact class of bug on Redshift: SCS-1193 (`b3a2a9df`) abandoned `information_schema.schemata` for `svv_all_schemas` for the same ownership-visibility reason, and the Redshift dialect already routes tables/columns/schemata through `pg_catalog`. The new PK path bypasses that established pattern.
- CI can't catch this: the redshift job connects as the table owner (the harness creates the table on the same connection it then introspects), so the green check exercises only the owner path — not the read-only path this feature runs as in production.

Suggested direction: give Redshift its own `pg_catalog`/`svv_`-based PK override, consistent with its existing `svv_*` overrides, and validate it against a live Redshift **as a SELECT-only role**. Note it likely can't reuse the Postgres query verbatim — Redshift's leader-node catalog has limited array-function support (`array_position` / `= ANY(array)` may not be available), so it needs its own formulation. If a full override isn't in scope for this PR, the minimum bar is a warning when tables are found but the PK query returns empty, plus a documented known-limitation — today the failure is completely silent.

(Severity hinges on the onboarding connection being read-only; if it ever runs as an owner/admin the impact is lower.)

## Snowflake `SHOW PRIMARY KEYS IN SCHEMA` can silently truncate

`SnowflakeMetadataPrimaryKeysQuery.execute` runs one schema-wide `SHOW PRIMARY KEYS`, but SHOW returns a maximum of ~10,000 rows (one row per PK column per table). A large schema drops PKs for tables past the cap with no error. Since `dataset_names` is already known and bounded, per-table `SHOW PRIMARY KEYS IN TABLE <db>.<schema>.<table>` avoids the cap (and over-fetching); alternatively detect `len(rows) >= 10000` and warn/fall back. (Privilege model is fine — SHOW needs only one access privilege, so read-only works.)

## Runtime opt-out isn't enforced (Databricks Hive / Fabric)

The capture-schema consumer calls `get_primary_keys` unconditionally — it doesn't consult `supports_primary_keys()`; only the integration test gates on it. But `DatabricksDataSourceImpl.get_primary_keys` unconditionally hits `information_schema`, so a **Databricks Hive-metastore** source (no `information_schema`) would raise at runtime — the `DatabricksHiveSqlDialect.supports_primary_keys()=False` gate is never consulted on this path (its docstring "gates itself off via DatabricksHiveSqlDialect" is misleading). **Fabric** similarly inherits SQL Server's delegate and would return real PKs despite opting out. Suggest gating inside the impl (`if self._is_hive_catalog(): return {}`) and/or having the consumer respect `supports_primary_keys()`.

## Test gap — the Snowflake SHOW path has no unit coverage

The riskiest code in the PR (hardcoded SHOW column offsets 2/3/4/5, `int()` cast, composite sort, exact-case matching) is exercised only by the live-Snowflake integration job. A stubbed-cursor unit test feeding representative SHOW rows (including out-of-order `key_sequence` and a non-requested table) is cheap and would pin the offsets/ordering/case-matching. The new `test_primary_keys_metadata.py` covers the base information_schema JOIN, not this path.

## Minor

- Snowflake positional parsing by fixed index is fragile if SHOW output order ever changes; a name-based lookup from `query_result.columns` would be safer and self-documenting (indices are correct today).
- `test_bigquery_dialect.py`: the comment "its columns must be NOT NULL" states a Soda choice as if it were a BigQuery grammar rule — BigQuery's own CREATE TABLE examples declare PKs without NOT NULL. Worth rewording; the `NOT ENFORCED` assertion is correct.

## Solid / verified

- set→ordered-`list` return type is consistent across the base query, `DataSourceImpl`, all OSS sources here, the soda-extensions `capture_schema_run.py` consumer (already list-based, no drift), and the enterprise DB2/DB2z/Oracle sources (inherit the base `execute`/`get_results`) — resolves the "PK order lost" note from #2811.
- PK-DDL mutation fix is real: `_create_table_with_primary_key_columns_not_null` deepcopies before setting `nullable=False`; Synapse's custom HEAP builder consumes the copy and doesn't mutate the caller.
- DDL correctness: BigQuery/Synapse `NOT ENFORCED` (+ Synapse `NONCLUSTERED`) required and correct; Fabric opt-out (no inline PK grammar) accurate; DuckDB composite-key handling relies on a fix present in the pinned `duckdb>=1.2.0`; Trino/Athena opt-outs factually right.
- Opt-in/opt-out inheritance matrix is complete — SparkDF←Databricks, Fabric/Synapse←SqlServer, DatabricksHive←Databricks all resolve to the right default.
- `test_table_metadata.py` `schema_index` change correctly handles single-prefix sources (DuckDB).

mivds:
Addressed the review in ba8eb0c3 (+ 2b775040 for Redshift earlier):

- **Redshift** — own `pg_catalog.pg_constraint` query. No `array_position()` on Redshift, so key columns are parsed from `pg_get_constraintdef(oid)` (the function `v_generate_tbl_ddl` uses), which renders them in declared order.
- **Runtime opt-out now enforced centrally** — base `get_primary_keys` gates on `sql_dialect.supports_primary_keys()` and delegates to `create_metadata_primary_keys_query()`; the seven per-source unconditional delegates are removed. Databricks-Hive returns `{}` without touching information_schema; Fabric's opt-out is now effective. Unit tests pin both gate directions (opt-out → no query built).
- **Snowflake SHOW cap** — replaced the schema-wide SHOW with a two-step flow: `INFORMATION_SCHEMA.TABLE_CONSTRAINTS` (real view — filterable, no ~10K cap) narrows to the requested tables that have a PK, then one `SHOW PRIMARY KEYS IN TABLE` per such table (bounded output). Also switched SHOW parsing to name-based lookup from the cursor description per the minor note.
- **Snowflake unit coverage** — stubbed-connection tests pin the TABLE_CONSTRAINTS statement shape, the per-table SHOW statements, out-of-order `key_sequence` sorting, absence of no-PK tables, and the name-based (reordered-output) parsing. `_SCHEMA_NAME_COLUMN` and friends are gone with the rewrite.
- **BigQuery test comment** — reworded: NOT NULL on key columns is Soda's uniform DDL choice, not BigQuery grammar.

Deliberately NOT done: removing the Oracle/DB2/DB2z delegates in soda-extensions#533 — extensions resolves soda-core from the published dev index (`>=4.19.0.dev2695`), which predates the central gate, so those delegates are the only working PK path there until this PR is released. They're forward-compatible; cleanup is a follow-up after the core release.

PR #2815, opened 2026-08-02
refactor(cli): single mark-scan-failed site at the CLI failure boundary

mivds:
## Problem

Two related defects in how early scan failures (before any results exist) reach Soda Cloud:

**1. Silent success on undelivered failures.** The exit code is a delivery signal the launcher relies on:

| exit | meaning | launcher action |
|---|---|---|
| `LOG_ERRORS` (3) | the engine already marked the scan FAILED on Cloud | trust it, container exits 0 |
| `RESULTS_NOT_SENT_TO_CLOUD` (4) | nothing reached Cloud | launcher marks the scan failed itself |

`handle_verify_contract` returned `LOG_ERRORS` (3) for **every** exception out of `verify_contract`, including failures whose Cloud report never landed — the launcher trusts exit ≤3, k8s marks the Job Succeeded, and a failed scan shows up as success with an empty Cloud record.

**2. Multiple `sodaCoreMarkScanFailed` per failure.** A failed managed scan marked the Cloud scan FAILED up to **three times**: the session's abort-on-first-error pre-reraise hook, `verify_contract`'s except arm, and the handler's failure report. Verified against the backend: the repeats are *accepted* (`markScanFailed` skips the scan-state check, and the instruction transition short-circuits on final states) — but each mark re-dispatches `FullScanEndedEvent` / `SodaCoreScanSuccessfulEvent` / `ManagedScanStateChangedEvent`, and `ManagedScanNotificationModule` sends failed-scan notifications per event **with no dedupe**, plus `promoteOngoingScanLogsToParquet` runs again per mark.

## Fix

Make the CLI failure boundary (`run_with_failure_reporting` → `report_scan_execution_failure`) the **only** Cloud-marking site for escaped exceptions:

- **`session.py`**: the abort-on-first-error paths re-raise verbatim; `_report_runner_scan_failed_before_reraise` is deleted.
- **`verify_api.py`**: `verify_contract`'s `except: mark + re-raise` arm is deleted. Failures propagate raw. This is invisible to programmatic callers: they run without a launcher-created scan (no `SODA_SCAN_ID`), where `mark_scan_as_failed` already returned without sending anything. One narrow exception: an out-of-tree caller that sets `SODA_SCAN_ID` itself (emulating a managed run outside the launcher) previously got a Cloud mark from this arm via the env fallback and now gets only the raised exception.
- **`contract.py` / `cli.py`**: `handle_verify_contract` is a bare command like `handle_monitor` (takes the wrapper's `Logs`, propagates failures raw); the `cli.py` verify wiring resolves the reporting channel (`resolve_soda_cloud_for_failure_report`, now in `dependencies.py`, swallow-to-None with a debug log) and wraps the command in `run_with_failure_reporting` — the same shape discover/monitor use. The reporter returns 3 when Cloud accepted the mark (or ad-hoc), 4 when it didn't so the launcher's fallback runs.
- **Logs threading**: `verify_contract` → `ContractVerificationSession.execute` → the session impl accept an optional `Logs` collector (the impl previously always constructed its own, displacing the wrapper's). The single mark therefore carries the **full engine log records** — fixing the earlier iteration's `log_records=None` report, which was the exact FAILED-with-no-logs Cloud record SAS-13001 is about.

Every managed scan enters through the CLI (the launcher builds `soda contract verify` / `soda monitor run` / `soda discover` command lines), so the boundary covers all of them. The invariant after this PR: **Cloud marking happens either at the send-results site as an upload substitution recorded on the result (which never raises), or at the CLI boundary for escaped exceptions — never both.**

## Tests

TDD'd — the new invariants were written first and failed for the right reasons (2 marks at handler level, 3 end-to-end):

- session abort paths (construction + verify) re-raise **without** any `sodaCoreMarkScanFailed`;
- end-to-end through the verify wiring shape (real session, duckdb): **exactly one** mark carrying the engine logs + exit 3, for **both** the verify abort and the phase-1 construction abort; rejected mark → exit 4 (SAS-13001 regression tests, relocated from session-level to boundary-level);
- success-path E2E: a clean managed run uploads results and never touches mark-scan-failed (exercises the shared wrapper→session `Logs` collector on the normal path);
- exception-path exit codes: managed + undelivered (rejected mark) → 4, managed + no channel (broken Cloud config) → 4, managed + delivered → 3, ad-hoc → 3.

Full suites green: unit 1178, feature 13, integration 165 (+ skips).

## Related

- Motivated by review on sodadata/soda-extensions#543 (monitor's `abort_on_first_error=True` double-marks until this ships): once the soda-core floor bumps, monitor single-marks with no soda-extensions change.
- Launcher follow-up (from review): the launcher's exit-4 mark-failed fallback is wired for discover/profile/monitor and batched scans but not for `verify` — verify's undelivered failures are now visible (non-zero exit) but rely on the orchestrator marking on non-zero exit until the launcher wires the >3 fallback for verify.

### DIFF
--- a/soda-core/src/soda_core/cli/handlers/contract.py
+++ b/soda-core/src/soda_core/cli/handlers/contract.py
@@ -10,7 +10,6 @@ from soda_core.common._deprecation import deprecated_kwarg
 from soda_core.common.exceptions import (
     ContractParserException,
     InvalidArgumentException,
-    InvalidDataSourceConfigurationException,
 )
 from soda_core.common.logging_constants import Emoticons, soda_logger
 from soda_core.common.yaml import ContractYamlSource
@@ -46,7 +45,20 @@ def handle_verify_contract(
         use_runner = False
     if check_selectors is None:
         check_selectors = []
-    try:
+
+    # Local import avoids a soda_cloud <-> contracts.api cycle.
+    from soda_core.cli.handlers.dependencies import run_with_failure_reporting
+    from soda_core.common.logs import Logs
+
+    # The reporting channel resolves first, outside the wrapped command — it is the single
+    # Cloud-marking site for failures (delivery-aware, as discover/monitor: exit 3 when Cloud
+    # has the failure or the run is ad-hoc, 4 when a managed run's failure couldn't reach
+    # Cloud so the launcher's fallback reports). A broken Cloud config resolves to None here;
+    # the command re-raises the real config error (verify_contract builds its own client) and
+    # the boundary reports it through the None channel.
+    soda_cloud = _resolve_soda_cloud_for_failure_report(soda_cloud_file_path, variables)
+
+    def verify_and_interpret(logs: Logs) -> ExitCode:
         dwh_files: Optional[DiagnosticsWarehouseFiles] = None
         if diagnostics_warehouse_file_path or metadata_dwh_file_path:
             dwh_files = DiagnosticsWarehouseFiles(
@@ -68,21 +80,14 @@ def handle_verify_contract(
             check_paths=check_paths,
             check_selectors=check_selectors,
             dwh_data_source_file_path=dwh_files,
+            # The wrapper's collector, threaded into the session so every construction/
+            # verify record lands in the collector the failure report captures from.
+            logs=logs,
         )
 
         return interpret_contract_verification_result(contract_verification_result)
 
-    except (InvalidArgumentException, InvalidDataSourceConfigurationException, Exception) as exc:
-        # Shared delivery-aware reporting (as discover/monitor do): marks the managed scan failed and
-        # returns 3 when Cloud has the failure (or ad-hoc), 4 when it doesn't so the launcher reports.
-        # Local import avoids a soda_cloud <-> contracts.api cycle.
-        from soda_core.cli.handlers.failure_reporting import (
-            report_scan_execution_failure,
-        )
-
-        soda_logger.exception(f"An unexpected exception occurred: {exc}")
-        soda_cloud = _resolve_soda_cloud_for_failure_report(soda_cloud_file_path, variables)
-        return report_scan_execution_failure(soda_cloud, log_records=None)
+    return run_with_failure_reporting(soda_cloud, verify_and_interpret)
 
 
 def _resolve_soda_cloud_for_failure_report(
--- a/soda-core/src/soda_core/cli/handlers/contract.py
+++ b/soda-core/src/soda_core/cli/handlers/contract.py
@@ -1,6 +1,9 @@
 from os.path import dirname, exists
 from pathlib import Path
-from typing import Dict, Optional
+from typing import TYPE_CHECKING, Dict, Optional
+
+if TYPE_CHECKING:
+    from soda_core.common.soda_cloud import SodaCloud
 
 from soda_core.cli.exit_codes import ExitCode
 from soda_core.common._deprecation import deprecated_kwarg
@@ -70,8 +73,31 @@ def handle_verify_contract(
         return interpret_contract_verification_result(contract_verification_result)
 
     except (InvalidArgumentException, InvalidDataSourceConfigurationException, Exception) as exc:
+        # Shared delivery-aware reporting (as discover/monitor do): marks the managed scan failed and
+        # returns 3 when Cloud has the failure (or ad-hoc), 4 when it doesn't so the launcher reports.
+        # Local import avoids a soda_cloud <-> contracts.api cycle.
+        from soda_core.cli.handlers.failure_reporting import (
+            report_scan_execution_failure,
+        )
+
         soda_logger.exception(f"An unexpected exception occurred: {exc}")
-        return ExitCode.LOG_ERRORS
+        soda_cloud = _resolve_soda_cloud_for_failure_report(soda_cloud_file_path, variables)
+        return report_scan_execution_failure(soda_cloud, log_records=None)
+
+
+def _resolve_soda_cloud_for_failure_report(
+    soda_cloud_file_path: Optional[str], variables: Optional[Dict[str, str]]
+) -> "Optional[SodaCloud]":
+    """Soda Cloud channel for reporting an early verify failure, or None when Cloud isn't configured
+    or can't be built (report_scan_execution_failure then returns 3 for ad-hoc, 4 for a managed run)."""
+    from soda_core.common.soda_cloud import SodaCloud
+
+    if not soda_cloud_file_path:
+        return None
+    try:
+        return SodaCloud.from_config(soda_cloud_file_path, variables)
+    except Exception:
+        return None
 
 
 def contract_verification_is_not_sent_to_cloud(
--- a/soda-core/src/soda_core/cli/handlers/contract.py
+++ b/soda-core/src/soda_core/cli/handlers/contract.py
@@ -1,9 +1,6 @@
 from os.path import dirname, exists
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, Optional
-
-if TYPE_CHECKING:
-    from soda_core.common.soda_cloud import SodaCloud
+from typing import Dict, Optional
 
 from soda_core.cli.exit_codes import ExitCode
 from soda_core.common._deprecation import deprecated_kwarg
@@ -12,6 +9,7 @@ from soda_core.common.exceptions import (
     InvalidArgumentException,
 )
 from soda_core.common.logging_constants import Emoticons, soda_logger
+from soda_core.common.logs import Logs
 from soda_core.common.yaml import ContractYamlSource
 from soda_core.contracts.api import test_contract, verify_contract
 from soda_core.contracts.api.publish_api import publish_contract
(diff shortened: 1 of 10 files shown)


soda-core/src/soda_core/cli/handlers/contract.py
```diff
@@ -65,13 +80,29 @@ def handle_verify_contract(
...
+            # verify record lands in the collector the failure report captures from.
+            logs=logs,
         )
 
         return interpret_contract_verification_result(contract_verification_result)
 
-    except (InvalidArgumentException, InvalidDataSourceConfigurationException, Exception) as exc:
-        soda_logger.exception(f"An unexpected exception occurred: {exc}")
-        return ExitCode.LOG_ERRORS
+    return run_with_failure_reporting(soda_cloud, verify_and_interpret)
```
Niels-b: Minor consistency nit: `verify` wraps `run_with_failure_reporting` *inside* the handler and resolves its own failure channel, whereas `discover` wraps at the `cli.py` wiring layer (the pattern the module docstrings describe — "dependency construction lives at the wiring layer"). Not a bug, but it makes `verify` the odd one out. Consider moving the wrap + channel resolution into the `cli.py` verify wiring for consistency, or add a one-line note on why verify differs.

soda-core/src/soda_core/cli/handlers/contract.py
```diff
@@ -65,13 +80,29 @@ def handle_verify_contract(
...
+) -> "Optional[SodaCloud]":
+    """Soda Cloud channel for reporting an early verify failure, or None when Cloud isn't configured
+    or can't be built (report_scan_execution_failure then returns 3 for ad-hoc, 4 for a managed run)."""
+    from soda_core.common.soda_cloud import SodaCloud
+
+    if not soda_cloud_file_path:
+        return None
+    try:
+        return SodaCloud.from_config(soda_cloud_file_path, variables)
+    except Exception:
```
Niels-b: Fragility (minor): this bare `except Exception: return None` is correct today *only* because `verify_contract` re-constructs the same client and re-raises the real config error to the boundary (which then maps a managed run → 4). If a future change makes `verify_contract` tolerant of a null client, a genuinely broken managed Cloud config would silently become a bare exit-4 with no diagnostic. Suggest narrowing the catch to the expected config exceptions (mirror `resolve_soda_cloud`'s `InvalidSodaCloudConfigurationException` / `YamlParserException` / `ValueError`) and/or logging at debug when returning `None`, so the reason is never fully invisible.

soda-core/src/soda_core/contracts/api/verify_api.py
```diff
@@ -290,61 +292,58 @@ def verify_contract(
     else:
         contract_file_paths = None
 
+    # Failures propagate raw, without touching Cloud: the CLI failure boundary
```
Niels-b, line 295: Nit (worth a changelog/docstring line): removing the `except` arm also drops `verify_contract`'s engine-side `soda_logger.error(exc)` on failure. CLI callers still get it at the boundary (`run_with_failure_reporting` → `soda_logger.exception`), but **programmatic callers** now get the raw exception with no engine-side error log. Also a narrow behavior change: an out-of-tree caller that sets `SODA_SCAN_ID` itself previously got a Cloud mark here and now gets only the raised exception (ordinary callers without `SODA_SCAN_ID` are unaffected — that path was already a no-op).

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -215,7 +216,10 @@ def execute(
         if soda_cloud_use_runner_blocking_timeout_in_minutes is None:
             soda_cloud_use_runner_blocking_timeout_in_minutes = 60
 
-        logs: Logs = Logs()
+        # A caller-supplied collector (the CLI failure-reporting wrapper's) is reused so
+        # every session record reaches its failure report; constructing an inner Logs
+        # here would displace the caller's active collector.
+        logs: Logs = logs if logs is not None else Logs()
```
Niels-b, line 222: Verified this reuse is safe: the wrapper's `Logs` is threaded through every hop, the impl/session never `close()` the caller-supplied collector, and `report_scan_execution_failure` reads `get_log_records()` before the wrapper's `finally: close()`, so records survive to the single mark. One residual: the `use_runner=True` path doesn't thread this `logs` into `_execute_on_runner`, so escaped-to-boundary failures there are narrower on log capture — worth a test if the runner path matters.

soda-tests/tests/unit/test_contract_marks_scan_failed_on_connection_error.py
```diff
@@ -267,15 +278,83 @@ def _boom_init(self, *args, **kwargs):
...
+            verbose=False,
+            use_runner=False,
+            blocking_timeout_in_minutes=10,
+            check_paths=None,
+            check_selectors=[],
+            diagnostics_warehouse_file_path=None,
+        )
+
+
+def test_cli_boundary_marks_scan_failed_exactly_once_with_engine_logs(monkeypatch, tmp_path):
```
Niels-b, line 323: Coverage here is strong (exactly-one-mark-with-logs E2E, rejected-mark → 4). Three gaps worth adding:

1. A **success-path E2E** through `handle_verify_contract` (real duckdb) asserting exit 0, `_mark_requests(mock_cloud) == []`, and results uploaded — the path users hit constantly, and the only one that exercises the shared-`Logs` change on the *normal* path.
2. A **construction-abort E2E** through the boundary. This file only covers the *verify*-abort E2E; the phase-1 construction abort is only checked at session level. The PR's thesis is "the boundary now catches what the session used to" for *both* phases.
3. A **None / broken-cloud-config managed run → exit 4** test. `test_handle_verify_contract_early_failure_undelivered_exits_results_not_sent` actually hits the *rejected-mark* branch (mock returns non-None, `mark → False`), not the `soda_cloud is None` branch — so the no-channel path is currently untested.

Niels-b (approved):
**Approve** — thorough refactor, and it fixes both defects it targets. I validated the load-bearing parts (including a runtime check on the Logs lifecycle) and they hold:

- **Exit-code mapping is correct** for all four cases (ad-hoc → 3, delivered → 3, undelivered/rejected/throwing-mark → 4, no-channel managed → 4), and the old handler's always-3 "silent success on undelivered failure" is genuinely fixed.
- **Single-mark invariant holds**: the only remaining `mark_scan_as_failed` sites are the two send-results substitution paths (`base.py`, `session.py`) and this CLI boundary — mutually exclusive via `abort_on_first_error`. `session.py` and `verify_contract` no longer mark.
- **Logs survive to the single mark** (no double-close / empty-logs regression: `close()` doesn't clear records and the boundary reads `get_log_records()` before the wrapper's `finally: close()`), so the mark carries real engine logs — the SAS-13001 fix is genuine.
- **`verify_contract` un-indent is faithful**; backward-compat holds (the deleted `mark_scan_as_failed(exc=exc)` was already a no-op without `SODA_SCAN_ID`).

**One cross-repo caveat worth tracking (not a blocker here):** I checked the launcher — its exit-4 fallback (`mark_scan_failed_safely`) is wired for `_run_v4_dataset_scan` (discover/profile/monitor) and `batched_scan`, but **not** for the `verify` command. The `command_executor` gate is `return_code > 3`, yet `cli.py::verify` only does `except Exception: exit_code = 1; raise`, and the `ExecutionContextManager` marks only *pre-execution* failures. So for a contract verify whose failure couldn't reach Cloud, exit 4 now makes the failure **visible** (non-zero container exit — strictly better than the old silent exit-3), but the launcher won't itself mark the Cloud scan FAILED unless the orchestrator does so on any non-zero exit. This fully closes the loop for **monitor** (the #543 goal); worth a launcher follow-up / orchestrator confirmation for **verify**.

The inline notes below are all minor/nits — none block merge.

soda-core/src/soda_core/cli/handlers/contract.py
mivds: Moved in 8235068e: handle_verify_contract is now a bare command taking the wrapper's logs (like handle_monitor); the cli.py verify wiring resolves the channel and wraps. The swallow-to-None resolver moved to dependencies.py as resolve_soda_cloud_for_failure_report.

soda-core/src/soda_core/cli/handlers/contract.py
mivds: Added a debug log with the swallowed cause in 585fb885 so a bare exit-4 is never undiagnosable. Kept the broad catch deliberately: the resolver now runs in the cli.py wiring *outside* the failure boundary, so an exception escaping a narrowed catch would crash the CLI with a raw traceback instead of mapping to a delivery-aware exit code. The no-channel branch is also covered now (test_handle_verify_contract_broken_cloud_config_managed_exits_results_not_sent).

soda-core/src/soda_core/contracts/api/verify_api.py
mivds: Documented in the PR body rather than code: programmatic callers get the raised exception with the full traceback in their own hands, so the deleted error(exc) was redundant with that. The narrow SODA_SCAN_ID-setting out-of-tree case is called out explicitly in the verify_api bullet — it's emulating a managed run outside the launcher, which isn't a supported pattern.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
mivds: Deferring the runner path: its collectors are impl-local and pre-existing (the impl's own Logs is closed right after construction; verify_contract_on_runner late-binds another), and ambient capture still routes soda_logger records to the wrapper's collector outside those windows — so escaped runner failures reach the boundary with the same partial logs as before this PR. Nothing regressed; improving runner-path capture is a separate change.

soda-tests/tests/unit/test_contract_marks_scan_failed_on_connection_error.py
mivds: All three added in d4b2aaef: success-path E2E (real duckdb, exit 0, no marks, results uploaded — with the scanId-carrying insert response the send path requires), construction-abort E2E through the boundary (exactly one mark with the phase-1 engine logs), and the broken-cloud-config managed run → exit 4 (the None-channel branch you spotted was untested).

PR #2816, opened 2026-08-03
feat(sql-dialect): add distinct parameter to select_all_paginated_sql

m1n0:
## What

Adds a `distinct: bool = False` keyword parameter to `SqlDialect.select_all_paginated_sql`, so a
dialect renders `SELECT DISTINCT` itself instead of the caller string-rewriting the SQL it returns.

- `soda-core`: `select_all_paginated_sql(..., distinct=False)`; the flag rides on the `SELECT` AST
  node (new `distinct` field) and is rendered by `_build_select_sql_lines`.
- `soda-sqlserver`: same parameter. `TOP n` is now emitted *after* `DISTINCT`, matching the T-SQL
  `SELECT [ALL | DISTINCT] [TOP n] <fields>` grammar (previously the TOP rewrite would have produced
  `SELECT TOP n DISTINCT ...`).
- `soda-synapse`: de-duplicates in a leading CTE that the paginating CTE then reads.

Default is `False`, so every existing call site renders byte-identical SQL.

## Why

Consumers that need a de-duplicated page of rows (reconciliation's reference-diff row iterators)
currently post-process the returned SQL with `paginated_sql.replace("SELECT", "SELECT DISTINCT", 1)`.
That assumes the statement starts with the SELECT that should be de-duplicated. It does not hold for
Synapse dedicated SQL pools, which have no `OFFSET ... FETCH` and therefore paginate through a
`ROW_NUMBER()` CTE: the leading token there belongs to the *inner* select, so the rewrite
de-duplicates the wrong rows — and is a silent no-op anyway, because `ROW_NUMBER()` makes every row
unique. The in-code comment on the existing hack already prescribed this parameter as the fix.

Pushing it down to the dialect also removes a landmine before more SQL shapes (custom per-side
queries, per-side filters) are added on top of the same code path.

Per-dialect notes:

- Base / Postgres / Snowflake / BigQuery / Databricks / Redshift / DuckDB / Athena / Trino:
  `SELECT DISTINCT <cols> ... ORDER BY ... LIMIT/OFFSET`.
- SQL Server: `SELECT DISTINCT <cols> ... ORDER BY ... OFFSET n ROWS FETCH NEXT m ROWS ONLY`.
- Synapse: DISTINCT can be applied neither alongside `ROW_NUMBER()` (no-op) nor on the outer select
  (T-SQL rejects `SELECT DISTINCT ... ORDER BY <rn>`, and `rn` must stay projected out), so it goes
  into an extra leading `deduplicated` CTE. This requires the `order_by` columns to be part of
  `columns`, which holds for the callers that ask for distinct.

Internal ticket: DTL-1818.

## Stack position

Base of the stack — nothing merges before this one. A **soda-extensions** PR consumes the new
parameter (it replaces the `"SELECT" -> "SELECT DISTINCT"` string rewrite in the reconciliation row
iterators) and must merge **after** this PR.

## Test evidence

All commands run from the repo root on this branch.

Full unit suites (engine + every connector package):

```
uv run pytest soda-tests/tests/unit soda-tests/tests/feature \
  soda-postgres/tests/unit soda-snowflake/tests/unit soda-databricks/tests/unit \
  soda-duckdb/tests soda-bigquery/tests/unit soda-sqlserver/tests/unit \
  soda-synapse/tests/unit soda-redshift/tests/unit soda-fabric/tests/unit \
  soda-athena/tests/unit soda-trino/tests/unit -q

1 failed, 1429 passed, 2 skipped, 16 warnings in 3.08s
```

The single failure is `soda-duckdb/tests/data_sources/test_duckdb_connections.py::test_connection_reused_in_memory`
(`ConnectionException: Connection already closed!`). It is **pre-existing and unrelated**: with this
branch's changes stashed, the same selection reproduces it — `1 failed, 1194 passed, 2 skipped`. It
only fails when the duckdb data-source tests run in the same session as the `soda-tests` suites; the
duckdb package's own suite passes standalone (`uv run pytest soda-duckdb/tests -q` -> `17 passed`).

New tests only:

```
uv run pytest soda-tests/tests/unit/test_sql_generation.py soda-sqlserver/tests/unit \
  soda-synapse/tests/unit soda-duckdb/tests/unit -q

88 passed, 6 warnings in 0.17s
```

Added coverage:

- `soda-tests/tests/unit/test_sql_generation.py` — base dialect: not-distinct default, distinct,
  distinct + filter, distinct with no columns/order-by (`SELECT DISTINCT *`), and the `SELECT` AST
  node flag.
- `soda-sqlserver/tests/unit/test_sqlserver_dialect.py` — paginated select with and without
  distinct, plus `SELECT DISTINCT TOP n` ordering (and the unchanged `SELECT TOP n` form).
- `soda-synapse/tests/unit/test_synapse_dialect.py` — the unchanged ROW_NUMBER shape, the
  de-duplicating CTE shape, and the no-filter/no-order-by variant.
- `soda-duckdb/tests/unit/test_duckdb_dialect.py` — executes the rendered SQL against an in-memory
  DuckDB, so DISTINCT is verified on real semantics: duplicates kept without the flag, removed with
  it, de-duplication happening before LIMIT/OFFSET (pages don't overlap), and distinct + filter.

Pre-commit on the changed files: all hooks pass (`trailing-whitespace`, `check-ast`, `autoflake`,
`isort`, `black`).

Not run: live data-source suites (Postgres/SQL Server/Synapse/Snowflake/etc.) — the local Docker
daemon is unavailable and the remote warehouses are not reachable from this environment. The Synapse
and SQL Server changes are covered by rendering tests only; DuckDB provides the executed-SQL evidence.

## Post-merge update (2026-08-05)

Merged `origin/main` into this branch (merge, not rebase) to resolve against #2812
("default-off engine hooks for cross-source reconciliation", released in v4.20.0). #2812 added
`normalize_key_columns` to this same `select_all_paginated_sql`, so the two features had to be
*composed*, not just stitched together.

- The signature keeps main's parameter order with `distinct` appended last, on the base dialect and
  the SQL Server / Synapse overrides.
- `distinct=True` alone, and `normalize_key_columns` alone, render exactly as before. Together they
  cannot: SQL requires every ORDER BY term of a `SELECT DISTINCT` to appear in its select list, and
  #2812's case-fold (`LOWER(<key>)`) is deliberately *not* projected — callers read a page's rows
  positionally, so the projection must stay exactly `columns`. The combination therefore
  de-duplicates in a CTE and orders/paginates that CTE — the same "de-duplicate in an inner scope"
  shape Synapse already used — which keeps DISTINCT before LIMIT/OFFSET, so pages still neither
  overlap nor carry duplicates.
- That rule lives in one new shared helper, `SqlDialect._paginated_select_statements`, which the SQL
  Server override now reuses (it only picks the OFFSET/FETCH element order).
- Synapse composes with no shape change — its `ROW_NUMBER()` select is not DISTINCT, so the fold is
  legal inside `OVER(...)` — and now *enforces*, rather than assumes in a comment, that its
  de-duplicated CTE projects every order-by column. `_validate_distinct_pagination` raises a clear
  `ValueError` for `distinct` + an unprojected order-by column on every dialect, instead of shipping
  SQL that Postgres/T-SQL/BigQuery reject.

Tests: the distinct x normalize matrix (distinct only / normalize only / both) as rendered-SQL
goldens on the base dialect, SQL Server and Synapse, plus executed DuckDB coverage of the composed
shape (de-duplication, case-insensitive order with the raw tiebreaker, non-overlapping pages) and
the rejection cases.

```
uv run pytest -p no:helpers.memory_container_plugin soda-tests/tests/unit soda-tests/tests/feature \
  soda-duckdb/tests soda-sqlserver/tests/unit soda-synapse/tests/unit -q

1 failed, 1293 passed, 2 skipped, 14 warnings in 2.34s
```

The single failure is the same pre-existing, unrelated
`test_duckdb_connections.py::test_connection_reused_in_memory` documented above — it still fails
with this branch's duckdb test file deselected. The remaining connector unit suites (fabric,
postgres, trino, snowflake, bigquery, athena, redshift, databricks) are green: `197 passed`.

Rebased on main (2026-08-26): merged 23 further commits — #2841/#2843 made `_build_regex_like_sql` `@final` behind two new seams (so SQL Server's override became `_rewrite_regex_pattern` + `_regex_like_sql`) and #2842 widened isort to 120. Conflicts were import lists plus side-by-side test additions; the distinct x normalize composition is unchanged. `1393 passed, 2 skipped`, with the same lone pre-existing duckdb connection-reuse flake (that suite is 24/24 on its own).

## Review round (2026-08-31)

Follow-ups on @mivds's review (4a1f7902).

- **`order_by` / `columns` were annotated `list[str]` while the code `isinstance`-checked them.**
  The annotation was wrong, not the check: a caller may pass an expression it built itself
  (soda-reconciliation's per-side column expressions project `(<expr>) AS "<col>"` and order by
  `(<expr>)`). That union now has a name — `SqlColumnTerm = Union[str, SqlExpression]` in
  `sql_ast` — and both parameters use it, on the base dialect and the SQL Server override.
  Synapse deliberately keeps `list[str]`: it hand-builds raw SQL and pushes every term through
  the identifier quoter, so it takes names only (documented on the method).

- **"Why not use the CTE in both cases?"** Because the CTE is not a superset of the flat shape.
  Its outer select reads from the CTE, so it can only render terms resolving against the CTE's
  *output* columns; the flat select renders against the base table. An expression term binds in
  the flat shape and fails in the CTE (`column "a" does not exist` — verified on DuckDB), while
  the case-fold `LOWER(<key>)` is deliberately unprojected and so only works in the CTE. Their
  *union* is therefore expressible in neither shape — and the branch condition was letting that
  combination through as SQL the warehouse rejects. New `_validate_distinct_page_cte` raises a
  named error instead. **No behavior change for any caller:** reference_diff is the only
  `distinct=True` caller and it does not pass `normalize_key_columns`, so this is a tripwire for
  whoever wires the two together next, and the reasoning now lives in the docstring rather than
  being implied by the branch.

- **Executed-SQL duckdb tests moved out of `tests/unit`** into
  `soda-duckdb/tests/data_sources/test_duckdb_paginated_select.py` — they open a connection and
  run the rendered SQL, so they are integration tests even with duckdb in-process. The
  rendering-only goldens for the same method stay in `soda-tests/tests/unit`.

New coverage in `soda-tests/tests/unit/test_sql_dialect_seams.py`: the CTE shape rejects an
expression term beside a normalized key, and rejects an expression in the projection; an
expression term *without* normalization still renders flat (the shape that actually serves
expression callers); and the live distinct×normalize path is unchanged.

```
uv run pytest -p no:helpers.memory_container_plugin soda-tests/tests/unit soda-tests/tests/feature \
  soda-duckdb/tests soda-sqlserver/tests/unit soda-synapse/tests/unit -q

1 failed, 1397 passed, 2 skipped, 14 warnings
```

Same lone pre-existing `test_duckdb_connections.py::test_connection_reused_in_memory` documented
above (duckdb's own suite is 24/24 standalone). Remaining connector unit suites green:
`213 passed`. Pinned `isort` / `black` / `autoflake` clean on the changed files.

### Synapse: one paginator, not two (b7059619)

The `distinct` branch had been a copy of the entire ROW_NUMBER paginator with an extra leading
CTE spliced in. The two shapes only ever differ by *where the paginating CTE reads from* — the
table, or a `deduplicated` CTE that the filter moves into — so those two values are now picked
up front and the statement is rendered once:

```python
if distinct:
    self._validate_distinct_pagination(columns=columns, order_by=order_by)
    lead_cte = f"deduplicated AS (\n    SELECT DISTINCT {columns_csv}\n    FROM {qualified_table}\n    {where_sql}\n),\n"
    page_source, page_where = "deduplicated", ""
else:
    lead_cte, page_source, page_where = "", qualified_table, where_sql

return (
    f"WITH {lead_cte}paginated AS (\n"
    f"    SELECT {columns_csv}, ROW_NUMBER() OVER (ORDER BY {order_by_csv}) AS {rn_alias}\n"
    f"    FROM {page_source}\n"
    f"    {page_where}\n"
    ...
)
```

Behaviour is unchanged: the non-distinct SQL is byte-identical (its 14 goldens pass untouched),
and the distinct SQL differs only by a blank line inside the paginated CTE — the same one the
no-filter non-distinct shape already emits — so the 3 distinct goldens are updated to match.

Worth stating plainly why the Synapse block exists at all, since it is the least obvious part of
this PR: the caller on `main` does `paginated_sql.replace("SELECT", "SELECT DISTINCT", 1)`. On
Synapse that first `SELECT` belongs to the *inner*, pre-`ROW_NUMBER()` select — and `ROW_NUMBER()`
makes every row unique, so the `DISTINCT` is a no-op. **`reference_diff` against a Synapse
dedicated SQL pool silently returns duplicate rows today.** That is the bug this parameter fixes,
and the in-code comment above the string rewrite on `main` already prescribes exactly this fix.


soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -437,22 +440,125 @@ def select_all_paginated_sql(
...
 
         if filter:
             where_clauses.append(SqlExpressionStr(filter))
 
-        statements = [
-            SELECT(columns or [STAR()]),
+        order_by_terms = [term for c in order_by for term in self._order_by_key(c, normalize_key_columns)]
+
+        if distinct:
+            self._validate_distinct_pagination(columns=columns, order_by=order_by)
```
mivds, line 523: I don't quite understand the approach here. When using `normalize_key_columns`, we use a CTE to work around the constraint of `ORDER BY` columns having to be in the `SELECT` list. When using a regular `order_by`, we raise an error for the same constraint. Why not use the CTE in both cases?

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -437,22 +440,125 @@ def select_all_paginated_sql(
...
+            if any(isinstance(column, str) and column in normalize_key_columns for column in order_by):
```
mivds, line 524: The `order_by` argument is typed as `list[str]`. What do we expect here that is not a string?

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -437,22 +440,125 @@ def select_all_paginated_sql(
...
+        if not columns:
+            # No explicit projection -> `SELECT DISTINCT *` projects every column.
+            return
+        # Only plain column NAMES can be matched against the projection by name. A caller may
+        # also pass built expression terms (soda-reconciliation orders a side by a per-column
+        # SQL expression, projected as `(<expr>) AS "<col>"`); those are the caller's own
+        # terms, matched to its own projection, so this validation has nothing to say about
+        # them. Skipping them is what keeps the check a real check for the string case
+        # instead of a false alarm for every expression caller.
+        missing_order_by_columns = [column for column in order_by if isinstance(column, str) and column not in columns]
```
mivds, line 573: Same question as above, with `isinstance` when typing suggests we only accept strings?

soda-duckdb/tests/unit/test_duckdb_dialect.py
```diff
@@ -9,6 +13,124 @@ def test_random():
...
+
+
+def _connection_with_codes():
+    connection = duckdb.connect(":memory:")
+    connection.execute('CREATE TABLE "CODES" ("code" VARCHAR, "label" VARCHAR)')
+    connection.execute("INSERT INTO \"CODES\" VALUES ('b', 'x'), ('B', 'x'), ('a', 'y'), ('a', 'y'), ('C', 'z')")
+    return connection
+
+
+def _code_page(connection, *, distinct: bool, normalize: bool, limit: int = 10, offset: int = 0) -> list:
```
mivds: I would move this to a different file. Unit tests suggest fast in-memory tests. While this is executing queries, so I'd argue it's more an integration test. The fact it's duckdb I guess makes it more ambiguous, but I'd stick with the rule of thumb of not relying on actual connection here

mivds (approved):

soda-core/src/soda_core/common/sql_dialect.py
m1n0: Because the CTE isn't a superset of the flat shape — I'd assumed the same until I tried it.

Its outer select reads from the CTE, so it can only render terms that resolve against the CTE's *output* columns. The flat select renders against the base table. So a caller-built expression term — reconciliation projects `(<expr>) AS "col"` and orders by `(<expr>)` — binds in the flat shape and fails in the CTE with `column "a" does not exist` (checked on DuckDB, same on Postgres/BigQuery). The case-folded key is the opposite case: `LOWER(key)` is deliberately not projected, so it can only work in the CTE.

Which also means their *union* is expressible in neither shape, and the branch condition was letting that combination through as SQL the warehouse rejects. Nothing reaches it today — reference_diff is the only `distinct=True` caller and it doesn't normalize keys — so I've added `_validate_distinct_page_cte` to fail with a named error rather than leave it for whoever wires those two together next. Reasoning is in the docstring now instead of implied by the branch. 4a1f7902

soda-core/src/soda_core/common/sql_dialect.py
m1n0: An expression the caller built itself. Reconciliation's per-side column expressions order by `(<expr>)` instead of a column name, and project `(<expr>) AS "<col>"` — so `columns` was lying in the same way.

The annotation was wrong, not the check. Both are `SqlColumnTerm` (`Union[str, SqlExpression]`) now, with a note on the method saying why the isinstance guards exist. Synapse keeps `list[str]`: it hand-builds raw SQL and pushes every term through the identifier quoter, so it really does take names only. 4a1f7902

soda-core/src/soda_core/common/sql_dialect.py
m1n0: Same — `SqlColumnTerm` here too. Only a name can be matched against the projection, so the isinstance is what lets an expression term through rather than false-alarming on every one of them.

soda-duckdb/tests/unit/test_duckdb_dialect.py
m1n0: Agreed, it opens a connection and runs the SQL. Moved to `soda-duckdb/tests/data_sources/test_duckdb_paginated_select.py`; the rendering-only goldens for the same method stay in `soda-tests/tests/unit`. 4a1f7902

PR #2817, opened 2026-08-04
build(athena): loosen PyAthena to >=2.2.0,<4.0 (remove <3.0 cap, CVE-2026-65321)

Niels-b:
## What
Loosen the `soda-athena` PyAthena constraint from `>=2.2.0,<3.0` to **`>=2.2.0,<4.0`** — i.e. remove the `<3.0` cap so PyAthena **3.x** (which contains the fix for **CVE-2026-65321**) can resolve, while keeping 2.x installable and guarding against a future 4.x major.

## Why this shape
Mirrors `soda-snowflake`'s uncapped `snowflake-connector-python>=3.0`: keep the library constraint thin, and let the **deployable launchers pin the fixed floor** (`pyathena>=3.35.4`, following the Snowflake bump precedent). This avoids forcing 3.x onto every downstream `soda-athena` consumer.

## Verification
Resolves to PyAthena **3.35.4** today (latest `<4.0`) — the exact version already validated against **real Athena** on this PR (v4: `[REPLAY=OFF]` matrix; v3: the `athena` job). Environment is unchanged, so that verification still holds.

## Security context
Soda is **not exploitable** by CVE-2026-65321 on any PyAthena version — it always builds fully-rendered SQL and calls single-arg `cursor.execute(sql)`, never passing a `parameters` dict, so the vulnerable `DefaultParameterFormatter` branch is unreachable. This change is dependency hygiene + unblocking the launcher floor-pin, not an exploit fix.

### DIFF
--- a/soda-athena/pyproject.toml
+++ b/soda-athena/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 ]
 dependencies = [
     "soda-core==4.19.1rc0",
-    "PyAthena>=3.35.4,<4.0",
+    "PyAthena>=2.2.0,<4.0",
     "setuptools>=60",
 ]
 
--- a/soda-athena/pyproject.toml
+++ b/soda-athena/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 ]
 dependencies = [
     "soda-core==4.19.1rc0",
-    "PyAthena>=2.2.0,<3.0",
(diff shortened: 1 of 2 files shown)


Niels-b:
soda-extensions run: https://github.com/sodadata/soda-extensions/actions/runs/30893347225

mivds (approved):

PR #2818, opened 2026-08-05
feat(thresholds): additional threshold in check thresholds — contracts and data standards (DET-107)

m1n0:
A check threshold can carry a second threshold at the other level:

```yaml
threshold:
  must_be_less_than: 10          # fail (the default level)
  additional:
    must_be_less_than: 5         # warn
    level: warn
```

Rules: one comparison per threshold object; `additional` takes no `metric`/`unit`/`additional` of its own; the two effective levels must differ (no `level` means fail); either one can be on the outside. Evaluation is fail-first, one outcome per check. A warn threshold that can never fire on its own gets a log warning.

What changes for existing contracts: flat `level: warn` bounds now report under `diagnostics.warn` instead of `fail`; every check payload carries a `warn` diagnostics key (null when unused); the bundled schema accepts `level` and no longer rejects a flat threshold with several comparisons — arity is only enforced next to `additional`, same as the Cloud schema. Evaluation of thresholds without `additional` is unchanged, pinned by tests.

Data standards get this too: same language, same engine.

Tests: unit suite 1254 green; bundled-schema matrix 48/48 under Draft 7 and 2020-12.

Ships in the first release cut after merge — 4.21.0 and 4.21.1 went out without it. Companions: soda#13165 and soda#13015 (Cloud), soda-extensions#547 (migrator, group_by; pin bump is its gate), docs#1034.

```customer-facing-release-note
Contract and data standard checks can now set warn and fail thresholds together.
```


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -901,30 +933,30 @@ def create(
...
+            return ThresholdImpl(
+                type=ThresholdType.OUTER_RANGE,
+                level=threshold_level,
+                must_be_greater_than=threshold_yaml.must_be_not_between.greater_than,
+                must_be_greater_than_or_equal=threshold_yaml.must_be_not_between.greater_than_or_equal,
+                must_be_less_than=threshold_yaml.must_be_not_between.less_than,
+                must_be_less_than_or_equal=threshold_yaml.must_be_not_between.less_than_or_equal,
+            )
+
+        return None
```
Niels-b, line 976: **This `return None` is a check that silently never runs.**

A flat threshold with two comparisons — a natural way to write a range:

```yaml
threshold:
  must_be_greater_than: 0
  must_be_less_than: 100
```

lands here: `has_exactly_one_comparison()` is False, neither `must_be_between` nor `must_be_not_between` is set, so we return `None` with no log line at all. Downstream: `self.threshold` is `None` → `evaluate_threshold` returns `NOT_EVALUATED` → `_get_contract_verification_status` returns `UNKNOWN` (not FAILED, not WARNED) → `session_result_to_exit_code` returns `ExitCode.OK`. Green scan, exit 0, check never ran.

The behaviour is pre-existing — the function used to fall off the end — so this isn't a regression. But this PR is where leaving it gets odd: `contract_yaml.py:815` documents the exact hazard ("the check stays NOT_EVALUATED forever") and guards it, *only* when `additional:` is present. And the last commit removed the bundled schema's flat-threshold arity rule, so nothing static catches it any more either.

A `logger.error` here would make the new `additional` arity rule a special case of a general rule instead of a one-off, and would make the schema loosening free rather than a net loss of coverage. Something like: `Threshold has more than one comparison: <keys>. Specify exactly one.`

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -850,15 +852,32 @@ class ThresholdLevel(Enum):
...
+
+    Returns (primary_threshold_yaml, primary_level, other_threshold_yaml). The
+    FAIL-level threshold takes the primary slot — that is the outer one unless the
+    outer one is the warn threshold and the additional one is the fail threshold.
+    Contract parsing rejects two thresholds with the same effective level, so exactly
+    one of them is the fail threshold.
+    """
+    outer_level: ThresholdLevel = ThresholdLevel.from_str(threshold_yaml.level)
+    additional_level: ThresholdLevel = ThresholdLevel.from_str(threshold_yaml.additional.level)
+    if outer_level == ThresholdLevel.WARN and additional_level == ThresholdLevel.FAIL:
```
Niels-b, line 879: **The reversed orientation is a fail-open on a pre-#2818 engine.**

Supporting either orientation is nice for authoring, but the reversed one has a sharp edge on an engine that predates this PR:

```yaml
threshold:
  level: warn
  must_be_less_than: 5
  additional:
    must_be_less_than: 10
```

An older engine reads `level` (it always has — `read_string_opt("level", default_value="fail")` is on main), ignores `additional` entirely (no unknown-key rejection anywhere in `ThresholdYaml`/`YamlObject`, and `json_schema_verifier.py` is commented out end to end), and then evaluates the *tighter warn* bound at WARN level. The check can never produce FAILED — fail-blocking gates and alerts stop firing while it still looks healthy.

The migrator can't emit this shape (it always puts the fail condition on the outer threshold), so migrated contracts are safe. But the Cloud editor can, and as far as I can tell `changeThresholdLevel` isn't behind `multipleCheckThresholdsEnabled` — only add/remove row are — so even a flag-off tenant can flip an existing two-level check into it. CLI/API publishes are ungated from the moment the Cloud schema lands.

Suggestion: log a notice at parse time when the outer threshold is the warn one (probably best placed next to the existing `warn_can_fire_alone` lint in `CheckImpl.parse_check`). It's the only mitigation that reaches **Cloud-authored** contracts — the migrator's runner-version remark structurally can't.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -1249,6 +1346,7 @@ def _build_check_info(self) -> Check:
             contract_file_line=self.check_yaml.check_yaml_object.location.line,
             contract_file_column=self.check_yaml.check_yaml_object.location.column,
             threshold=self._build_threshold(),
+            warn_threshold=self._build_warn_threshold(),
```
Niels-b, line 1399: **Nothing tests this line, and dropping it is silent.**

This kwarg is the only thing carrying the warn threshold out of the engine, and `Check.warn_threshold` defaults to `None` — so deleting it isn't even a `TypeError`. I removed it and the whole new suite stays green: all 63 new core tests here, and all 197 in soda-extensions#547. Cloud loses `diagnostics.warn` for every two-level check and the CLI drops the second threshold row — the entire user-visible payoff — with no failing test.

The cause is that the impl-level tests assert `check.warn_threshold` on the `CheckImpl` (a `ThresholdImpl`), and the wire tests hand-build the `Check` dataclass, so nothing crosses this seam.

Worth saying that slot assignment itself is well defended — I tried inverting the condition in `split_threshold_yamls_by_level`, deleting the swap entirely, and diverging the two call sites, and each was caught at multiple altitudes. It's specifically the exit from the engine that's undefended.

One end-to-end test through `MockSodaCloud` that parses a contract with `additional:` and asserts `diagnostics["warn"]` closes it — the harness already exists in `test_soda_cloud.py`.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
```diff
@@ -711,8 +711,67 @@ def get_valid_units(self) -> list[str]:
...
+_OUTER_ARITY_MESSAGE: str = (
+    "It must specify exactly one comparison itself "
+    "(one must_be_* key, or one must_be_between/must_be_not_between range)"
+)
+
+
 class ThresholdYaml:
-    def __init__(self, threshold_yaml_object: YamlObject):
+    def __init__(self, threshold_yaml_object: YamlObject, is_additional: bool = False):
+        if is_additional:
```
Niels-b, line 775: **Strict key checking inside `additional`, none on the threshold holding it.**

A one-letter typo silently produces a fail-only check:

```yaml
threshold:
  must_be_less_than: 10
  aditional: { must_be_less_than: 5, level: warn }
```

The outer object is read key-by-key (`read_number_opt`, `read_object_opt("additional")`) with no allowed-key set, and `YamlObject` has no generic unknown-key check — so `aditional` is dropped with no error and no warning, and the check verifies as fail-only. That's precisely the outcome the feature exists to prevent. Meanwhile `must_be_les_than` *inside* `additional` is a hard error, so the strictness is inverted relative to where the risk is.

The outer key set is now fully enumerated (`THRESHOLD_COMPARISON_KEYS | {level, additional, metric, unit}`), so running the same loop over the outer object is nearly free. If that's too broad a behaviour change to fold in here, even a targeted did-you-mean on `additional` would cover the new key.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -873,6 +892,13 @@ def create(
                 logger.error(f"Threshold required, but not specified")
                 return None
 
+        if threshold_yaml.additional is not None:
```
Niels-b, line 896: **`additional` can't be combined with a check type's default threshold.**

This early return happens before the `default_threshold` handling below, so the natural way to write "fail on any missing value (the default), warn at 5" is rejected:

```yaml
- missing:
    threshold:
      additional: { must_be_less_than: 5, level: warn }
```

→ *"A threshold with an 'additional' threshold must specify a comparison itself"*. The user has to rediscover and restate each check type's default (`must_be: 0` for missing/invalid/duplicate/failed_rows, `must_be_greater_than: 0` for row_count).

What makes it surprising is the asymmetry: the **flat** path already supports default-comparison-plus-explicit-level — `threshold: {level: warn}` on a `missing` check works, because `create` applies `default_threshold.level = threshold_level` and returns the default. Adding a second level is what takes the convenience away.

`test_additional_ignores_default_threshold` pins this as intended, so I may be missing a reason — but nothing in the design notes seems to require it, and `split_threshold_yamls_by_level` could fall back to `default_threshold` for whichever side has no comparison of its own. If it stays as-is, the error message should name the default to write ("this check type's default threshold is `must_be: 0`; state it explicitly").

Niels-b (dismissed):
Claude still picked up on some things I think are worth discussion.

Also, do we only want this for contract verification or also for other things?

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Warning, not error — f9a17ad7. An error here flips `has_errors` and the whole verification goes ERRORED — and the Cloud schema still accepts this shape at publish time (arity deliberately scoped to `additional`-present, master parity), so published contracts that verify green today would break on engine upgrade. The warning names the comparisons and says the check is not evaluated. Tightening schema + engine in lockstep, with an announcement, stays in DET-48.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Added in f9a17ad7 — a parse-time warning next to the dead-warn lint, firing when the outer threshold is the warn one: a pre-`additional` engine evaluates only the outer warn threshold, so the check cannot fail there; put the fail comparison on the outer threshold if the contract may run on an older runner. Docs carry the same caveat. It does fire on every scan of a warn-outer contract, including where it's safe — acceptable during the transition, removable once pre-#2818 runners are gone.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Added `test_soda_cloud_results_with_additional_threshold` (f9a17ad7): a full verification through MockSodaCloud asserting the upload payload carries `outcome: warn` plus both `diagnostics.fail` and `diagnostics.warn`. Deleting the `warn_threshold=` kwarg now fails it.

soda-core/src/soda_core/contracts/impl/contract_yaml.py
m1n0: Added the loop over the outer keys (f9a17ad7) — as a warning, not an error: thresholds never had unknown-key validation, so stray keys in published contracts have to keep verifying. `aditional:` now logs "'aditional' is not a known threshold key and is ignored". Allowed set is the comparison keys + level/additional/metric/unit — metric and unit are read off the same object by ThresholdCheckYaml, and whether they're allowed is per check type and validated there.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Keeping the rejection: the Cloud contract and data-standard schemas (merged in soda#13165) require the outer comparison when `additional:` is present, so an engine-side default fallback would produce contracts that verify locally but cannot publish. The error now names the default though (f9a17ad7): "A check type's default threshold does not combine with an 'additional' threshold. State this check type's default explicitly: must_be: 0". If the fallback is wanted later, both Cloud schemas and the engine have to move together — added to DET-48.

m1n0:
On "only contract verification or also other things": this lands wherever the shared threshold code runs. Data standards use the same parser/impl — their Cloud schema is updated in soda#13165 and the editor/results treat them the same. Extension check types that route through `evaluate_threshold` inherit it: group_by gets explicit per-group warn evaluation in soda-extensions#547, reconciliation and the rest go through `evaluate_threshold` unchanged. Not covered: v3 SodaCL (has its own warn/fail config) and metric monitoring (its own alerting model).

Round of fixes for the inline comments pushed as f9a17ad7 — replies in the threads. Unit suite 1262 passed, 2 skipped.

Niels-b (approved):
Thanks for fixing the issues 👍 .

# AI written feedback:

Review round addressed everything I raised — and each fix came with a negative control (`test_fail_outer_orientation_logs_no_older_runner_notice`, `test_metric_unit_level_and_additional_are_known_outer_keys`), which is the half that usually gets skipped when you add log-emitting paths.

Verified rather than taken on trust:

- `test_soda_cloud_results_with_additional_threshold` runs a real contract through the engine and asserts the upload payload carries both `fail: {greaterThan: 5}` and `warn: {greaterThan: 1}`. Dropping the `warn_threshold=` kwarg now fails a test, which was the one seam with no coverage.
- The reversed-orientation notice sits inside the `threshold and warn_threshold` guard, so it can't misfire on a plain flat `level: warn` check.
- The new default-threshold error is gated on `has_any_configurations()`, which counts comparisons only and not `additional` — so it actually fires for the "default + additional only" shape rather than being dead code.

One deviation I'm happy to accept: I asked for `logger.error` on the no-threshold path and you chose `logger.warning`, on the grounds that an error would flip contracts that verify green today to ERRORED on upgrade while the Cloud schema still accepts them at publish. That's the right call. Worth being explicit about the residual, since it isn't obvious from the diff: `has_errors` keys on ERROR only, so such a check still reports `UNKNOWN` and the CLI still exits 0 — the shape is now diagnosable in the logs, but a green scan can still contain a check that never evaluated. Fine as a deliberate step given the lockstep-with-Cloud constraint is tracked.

Approving. The interval arithmetic behind `warn_can_fire_alone` is the part I'd have expected to be subtly wrong, and it holds up under brute force — nice work.

PR #2821, opened 2026-08-10
fix: relax ruamel.yaml version constraint to support Airflow 3.2.1 (f…

### LINKED ISSUE #2746: soda-core 4.7.0 incompatible with Airflow 3.2.1 constraints due to ruamel.yaml version conflict
We were trying to install soda-snowflake==4.7.0 in an MWAA airflow v3.2.1 environment using the official Apache Airflow 3.2.1 [constraints file](https://raw.githubusercontent.com/apache/airflow/constraints-3.2.1/constraints-3.12.txt) (Python 3.12):

The installation fails due to a dependency conflict between Airflow constraints and Soda Core requirements.

```
soda-core 4.7.0 requires:
  ruamel.yaml >=0.17.0, <0.18.0

Airflow 3.2.1 constraints require:
  ruamel.yaml ==0.19.1
```

This prevents running Soda data quality checks directly inside Airflow environments using official constraints,

Are there any planned updates to soda-core to relax the ruamel.yaml constraint to support newer versions (e.g., >=0.19.x) used in Airflow 3.2.1+ environments?

Our current workaround is going to be having Airflow call a Lambda or ECS to verify the contracts instead of doing so in Airflow.

Thanigaivendhan:
## Summary

Fixes #2746 — relaxes the `ruamel.yaml` version constraint in `soda-core/pyproject.toml` from `>=0.17.0,<0.18.0` to `>=0.17.0,<0.20.0` to resolve a dependency conflict with Apache Airflow 3.2.1, which requires `ruamel.yaml==0.19.1`.

## Verification

Before making this change, I checked whether soda-core's actual usage of ruamel.yaml would be affected by the newer version:

- Identified all usages of `ruamel.yaml` in the codebase (`soda_core/common/yaml.py` and `soda_core/contracts/impl/contract_verification_impl.py`) — usage is limited to `YAML`, `CommentedMap`, `CommentedSeq`, `MarkedYAMLError`, `.load()`, `.dump()`, `preserve_quotes`, and `.lc.value()` / `.lc.item()` for line/column tracking.
- Confirmed all of these APIs behave identically on `ruamel.yaml==0.19.1`.
- Ran the existing YAML unit test suite (`soda-tests/tests/unit/test_yaml_api.py` and related tests) against `ruamel.yaml==0.19.1`: 121 tests passed, 0 regressions related to this change.

## Why `<0.20.0` rather than unpinning entirely

Kept an upper bound rather than removing the constraint entirely, since ruamel.yaml has had breaking changes across major versions historically  this keeps the same conservative approach as the original pin while unblocking the immediate Airflow compatibility issue.

Happy to adjust the range further if maintainers have a different preference.

### DIFF
--- a/soda-core/pyproject.toml
+++ b/soda-core/pyproject.toml
@@ -8,7 +8,7 @@ authors = [
     {name = "Soda Data N.V.", email = "info@soda.io"}
 ]
 dependencies = [
-    "ruamel.yaml>=0.17.0,<0.18.0",
+    "ruamel.yaml>=0.17.0,<0.20.0",
     "requests>=2.32.3,<2.34.0",
     "pydantic>=2.11,<3.0",
     "opentelemetry-api>=1.16.0,<2.0.0",
--- a/uv.lock
+++ b/uv.lock
@@ -506,7 +506,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
+    { name = "typing-extensions" },
 ]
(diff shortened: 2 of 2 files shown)


mivds:
Hi @Thanigaivendhan, thanks for preparing a fix for this issue!

Could you please accept the license agreement? Otherwise we won't be able to accept your change

PR #2826, opened 2026-08-17
fix(databricks): refuse to connect without credentials instead of hanging in interactive OAuth

m1n0:
## What

Databricks: refuse to connect when there are no credentials, instead of letting `databricks-sql-connector` fall back to its **interactive browser OAuth (U2M)** flow.

## Why

An empty `access_token` — typically `access_token: ${env.VAR}` with `VAR` unset or empty where the verification runs — reached `databricks.sql.connect(access_token="")`. The connector treats a falsy token as "no auth configured" and starts its U2M flow: it prints

```
Opening https://<workspace>/oidc/oauth2/v2.0/authorize?...client_id=databricks-sql-python&redirect_uri=http://localhost:8020...
Listening for OAuth authorization callback at http://localhost:8020
```

and blocks forever waiting for a browser callback — an infinite hang in any headless run (CI, agent).

Three parts:

1. `DatabricksTokenAuth.access_token` gets `min_length=1`: an empty token is rejected by the model at parse time (a missing/`None` one already was; an empty one produced a model with an empty secret).
2. `DatabricksDataSourceConnection._create_connection`: never calls `sql.connect()` without a `credentials_provider` or a non-empty `access_token`. This is the guard that ends the hang class regardless of how the token went missing.
3. (review follow-up, @mivds) the `connection` block is now a pydantic discriminated union keyed on `auth_type`, and the hand-written `infer_connection_type` validator plus its dispatch table are gone:

```python
DatabricksConnection = Annotated[
    Union[
        Annotated[DatabricksTokenAuth, Tag("personal-access-token")],
        Annotated[DatabricksOAuthM2M, Tag("databricks-oauth-m2m")],
        Annotated[DatabricksAzureServicePrincipal, Tag("azure-service-principal")],
    ],
    Discriminator(_discriminate_auth_type),   # omitted auth_type -> PAT (backward compatible)
]
```

   `DatabricksTokenAuth.auth_type` is `Optional[Literal["personal-access-token"]] = None`, the OAuth members require their Literal. A *callable* discriminator (rather than `Field(discriminator="auth_type")`, which needs a required Literal on every member, or a plain `Union`, which reports one bad input against all three members) keeps errors scoped to the selected mode:

```
connection.personal-access-token.access_token
  Field required                                      # no auth_type, no token
connection.personal-access-token.access_token
  Value should have at least 1 item ...               # empty token
connection.databricks-oauth-m2m.client_secret
  Field required
connection
  Input tag 'nope' ... does not match any of the expected tags: 'personal-access-token', 'databricks-oauth-m2m', 'azure-service-principal'
```

   `auth_type` is stripped in `DatabricksSharedConnectionProperties.to_connection_kwargs()` so the discriminator never reaches `sql.connect(**kwargs)` — the connector has its own, differently-valued `auth_type` argument (`databricks-oauth` selects the browser flow); regression test added. Same pattern as Trino's `connection_properties` union, plus the callable discriminator for the error scoping.

## Tests

`soda-databricks/tests/unit/` — 42 passing, no live database (`databricks.sql.connect` mocked): empty/None token rejected at parse time (with and without explicit `auth_type`), connection guard for empty/None, a control that a real token still connects, `auth_type` never in connect kwargs. In `test_databricks_oauth.py` the two message assertions for "unknown auth_type" / "no auth_type, no token" follow the pydantic wording and the `infer` helper validates through `DatabricksDataSource` like a YAML load does.

```customer-facing-release-note
Databricks contract verifications with a missing or empty access token now fail fast.
```

### DIFF
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
@@ -50,8 +50,7 @@ class DatabricksDataSourceConnection(DataSourceConnection):
             raise ValueError(
                 "Databricks connection has no credentials: 'access_token' is missing or empty and no OAuth "
                 "auth_type is configured. Refusing to connect, because the Databricks SQL connector would "
-                "otherwise start an interactive browser login that cannot complete in a non-interactive run. "
-                "If the token is provided through ${env.VAR}, make sure VAR is set where the verification runs."
+                "otherwise start an interactive browser login that cannot complete in a non-interactive run."
             )
         return sql.connect(
             user_agent_entry="Soda Core",
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
@@ -42,7 +42,17 @@ class DatabricksDataSourceConnection(DataSourceConnection):
                 **connection_kwargs,
             )
 
-        # Token (PAT) auth: access_token flows through connection_kwargs unchanged.
+        # Token (PAT) auth: access_token flows through connection_kwargs unchanged. Never hand
+        # the connector a credential-less connect(): with no credentials_provider and a falsy
+        # access_token it falls back to its interactive browser OAuth (U2M) flow, which opens a
+        # browser URL and blocks on localhost:8020 — an infinite hang in a headless run.
+        if not connection_kwargs.get("access_token"):
+            raise ValueError(
+                "Databricks connection has no credentials: 'access_token' is missing or empty and no OAuth "
+                "auth_type is configured. Refusing to connect, because the Databricks SQL connector would "
+                "otherwise start an interactive browser login that cannot complete in a non-interactive run. "
+                "If the token is provided through ${env.VAR}, make sure VAR is set where the verification runs."
+            )
         return sql.connect(
             user_agent_entry="Soda Core",
             **connection_kwargs,
--- a/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
+++ b/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
@@ -38,7 +38,10 @@ class DatabricksSharedConnectionProperties(DatabricksConnectionProperties, ABC):
 
 
 class DatabricksTokenAuth(DatabricksSharedConnectionProperties):
-    access_token: SecretStr = Field(..., description="Personal access token")
+    # min_length: an empty token (e.g. an unresolved variable reference) must be rejected here —
+    # the connector treats a falsy access_token as "no auth" and falls back to an interactive
+    # browser login (see DatabricksDataSourceConnection._create_connection).
+    access_token: SecretStr = Field(..., min_length=1, description="Personal access token")
 
 
 class DatabricksOAuthM2M(DatabricksSharedConnectionProperties):
--- a/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
+++ b/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
@@ -1,10 +1,16 @@
 from abc import ABC
-from typing import ClassVar, Dict, Optional
+from typing import Annotated, Any, ClassVar, Dict, Literal, Optional, Union
 
-from pydantic import Field, SecretStr
+from pydantic import Discriminator, Field, SecretStr, Tag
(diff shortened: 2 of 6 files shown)


soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
```diff
@@ -42,7 +42,17 @@ def _create_connection(
...
+        # Token (PAT) auth: access_token flows through connection_kwargs unchanged. Never hand
+        # the connector a credential-less connect(): with no credentials_provider and a falsy
+        # access_token it falls back to its interactive browser OAuth (U2M) flow, which opens a
+        # browser URL and blocks on localhost:8020 — an infinite hang in a headless run.
+        if not connection_kwargs.get("access_token"):
+            raise ValueError(
+                "Databricks connection has no credentials: 'access_token' is missing or empty and no OAuth "
+                "auth_type is configured. Refusing to connect, because the Databricks SQL connector would "
+                "otherwise start an interactive browser login that cannot complete in a non-interactive run. "
+                "If the token is provided through ${env.VAR}, make sure VAR is set where the verification runs."
```
mivds: Would probably drop this last line. Especially with this engine there are multiple channels to provide credentials, so I wouldn't call out `env` vars specifically

mivds (commented):
This works, but I would be inclined to use pydantic validation to solve this. Right now we're manually validating the input, which forces us to also do partial validation on connecting.

We could instead do something like this:
```python
class DatabricksTokenAuth(DatabricksSharedConnectionProperties):
    auth_type: Literal["personal-access-token"] | None = None
    access_token: SecretStr = Field(..., description="Personal access token", min_length=1)

class DatabricksOAuthM2M(DatabricksSharedConnectionProperties):
    auth_type: Literal["databricks-oauth-m2m"]
    client_id: str = Field(..., description="Databricks OAuth service-principal client ID")
    client_secret: SecretStr = Field(..., description="Databricks OAuth service-principal client secret")

class DatabricksAzureServicePrincipal(DatabricksSharedConnectionProperties):
    auth_type: Literal["azure-service-principal"]
    azure_client_id: str = Field(..., description="Entra ID (Azure AD) service-principal application/client ID")
    azure_client_secret: SecretStr = Field(..., description="Entra ID (Azure AD) service-principal client secret")
    azure_tenant_id: str = Field(..., description="Entra ID (Azure AD) directory/tenant ID")

DatabricksConnectionProperties = DatabricksTokenAuth | DatabricksOAuthM2M | DatabricksAzureServicePrincipal
```

This would handle all inference / decoding at the pydantic level. `auth_type` discriminates what authentication mechanism the user wants to use (same as it does today), and the `| None = None` fallback in `DatabricksTokenAuth` makes it backwards compatible.

The one thing I'm not sure about is how readable the output is if pydantic rejects the input

mivds (approved):
Pre-approving, I'll leave it up to you which direction you want to go

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source_connection.py
m1n0: Dropped in c3750266.

m1n0:
Went halfway on the pydantic route in c3750266: `access_token` has `min_length=1` on `DatabricksTokenAuth` and the manual check in `infer_connection_type` is gone. The connect-time guard stays — it's not input validation but the invariant that we never hand the connector a credential-less connect, whatever path the config took.

Held off on the union: a discriminated union needs the `auth_type` Literal on every member, so with `| None` on the PAT one pydantic falls back to a smart-mode union and reports a bad input against all three members — the readability issue you flagged. It'd also put `auth_type` into `model_dump()` and thus into `sql.connect(**kwargs)` unless stripped, and the current dispatch keys the same literals soda-library and the Cloud form use. Fine to do it as a follow-up if you feel strongly.

m1n0:
Correction to the above — the plain `Union` does work (Trino's `connection_properties` is exactly that); what I was pushing back on was the error output, and I put it badly. With `| None` on the PAT member pydantic can't use `Field(discriminator=...)`, and a plain `Union` reports one bad input against all three members (an empty token → 8 errors, 7 of them "Field required" for the OAuth modes). A callable `Discriminator` gives the same model without that: one error, scoped to the selected mode. Did it as #2830 (stacked on this one) so this PR can go into today's release as-is.

m1n0:
Folded the discriminated-union change into this PR (7303cab4) instead of the separate #2830 — one commit on top, `infer_connection_type` gone, errors scoped per mode, `auth_type` stripped from the connect kwargs. PR description updated.

PR #2827, opened 2026-08-17
fix(sparkdf): qualify SHOW TABLES with the catalog in HiveMetadataTablesQuery

Niels-b:
## Description

`HiveMetadataTablesQuery.build_sql_statement` accepts `database_name` and never uses it, emitting
`SHOW TABLES FROM <schema>`. An unqualified schema resolves against the **session's current
catalog**, not the catalog carried in the prefixes.

For the Databricks SQL connector that is harmless: the class is gated behind `_is_hive_catalog()`,
so it only runs when the connection catalog really is `hive_metastore`, and that catalog is passed
to `sql.connect()` so the session default matches.

`SparkDataFrameMetadataTablesQuery` inherits the method and, unlike the connector, is returned
**unconditionally** — no gate. Nothing in sparkdf ever issues `USE CATALOG`, so in catalog mode a
DWH prefix `[catalog, soda_diagnostics]` loses its catalog and the lookup lands in whatever
`current_catalog()` happens to be — commonly `hive_metastore` on clusters with no workspace default
catalog, regardless of which catalog the data actually lives in.

Because DWH writes are already fully qualified (`qualify_dataset_name` →
`` `catalog`.`soda_diagnostics`.`fr_…` ``), the read and write paths disagree within a single run:

1. `SHOW TABLES FROM \`soda_diagnostics\`` misses — or throws `AnalysisException` when that schema
   does not exist in the session catalog, which sparkdf's `execute()` swallows into an empty result
2. `DwhMetadataCache.has_table()` reports False for a table that exists
3. `failed_rows_check_extension.py:654` takes the create branch → `CREATE TABLE … AS SELECT`, which
   has no `IF NOT EXISTS` (`sql_dialect.py:612`) → `TABLE_OR_VIEW_ALREADY_EXISTS`

Symptom: the first run is clean — nothing exists yet, the CTAS succeeds, and `mark_table_exists`
masks the bad lookup for the remainder of that run — and every subsequent run fails. The same lookup
also feeds `rows_diff_check_extension.py:125` and `create_view_from_table`, which silently skips view
creation.

Present since sparkdf catalog mode landed (#2723 / #2730 / #2740).

**The fix:** qualify the `FROM` with `database_name` when it is set, keeping the unqualified form
when it is not. The guard is load-bearing — sparkdf legacy mode has
`get_database_prefix_index() == None`, so `database_name` is always `None` there and must keep
emitting the bare schema.

**Why this wasn't caught:** every sparkdf DWH test runs on local Spark where `spark_catalog` is the
only catalog, so the session catalog always equals the target catalog. The Databricks workflow
exercises `DATABRICKS_CATALOG=hive_metastore` and UC separately, never mixed.

**Any expected impact on downstream packages/services?**

One behaviour change for existing users: the Databricks hive branch now emits
`` SHOW TABLES FROM `hive_metastore`.`schema` `` instead of `` SHOW TABLES FROM `schema` ``.
Semantically identical — the session catalog is already `hive_metastore` — but it is a new statement
shape against a real metastore, so **the hive CI lane is the gate for this PR**
(`.github/workflows/main.workflow.yaml:291-295`).

Deliberately **not** in scope:

- `get_results` still hardcodes `database_name="hive_metastore"` (`:105,111`). Correct everywhere it
  currently runs — sparkdf overrides `get_results` to emit `None` — and fixing it means a signature
  change plus DQN blast radius via `DatasetIdentifier` / `build_dwh_qualified_table_names`.
- The per-connection `_is_hive_catalog()` gate is wrong for **mixed hive+UC workspaces**: a
  `hive_metastore` connection reading UC dataset prefixes drops the catalog, and a UC connection
  reading `hive_metastore` prefixes queries a `hive_metastore.information_schema` that does not
  exist. Separate defect, tracked separately.
- sparkdf's blanket `except Exception: pass` in `execute()` is why this degraded silently rather than
  erroring. Changing it would convert previously-silent misses into new failures elsewhere.

## Checklist

## Verification

- **The new tests fail without the fix** — ran them against the unfixed source: the 3
  catalog-qualification assertions fail, the 4 unchanged-behaviour guards pass.
- **No regressions** across all 13 unit suites: `1437 passed / 16 errors` before → `1444 passed /
  16 errors` after. Identical 16 pre-existing `data_source_test_helper_session` fixture errors
  (they need a live datasource); `+7` is exactly the new tests.
- `black --check` clean at line-length 120.
- **Not verified locally:** `SHOW TABLES FROM cat.schema` under Spark Connect / Databricks
  serverless. Worth a live check before undrafting.

### DIFF
--- a/soda-databricks/src/soda_databricks/common/statements/hive_metadata_tables_query.py
+++ b/soda-databricks/src/soda_databricks/common/statements/hive_metadata_tables_query.py
@@ -73,13 +73,16 @@ class HiveMetadataTablesQuery(MetadataTablesQuery):
         exclude_table_name_like_filters: Optional[list[str]] = None,
         object_type_to_fetch: TableType = TableType.TABLE,
     ) -> str:  # Return type for this function is a string, not a list (for the super method it is a list!)
-        schema_str = ""
+        from_str = ""
         if schema_name:
-            schema_str = f" FROM {self.sql_dialect.quote_default(schema_name)}"
+            # An unqualified schema resolves against the session's current catalog, which is not
+            # necessarily the catalog in the prefixes — sparkdf never issues USE CATALOG.
+            name_parts = [database_name, schema_name] if database_name else [schema_name]
+            from_str = " FROM " + ".".join(self.sql_dialect.quote_default(part) for part in name_parts)
         if object_type_to_fetch == TableType.TABLE:
-            return f"SHOW TABLES{schema_str}"
+            return f"SHOW TABLES{from_str}"
         elif object_type_to_fetch == TableType.VIEW:
-            return f"SHOW VIEWS{schema_str}"
+            return f"SHOW VIEWS{from_str}"
         else:
             raise ValueError(f"Invalid object type to fetch: {object_type_to_fetch}")
 
--- a/soda-databricks/tests/unit/test_hive_metadata_tables_query.py
+++ b/soda-databricks/tests/unit/test_hive_metadata_tables_query.py
@@ -66,8 +66,9 @@ def test_build_sql_statement(database_name, schema_name, object_type, expected_s
 
 
 def test_invalid_object_type_raises():
+    query = _query()
     with pytest.raises(ValueError, match="Invalid object type to fetch"):
-        _query().build_sql_statement(
+        query.build_sql_statement(
             database_name="my_uc_catalog",
             schema_name="soda_diagnostics",
             object_type_to_fetch=TableType.MATERIALIZED_VIEW,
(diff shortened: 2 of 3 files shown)


mivds (approved):
Fix looks fine, but looking at the code I do have 2 larger questions:

- This is a discovery query with optional `schema_name`. What happens when `schema_name = None`? I think spark will use the active schema, same way as it uses the active catalog when not specified.
- Since this is unity catalog, can we use `information_schema` instead of the `SHOW ...` queries? I.e. the same path that databricks uses. I'm not sure if that works on spark or not, so genuine question.

Note these questions may be moot. I'm not sure how spark integrates with schemas, but given the catalog bug we just encountered there might be more here...

Niels-b:
Thanks for the review @mivds!

1. The code now and before will do the same thing: spark will fall back to the current sessions schema. If we want to list it for all schemas, we'd have to run it in a loop (which is different than we have for most other datasources). This is fine though, as SparkDF is not used for discovery (it's only programatic).
2. Depends on the use-case. It may also depend on the version. I would say that is OOS for this PR, as this just fixes the issue. Adding full unity-catalog support for spark seems more like a feature request IMO. It should be possible, since we indeed use the same thing for Databricks. It's technically possible, but quick analysis says that this is quite a lot of work.

So questions are definitely not moot 😉

PR #2828, opened 2026-08-18
feat(cloud): add agent source to scan token usage

nikosml:
## Description

Scan token usage can now carry optional source attribution through the existing scan-results payload.

- add a trailing optional `agent_source` field to `ScanTokenUsage`
- serialize it as optional `agentSource`, omitting the key when unset
- consolidate token-usage dictionary construction while preserving result and entry order
- preserve the existing empty `tokenUsage` list behavior

Downstream packages that populate `agent_source` must require this or a later soda-core build. Existing callers remain compatible because the field is trailing and defaulted.

## Coordination

- Linear: [PLATL-970](https://linear.app/sodadata/issue/PLATL-970/thread-contract-autopilot-agentsource-into-scan-token-usage). Siblings: [soda#13146](https://github.com/sodadata/soda/pull/13146), [soda-core#2828](https://github.com/sodadata/soda-core/pull/2828), [soda-extensions#564](https://github.com/sodadata/soda-extensions/pull/564).
- Order: soda is independent; soda-core precedes soda-extensions, which must pin its published dev artifact.

## Checklist

### Customer-facing release note

```customer-facing-release-note
NONE
```

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1774,7 +1774,7 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
                 "totalTokens": tu.total_tokens,
                 "model": tu.model,
                 "operation": tu.operation,
-                "agentSource": tu.agent_source,
+                **({"agentSource": tu.agent_source} if tu.agent_source is not None else {}),
             }
             for tu in contract_verification_result.token_usage
         ]
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1774,6 +1774,7 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
                 "totalTokens": tu.total_tokens,
                 "model": tu.model,
                 "operation": tu.operation,
+                "agentSource": tu.agent_source,
             }
             for tu in contract_verification_result.token_usage
         ]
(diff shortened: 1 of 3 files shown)


soda-core/src/soda_core/contracts/contract_verification.py
```diff
@@ -558,6 +558,7 @@ class ScanTokenUsage:
     total_tokens: int
     model: Optional[str] = None
     operation: Optional[str] = None  # "autopilot" or "llmCheck" (matches server Gson @SerializedName)
+    agent_source: Optional[str] = None
```
m1n0, line 562: Document the allowed values like `operation` does — `"SODA"` / `"BYOK"`, case-sensitive; Cloud's Gson maps anything else to null without an error.

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1774,6 +1774,7 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
                 "totalTokens": tu.total_tokens,
                 "model": tu.model,
                 "operation": tu.operation,
+                "agentSource": tu.agent_source,
```
m1n0: This emits `agentSource: None`; the key only disappears because `to_jsonnable` strips nulls at :1884. Either say that in the description instead of "omitting when unset", or add `if tu.agent_source is not None`. As is, `test_build_token_usage_dicts_serialization` and the payload test look contradictory to the next reader.

m1n0 (approved):
Ran the unit tests locally since the fork PR can't fetch CI secrets — 4/4 green, and the 5 errors in the rest of `test_soda_cloud.py` are the same postgres-connection ones main has. Two nits inline.

soda-core/src/soda_core/common/soda_cloud.py
nikosml: Addressed in 67c330fa: `_build_token_usage_dicts` now omits `agentSource` when unset, and the helper-level test expectation was updated so it agrees with the final payload behavior.

soda-core/src/soda_core/contracts/contract_verification.py
nikosml: Addressed in 67c330fa: documented the case-sensitive `"SODA"` / `"BYOK"` values and that Cloud Gson maps any other value to null.

nikosml:
Superseded by #2835, which contains the same commits on a branch within sodadata/soda-core so the full trusted CI pipeline can run and the change can be merged upon acceptance.

PR #2830, opened 2026-08-19
refactor(databricks): discriminated union for the auth mode instead of a manual validator

m1n0:
## What

Follow-up to #2826, picking up @mivds's review suggestion: the Databricks `connection` block is now a pydantic discriminated union keyed on `auth_type`, and the hand-written `infer_connection_type` validator plus its dispatch table are gone.

```python
DatabricksConnection = Annotated[
    Union[
        Annotated[DatabricksTokenAuth, Tag("personal-access-token")],
        Annotated[DatabricksOAuthM2M, Tag("databricks-oauth-m2m")],
        Annotated[DatabricksAzureServicePrincipal, Tag("azure-service-principal")],
    ],
    Discriminator(_discriminate_auth_type),   # omitted auth_type -> PAT (backward compatible)
]
```

`DatabricksTokenAuth.auth_type` is `Optional[Literal["personal-access-token"]] = None`, the OAuth members require their Literal. A *callable* discriminator (rather than `Field(discriminator="auth_type")`, which needs a required Literal on every member, or a plain `Union`, which reports one bad input against all three members) keeps errors scoped to the selected mode:

```
connection.personal-access-token.access_token
  Field required                                      # no auth_type, no token
connection.personal-access-token.access_token
  Value should have at least 1 item ...               # empty token (min_length=1 from #2826)
connection.databricks-oauth-m2m.client_secret
  Field required
connection
  Input tag 'nope' ... does not match any of the expected tags: 'personal-access-token', 'databricks-oauth-m2m', 'azure-service-principal'
```

`auth_type` is stripped in `DatabricksSharedConnectionProperties.to_connection_kwargs()` so the discriminator never reaches `sql.connect(**kwargs)` — the connector has its own, differently-valued `auth_type` argument (`databricks-oauth` selects the browser flow). Regression test added for that.

Same pattern as Trino's `TrinoDataSource.connection_properties` union, with the callable discriminator on top for the error scoping.

## Tests

`soda-databricks/tests/unit/` — 42 passing, no live database. The two message assertions for "unknown auth_type" / "no auth_type, no token" updated to the pydantic wording; `infer` helper now validates through `DatabricksDataSource` like a YAML load does.

```customer-facing-release-note
NONE
```

### DIFF
--- a/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
+++ b/soda-databricks/src/soda_databricks/model/data_source/databricks_connection_properties.py
@@ -1,10 +1,16 @@
 from abc import ABC
-from typing import ClassVar, Dict, Optional
+from typing import Annotated, Any, ClassVar, Dict, Literal, Optional, Union
 
-from pydantic import Field, SecretStr
+from pydantic import Discriminator, Field, SecretStr, Tag
 from soda_core.model.data_source.data_source_connection_properties import (
     DataSourceConnectionProperties,
 )
+
+# Explicit auth_type discriminator values (Trino/Snowflake/BigQuery style). These literals
+# are the cross-repo contract: they match soda-library#759 and the Cloud connection form.
+AUTH_TYPE_TOKEN = "personal-access-token"
+AUTH_TYPE_OAUTH_M2M = "databricks-oauth-m2m"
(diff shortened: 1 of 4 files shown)


m1n0:
Folding this into #2826 instead.

PR #2834, opened 2026-08-20
fix(sqlserver): retry deadlock victims with jittered exponential backoff

mivds:
Re-introduces #2803, which was reverted in #2806 because it surfaced a bug in secret masking inside exception handlers. That bug is fixed on main (#2805, `d5d52d47`), so the retry can land again. This PR is the original change plus the review feedback from #2803 applied.

## Problem

SQL Server CI jobs fail intermittently with:

```
pyodbc.Error: ('40001', '... Transaction (Process ID N) was deadlocked on lock resources
with another process and has been chosen as the deadlock victim. Rerun the transaction. (1205) ...')
```

Root cause: under pytest-xdist all workers share one database. Per-worker schemas isolate the *user tables*, but not the **system catalog** — the test harness's frequent metadata scans (`information_schema.tables` lookups in `ensure_test_table` and schema teardown) take shared locks across catalog rows of *all* schemas, while other workers run `CREATE TABLE` / `DROP TABLE` / `DROP SCHEMA` DDL holding exclusive locks on the same system base tables. SQL Server resolves the resulting deadlock by killing the (cheaper) reader.

## Fix

`SqlServerDataSourceConnection.execute_query` / `execute_update` retry statements killed as deadlock victims (SQLSTATE `40001`), Microsoft's documented remedy for error 1205:

- up to 3 attempts, then the error propagates
- `rollback()` between attempts to clear the aborted transaction
- full-jitter exponential backoff so both parties of a killed deadlock don't retry in lockstep
- all other `pyodbc` errors propagate unchanged

Fabric and Synapse connections subclass this connection and inherit the behavior.

## Review feedback from #2803

All items from the review on #2803 are applied, except one:

- The wrapper now documents its scope: it assumes single-statement units (the recovery rollback is connection-wide, and SQL Server proper runs `autocommit=False`), and the callback/iterator fetch paths (`execute_query_one_by_one*`, `execute_query_iterate`) are deliberately **not** wrapped — a deadlock mid-fetch may strike after rows were already delivered to the callback, so re-executing would double-process rows.
- Retry methods moved below `__init__`, matching the sibling connections.
- New tests: failing-rollback branch (retry proceeds; a persistent deadlock still raises the original 40001 error, not the rollback error), actual jitter range with un-stubbed `random.uniform` (seeded), deadlock-then-different-error propagating unchanged, and `caplog` assertions on the retry/rollback warnings.
- Deadlock fixture now raises `pyodbc.OperationalError`, matching what the driver raises for 1205.
- **Not applied:** exposing `DEADLOCK_MAX_ATTEMPTS` / backoff as connection properties. The existing `connection_max_retries` property maps to the ODBC driver's `ConnectRetryCount` (connection establishment), a different concern, and nothing needs to tune deadlock retries today — subclassing remains the escape hatch if that changes.

## Testing

9 unit tests in `soda-sqlserver/tests/unit/test_sqlserver_deadlock_retry.py`; full `soda-sqlserver`, `soda-fabric`, `soda-synapse` suites pass (90 tests).

### DIFF
--- a/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
+++ b/soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
@@ -1,15 +1,18 @@
 from __future__ import annotations
 
 import logging
+import random
 import struct
+import time
 from abc import ABC
 from datetime import datetime, timedelta, timezone, tzinfo
-from typing import Any, Literal, Optional, Union
+from typing import Any, Callable, Literal, Optional, Union
 
 import pyodbc
 from pydantic import Field, SecretStr
 from soda_core.__version__ import SODA_CORE_VERSION
 from soda_core.common.data_source_connection import DataSourceConnection
+from soda_core.common.data_source_results import QueryResult
 from soda_core.common.exceptions import DataSourceConnectionException
 from soda_core.common.logging_constants import soda_logger
 from soda_core.model.data_source.data_source import DataSourceBase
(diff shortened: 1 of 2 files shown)


soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
```diff
@@ -106,13 +109,62 @@ def handle_datetimeoffset(dto_value):
...
+
+    def _execute_with_deadlock_retry(self, operation: Callable[[], Any]) -> Any:
+        """Assumes ``operation`` is a single-statement unit of work: recovery
+        issues a connection-wide rollback, discarding any earlier uncommitted
+        statements on this connection (SQL Server runs autocommit=False).
+        Only the buffered paths (execute_query/execute_update) are wrapped;
+        the callback/iterator paths (execute_query_one_by_one*,
+        execute_query_iterate) may have already delivered rows when a deadlock
+        strikes mid-fetch, so re-executing them would double-process rows."""
+        for attempt in range(1, self.DEADLOCK_MAX_ATTEMPTS + 1):
```
Niels-b: Nit, not reachable today: with `DEADLOCK_MAX_ATTEMPTS = 0` this is an empty loop and the method falls off the end returning `None` — so `execute_query` hands `None` to a caller annotated `-> QueryResult` and it surfaces somewhere far away as `AttributeError: 'NoneType' object has no attribute 'rows'`.

Worth guarding only because the PR description designates subclassing as the tuning escape hatch, which makes `DEADLOCK_MAX_ATTEMPTS = 0` the natural way to express "disable deadlock retries" — i.e. the value that reads as safest is the one that breaks. (`1` behaves correctly: single attempt, error propagates.)

Either a `raise AssertionError("unreachable")` after the loop or `range(1, max(1, self.DEADLOCK_MAX_ATTEMPTS) + 1)` covers it.

Tangentially: typing `operation` as `Callable[[], T]` with a `TypeVar` instead of `Callable[[], Any] -> Any` would restore return-type checking on both wrappers, and would have made this a missing-return that mypy flags.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
```diff
@@ -106,13 +109,62 @@ def handle_datetimeoffset(dto_value):
...
+                return operation()
+            except pyodbc.Error as e:
+                if not self._is_deadlock_error(e) or attempt == self.DEADLOCK_MAX_ATTEMPTS:
+                    raise
+                logger.warning(
+                    f"Deadlock victim on '{self.name}' "
+                    f"(attempt {attempt}/{self.DEADLOCK_MAX_ATTEMPTS}), retrying: {e}"
+                )
+                try:
+                    self.connection.rollback()
```
Niels-b: Suggest `self.rollback()` here instead of reaching through to `self.connection`.

The base class exposes `rollback()` (`data_source_connection.py:499`) and the sibling error-recovery wrapper in postgres uses `self.rollback()` for exactly this purpose (`postgres_data_source_connection.py:129,138,157`). Going direct to `self.connection` skips any subclass override and skips the wrapper in `snapshot_connection.py:1256`.

No behavioural difference today — just keeps the seam consistent with the other connection that does error-recovery rollbacks.

Niels-b (approved):
Approving — the change is sound and correctly scoped.

Two things I checked that seemed worth confirming out loud, since the whole design rests on them:

- **The fix covers the paths that actually deadlock.** The catalog scan goes `query_existing_test_tables` → `MetadataTablesQuery.execute` → `execute_query` (`metadata_tables_query.py:48`), and the racing DDL goes through `execute_update` (`sqlserver_data_source_test_helper.py:55,61,66`, `data_source_test_helper.py:1178`). Neither side runs through the deliberately-unwrapped callback/iterator paths, so the exclusion doesn't leave a hole in the actual fix.
- **The "single-statement unit of work" assumption holds at every call site.** `_cursor_execute_update_and_commit` commits per `execute_update`, so each update is its own transaction, and every bulk-insert path passes one multi-`VALUES` `INSERT` rather than a `;`-separated batch — I looked for a call site that would send a batch and found none. That matters specifically for Fabric and Synapse, which inherit this wrapper but run `autocommit=True`, so their retry safety comes entirely from that invariant and not from the rollback.

Two non-blocking nits inline. Neither is reachable today; both are one-liners if you're touching the file anyway.

soda-sqlserver/src/soda_sqlserver/common/data_sources/sqlserver_data_source_connection.py
mivds: Definitely an obscure edge case, but implemented the proposed `max(1, ...)` as well as the `TypeVar`

Niels-b: Thanks for fixing this. Indeed very obscure, but may happen when people use this programmatically.

PR #2835, opened 2026-08-20
feat(cloud): add agent source to scan token usage

nikosml:
## Description

Scan token usage can now carry optional source attribution through the existing scan-results payload.

- add a trailing optional `agent_source` field to `ScanTokenUsage`
- serialize it as optional `agentSource`, omitting the key when unset
- consolidate token-usage dictionary construction while preserving result and entry order
- preserve the existing empty `tokenUsage` list behavior

Downstream packages that populate `agent_source` must require this or a later soda-core build. Existing callers remain compatible because the field is trailing and defaulted.

## Coordination

- Linear: [PLATL-970](https://linear.app/sodadata/issue/PLATL-970/thread-contract-autopilot-agentsource-into-scan-token-usage). Siblings: [soda#13146](https://github.com/sodadata/soda/pull/13146), [soda-core#2835](https://github.com/sodadata/soda-core/pull/2835), [soda-extensions#564](https://github.com/sodadata/soda-extensions/pull/564).
- Order: soda is independent; soda-core precedes soda-extensions, which must pin its published dev artifact.

## Checklist

### Customer-facing release note

```customer-facing-release-note
NONE
```

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1774,7 +1774,7 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
                 "totalTokens": tu.total_tokens,
                 "model": tu.model,
                 "operation": tu.operation,
-                "agentSource": tu.agent_source,
+                **({"agentSource": tu.agent_source} if tu.agent_source is not None else {}),
             }
             for tu in contract_verification_result.token_usage
         ]
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -1774,6 +1774,7 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
                 "totalTokens": tu.total_tokens,
                 "model": tu.model,
                 "operation": tu.operation,
+                "agentSource": tu.agent_source,
             }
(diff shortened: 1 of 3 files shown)


m1n0 (approved):
Approving.

- The consolidation isn't cosmetic: on main `_build_token_usage_dicts` is only referenced by the tests, the payload builder kept its own inline copy. Wiring it in at soda_cloud.py:1841 is what actually gets the field on the wire — worth saying that in the description, "preserving result and entry order" reads like tidying.
- Otherwise the swap is behaviour-neutral: the old `if r.token_usage` and the helper's guard agree on None and [], result and entry order unchanged.
- Ran the token-usage tests at 67c330fa (4 passed), then deleted the `agentSource` line and reran — the helper test and the new payload test both fail, so they gate the field rather than just describing it.
- Backend side lines up: ScanTokenUsageDTO.agentSource is a Gson enum with @SerializedName SODA/BYOK, unknown values deserialize to null, and SodaCoreScanConverterModule null-guards. Your comment at contract_verification.py:561 matches what the server does.
- Only the first CI run is red; the run on the head commit is green on all 13 lanes, unit lane included. Moving off the fork did the job.

Merge window: soda#13146 is in, this goes next, then ext#564 — its pin wants bumping to the dev build this merge publishes. `soda-core>=4.21.0.dev2765` will resolve to something that has the field but doesn't require it.

PR #2837, opened 2026-08-21
chore: set codeowners & update PR template

mivds:
* Adds `CODEOWNERS` so reviewers are added automatically
* Updates PR template to remove reference to soda-internal repos

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,15 +1,11 @@
 ## Description
 
-Please provide a description of your changes:
+<!-- What does this PR change and why? Link the related issue if there is one. -->
 
-- What problem are you solving?
-- Any expected impact on downstream packages/services?
-- Reference issue # if available.
-
+Fixes #
 
 ## Checklist
 
-- [ ] I added a test to verify the new functionality.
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -1,15 +1,11 @@
...
-- Any expected impact on downstream packages/services?
-- Reference issue # if available.
+<!-- What does this PR change and why? Link the related issue if there is one. -->
 
+Fixes #
 
 ## Checklist
 
-- [ ] I added a test to verify the new functionality.
-- [ ] I verified this PR does not break [soda-extensions][1].
```
Niels-b, line 13: Are we sure we want to drop this soda-extensions one 😅 ?

Niels-b (approved):

.github/pull_request_template.md
mivds: I would say yes, because this is a public repo where we also get contributions from the community. Referencing a repo/pipeline they don't have access to doesn't make sense. Up to us maintainers to validate the impact

Niels-b: Ok, that makes sense indeed. We just have to be extra careful now then 😉

PR #2838, opened 2026-08-21
feat(cli): batched scan ingestion — streamed logs and async results

mivds:
## Problem

Cloud-launched scans (`SODA_SCAN_ID` set by the launcher) deliver nothing to Soda Cloud until the run ends: logs are gathered in memory and embedded in the single end-of-run `sodaCoreInsertScanResults` / `sodaCoreMarkScanFailed` payload. For a long scan nothing is visible mid-run, and a hard kill (OOM, SIGKILL) loses the logs entirely.

## Change

Adds opt-in batched ingestion for Cloud-launched CLI flows; ad-hoc runs are untouched.

Every CLI results-publishing command runs under a **`ScanContext`** — always present, sourced from a ContextVar (mirroring `logs._active_logs`) rather than threaded as a parameter — with two variants:

- `AtomicScanContext`: today's behavior — one indivisible end-of-run sync upload;
- `BatchedScanContext`: the async pipeline `sodaCoreScanStart` → `sodaCoreInsertScanDataBatch` → `sodaCoreScanEndAsync` (new stateless transport methods on `SodaCloud`), with logs streaming mid-run over the scan-id-keyed `batchV4` endpoint.

The CLI bracket `run_scan` is the single site that picks the variant and installs it: the wiring declares participation (`batched=True` on discover here, and on the profiling / capture-schema / metric-monitoring wirings in soda-extensions), and `SODA_SCAN_ID` decides whether it activates. Contract verification keeps the default and always runs under an `AtomicScanContext`, so its ingestion stays sync. Flows call `get_scan_context()` and never branch on a scan id, on `None`, or on which variant they got. `run_with_failure_reporting` is absorbed into the bracket — it had exactly one direct caller left in core (contract verify), and merging makes the "never end a scan after a failure report" invariant straight-line control flow, with one owner for the `Logs` lifecycle. The name stays importable from `dependencies` as a deprecated alias over `run_scan` for one release: the published profiling, metric-monitoring and capture-schema CLIs import it, and their soda-core pins are lower bounds only.

Because the backend only accepts `batchV4` uploads after a successful `sodaCoreScanStart` (which needs the flow's scan coordinates), the flow opens the scan itself via `context.start_scan(...)` once its names resolve — the run's `Logs` then upgrades from the in-memory collector to a streaming `LogsQueue` with the captured history replayed, so an ad-hoc run keeps today's payload-embedded logs. A failed `sodaCoreScanStart` — rejected or raised — **fails the run** (`ScanExecutionFailedException` → `sodaCoreMarkScanFailed` with the full in-memory record list, exit 3) instead of silently degrading to the sync upload: the backend refuses that command deliberately, and continuing would bury the misconfiguration. Transient start failures are accepted casualties until retries are handled at the Soda Cloud client level (separate task). Failure reports attach `records_for_failure_report()` — the full record list in-memory, only the undelivered (non-2xx-acknowledged) records when streaming — keeping the stream the single log channel.

Exit-code semantics are preserved: an unacked upload still exits `RESULTS_NOT_SENT_TO_CLOUD`, `run_scan` stays the single Cloud-marking site, and `context.end_scan()` sends `sodaCoreScanEndAsync` only for a run whose every upload was acknowledged, so an undelivered run leaves the scan's terminal state to the failure report / exit-code fallback. **No retries on `sodaCoreScanEndAsync`** — Cloud-command retries are a global concern tracked as separate work; until that lands, a transient failure at scan end loses the run's results (the exit code keeps the state truthful). Contract verification's results ingestion is unchanged: it runs under `run_scan` with an atomic context it never touches. Its diagnostics-warehouse log stream (soda-extensions) shares the rebuilt `LogsQueue`, so it gets the delivery behavior below.

The `LogsQueue` itself is rebuilt around one invariant: **the queue holds exactly the unacknowledged records** — a flush sends the head in one request and removes it only on a 2xx, so a failed send leaves the records queued and the worker's cadence re-sends them (the cadence is the retry; there is no retry loop, no drop accounting, no delivered-record bookkeeping). Only the two refusals the backend makes about the scan itself end the stream (verified against the log endpoint handlers: 404 `scan_not_found`, and 400 `invalid_scan_state` for a scan off its log-accepting state — transitions that never reverse); 408/429 stay queued like a 5xx, and any other 4xx (413, a malformed batch) drops that one batch and the stream goes on, as on main; unserializable records are evicted per record so one bad record cannot block the stream; the log-batch posts re-authenticate once on a 401 in the `SodaCloud` client so a token expiring mid-run doesn't kill the stream. Also fixes a latent bug this feature would have made reachable: `get_error_logs` read the nonexistent `LogRecord.level` (crash on any streaming run with an error record).

A batch upload lands in object storage, so Soda Cloud returns none of the ids (scan, dataset, check) that a synchronous insert hands back. Post-processing handlers that need those ids therefore cannot run on a batched run, and a run must not promise a stage nobody will complete: a stage left ONGOING never clears on Soda Cloud, and the backend holds the scan's logs pending while any stage is incomplete. So the two sides now declare their requirement: a handler sets `requires_result_handles` (the failed-rows extractor does — its warehouse tables are keyed by the dataset id the insert returns), a context reports `provides_result_handles` (false only for the batched variant), and `collect_post_processing_stages()` declares the intersection. **The trade: a batched run has no failed-row diagnostics.** That is acceptable for the flows opting in here — metric monitoring's anomaly checks do not extract failed rows today — and getting result handles without a synchronous insert is a backend ask for later.

## Known gaps (for the retries / logging follow-ups)

- A streaming run whose results insert is rejected exits 4 without a failure report; the launcher's fallback `markScanFailed` then replaces the scan's stored logs with its generic line, deleting the already-streamed history. No regression vs main (the sync path loses those logs too).
- The combine-uploads failure path (`session.py` mark site) attaches `log_records`, which is `[]` on a streaming run — carries a warning comment; must move to `records_for_failure_report()` before any combine-uploads flow opts into batching.

## Testing

Unit-tested (transport methods, the `LogsQueue` delivery invariant and accessors, context variants, `run_scan` bracket matrix including the two-channel streaming pipeline against a recording Cloud double, discovery wiring; full suite green in soda-core and the extensions companion). Previously verified end-to-end against a live Soda Cloud environment with launcher-managed scans: logs became visible while the scan was still running, results were ingested through the batch pipeline, and the final stored logs matched the run's full record count.


Niels-b:
Pushed a follow-up series (`39d8a44b..de6ab6ee`) on top of the original commits. The architecture is unchanged; the series fixes three verified defects, simplifies the layering, and rebuilds the tests around the real two-channel pipeline:

- **Failure reports no longer destroy the streamed logs.** `sodaCoreMarkScanFailed` REPLACES a scan's stored logs on the backend (the orphaned streamed rows then get deleted), so `LogsQueue.records_for_failure_report()` now flushes first and attaches only what the stream could not deliver — and **retires** the stream, since the backend rejects every `batchV4` upload after the mark (previously the run's own teardown ended every managed streamed failure with a false "records could not be delivered" alarm). Nothing else stops the stream: errors from a run that continues keep streaming to the end.
- **A lost `sodaCoreScanEndAsync` is fatal, and gated.** Batch uploads sit inert in object storage until the end command triggers reassembly, and no backend sweeper does it later — so a rejected/raising end now retries (permanent 4xx: one attempt) and exits `RESULTS_NOT_SENT_TO_CLOUD` instead of warning over exit 0. The end is only sent for a completed run (started ∧ every upload acked ∧ no escaped failure): a failure after a delivered insert never sends two terminal commands, and a cancelled run (SIGTERM) leaves the scan to the launcher fallback.
- **The `thread` grouping label survives.** `hasattr(record, "thread")` is always true (the stdlib stamps the OS thread ident), so the caller-set label is now marked explicitly (`THREAD_LABEL_ATTR`); unlabelled streams get the queue's uuid instead of a meaningless number.
- **Log-batch failures are retried and accounted.** 5xx/408/429/exceptions retried (backoff waits on the shutdown event), plain 4xx dropped after one post, every drop counted with a close-time summary; the stream's own diagnostics go to an uncapturable logger so a failing stream can't feed reports about itself into its own queue.
- **Layering:** `BatchedScanContext` moved to `soda_core/common` (the engine receives it — kills the TYPE_CHECKING dances and deferred imports), the bracket lost its `stage` parameter, the context owns the command-type stamp, and the results payload builder gained a `model_version` parameter (lets metric monitoring retire its `_execute_command` monkeypatch).
- **Tests** drive a real `LogsQueue` into a recording `MockSodaCloud`: the full command order, the empty-report-on-healthy-stream property with the flush provably preceding the mark, the terminal-state matrix (cancellation, failure-after-insert, partly rejected session, rejected/raising end), the retry/drop matrix, the wire value of `thread`, and `dataTimestamp` by object identity.

An independent adversarial review ran over the series (verdict: no blockers; its findings are fixed in the last three commits). Suites: 1259 unit, 178 postgres integration, feature + soda-postgres packages — all green.

Merge order unchanged: this PR first, then the soda-extensions floor bump per the capture-schema pattern.

soda-core/src/soda_core/cli/handlers/dependencies.py
```diff
@@ -1,30 +1,27 @@
...
 Handlers receive fully constructed dependencies (``DataSourceImpl``,
 ``SodaCloud``) instead of file paths. Resolvers raise
 ``ScanExecutionFailedException`` for expected/unusable-configuration shapes;
-``run_with_failure_reporting`` owns the failure-to-exit-code mapping for a
-command and knows nothing about construction — the wiring layer decides which
-resolutions run inside it. Like ``failure_reporting``, this module is a stable
-import point: soda-extensions CLIs reuse these utilities for their own
+``scan.run_scan`` owns the failure-to-exit-code mapping for a command and
+knows nothing about construction — the wiring layer decides which resolutions
+run inside it. Like ``failure_reporting``, this module is a stable import
```
m1n0, line 8: This still calls the module a stable import point, but `run_with_failure_reporting` is gone. The profiling, monitor and capture-schema CLIs on extensions main import it, and their soda-core pins only have a lower bound. An already published extension next to this core fails at plugin load, and since that only warns, the commands just disappear. Can we keep it as a thin alias over `run_scan` for a release?

soda-core/src/soda_core/common/logs_queue.py
```diff
@@ -3,38 +3,58 @@
...
+from soda_core.common.logs_base import LOG_STAGE_MAIN, STREAM_DIAGNOSTICS_LOGGER, THREAD_LABEL_ATTR, LogsBase
 from soda_core.common.soda_cloud import SodaCloud, to_jsonnable
 
 DEFAULT_FLUSH_INTERVAL = 5
 MAX_LOG_LINES = int(os.environ.get("SODA_LOGS_BATCH_LIMIT_COUNT", "1000"))
-MAX_RETRIES = 3
 
+# 4xx that can heal (timeout, rate limit). Any other 4xx on the log endpoints means the scan
+# permanently refuses uploads: deleted (404), or off its log-accepting state (400) — for good.
+_TRANSIENT_4XX = {408, 429}
```
m1n0, line 24: Any 4xx outside 408/429 now stops the stream for the rest of the run. On main a 4xx cost one batch and the stream kept going. This isn't behind `batched`, so it also hits streams that never opted in: test-connection, plus the diagnostics warehouse and lineage uploads, the first of which runs inside contract verify. A 413 on a 1000-line batch full of tracebacks would silence everything after it while the run still exits 0. I'd only go terminal on 404 and the scan-state 400, and drop the batch otherwise. The body also says contract verification is untouched, which isn't true for its log streams.

soda-core/src/soda_core/cli/cli.py
```diff
@@ -546,19 +546,19 @@ def handle(args):
...
                 soda_cloud,
                 lambda logs: handle_discover_data_source(
                     resolve_data_source(args.data_source),
-                    soda_cloud,
                     scan_definition_name=resolve_scan_definition_name(args.scan_definition_name),
                     include=args.include,
                     exclude=args.exclude,
                     logs=logs,
                 ),
+                batched=True,
```
m1n0, line 561: This moves every managed discover to batched as soon as a runner ships it, and the only way back is a new release. With no retry on `sodaCoreScanEndAsync`, one blip there loses the run's results, and the stuck-scan cleanup just times the scan out after 23h without ingesting. A retry isn't safe yet either, since a repeated end isn't deduped server-side. Can we get an env opt-out until retries land in the client?

m1n0 (changes requested):
Three things inline I'd want sorted before merge: the removed `run_with_failure_reporting`, the new 4xx handling in `LogsQueue` applying to every stream, and no way to turn batched discover off.

Also the branch doesn't have #2855 yet, which is why the databricks lanes are red on the extensions side. Merging main should fix that.

soda-core/src/soda_core/cli/handlers/dependencies.py
mivds: Restored in 570ebf4b as a deprecated alias over `run_scan` (via the existing `deprecated_alias` helper), with a comment saying why it stays and when to drop it. Default `batched=False` reproduces the old behavior exactly. The module docstring and PR body now say so.

soda-core/src/soda_core/common/logs_queue.py
mivds: Fixed in 094c83fb. I checked the backend handlers for both log endpoints (batchV3 and batchV4): they raise exactly two refusals about the scan itself, 404 `scan_not_found` and 400 `invalid_scan_state`. A 400 alone is ambiguous (a malformed batch is a 400 too), so the terminal decision now reads the error body's `code`. Every other 4xx drops that one batch with a warning and the stream keeps going, as on main. 408/429 stay queued like a 5xx. Tests cover each case, and the PR body no longer claims contract verification is untouched: its diagnostics-warehouse stream shares the rebuilt queue.

soda-core/src/soda_core/cli/cli.py
mivds: Not adding an opt-out. The sync path has no retry either, so a blip on `sodaCoreInsertScanResults` loses the run just the same. And for large data sources the discovery payload is big enough that the sync command fails outright, which is what the batched pipeline fixes. Falling back to sync would trade a rare transient loss for a deterministic one on exactly the sources this is for. Retries on the end command land with the client-level retry work.

mivds:
Merged main into the branch (e9921ad2), which brings in #2855 for the databricks lanes.

Niels-b (commented):
Two notes from a pass over the current head. Neither is blocking and neither reopens the three inline items already raised — those look settled to me. I checked the log-stream one against the backend rather than the PR body: `SodaCoreBatchLogV4UploadRequestHandler:56-59` and its V3 twin raise exactly `SCAN_NOT_FOUND` (404) and `INVALID_SCAN_STATE` (400), and `ErrorCodes.INVALID_SCAN_STATE` is the literal `"invalid_scan_state"`, so `_is_permanent_rejection` matches those two and nothing wider.

soda-core/src/soda_core/common/scan_context.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # Whether the results insert hands back the Cloud-minted ids (scan, dataset, check) that
+    # post-processing needs. False for a batch upload: it lands in object storage.
+    provides_result_handles: bool = True
+
+    def __init__(self, soda_cloud: Optional[SodaCloud]):
+        self.soda_cloud: Optional[SodaCloud] = soda_cloud
+        # The launcher-created scan this run reports into; None on an ad-hoc run. The run's
+        # identity, so flows read it here instead of the environment.
+        self.scan_id: Optional[str] = None
```
Niels-b: `scan_id` is only ever assigned by `BatchedScanContext`; the base leaves it `None`. So on a managed run that takes the atomic path — contract verify, or any `batched=False` flow with `SODA_SCAN_ID` set — this reads `None` while `EnvConfigHelper().soda_scan_id` has a value.

The comment above reads as a general accessor ("The run's identity, so flows read it here instead of the environment"), and soda-extensions#574 already takes it at its word: `MonitoringImpl.verify()` drops its own `self.soda_config.soda_scan_id` guard and passes `scan_context.scan_id` straight to `mark_scan_as_failed`. Safe there because MM is now always batched, but the next flow to read it on the atomic path gets a silent `None`. Either populate it from the env in `AtomicScanContext` too, or say in the comment that it is set on the batched variant only.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -160,6 +164,38 @@ def register(cls, verification_handler: ContractVerificationHandler) -> None:
...
+    loop inside ``verify()`` and the session executor's phase 3) go through here, so a handler
+    that declares no stage on this run is not run against ``None`` ids either."""
+    return [
+        handler
+        for handler in ContractVerificationHandlerRegistry.contract_verification_handlers
+        if _runs_on_current_scan(handler)
+    ]
+
+
+def collect_post_processing_stages() -> list[PostProcessingStage]:
```
Niels-b, line 186: Worth recording where this lands on the wire, because it isn't what the PR body implies. `SodaCoreInsertScanDataBatchCommand` has no `postProcessingStages` field — it sits on `SodaCoreScanEndAsyncCommand`, and `SodaCloud.scan_end_async()` sends that command with only `scanReference`. So on a batched run the stages collected here never reach Soda Cloud at all, declared or not.

The outcome is the one you want either way (no stage declared, nothing stuck ONGOING), and the filter is genuinely load-bearing on the atomic path where the stages do ride the results payload. Nothing to change — just noting that on the batched path this is belt-and-braces, and that `sodaCoreScanEndAsync.postProcessingStages` is the seam to use if a batched flow ever does need a stage declared.

soda-core/src/soda_core/common/scan_context.py
mivds: Populated on both variants in 4386433c: `ScanContext.__init__` takes the scan id and `run_scan` hands it to the atomic context as well, so a managed run on the atomic path reads the same value the environment has. The comment now says it is set on either variant; the variant only picks the pipeline. The managed-atomic test asserts the id, the ad-hoc one asserts None.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
mivds: Confirmed against the backend: `postProcessingStages` sits on `SodaCoreScanEndAsyncCommand` and not on `SodaCoreInsertScanDataBatchCommand`, and `scan_end_async` sends only `scanReference`. The docstring on `collect_post_processing_stages` now says so (415501a7), including that the end command is the place to declare a stage if a batched flow ever needs one. No behavior change.

soda-core/src/soda_core/common/scan_context.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class AtomicScanContext(ScanContext):
+    """One synchronous end-of-run upload — the ad-hoc behavior."""
+
+    def _send_results(self, payload: SodaCoreInsertScanResultsDTO) -> bool:
+        assert self.soda_cloud is not None, (
+            "insert_results needs a Soda Cloud client: run the flow under cli.handlers.scan.run_scan "
+            "(or inside using_scan_context with a context that has one)."
+        )
```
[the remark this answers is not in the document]
mivds: @AikidoSec ignore: pins an invariant of the CLI bracket, not a runtime guard. insert_results is only reachable under run_scan, which always installs a context with a Soda Cloud client; outside it the inert default exists so callers never check for None. A raise here would guard a state that cannot occur.

soda-core/src/soda_core/common/scan_context.py
```diff
@@ -0,0 +1,188 @@
...
+            raise ScanExecutionFailedException(
+                f"Soda Cloud did not accept sodaCoreScanStart for scan '{self.scan_id}'."
+            )
+        self.scan_reference = scan_reference
+        # The backend accepts batchV4 uploads only after the start, so the stream begins here.
+        self.logs.switch_gatherer(build_streaming_gatherer(self.soda_cloud, scan_id=self.scan_id))
+
+    def _send_results(self, payload: SodaCoreInsertScanResultsDTO) -> bool:
+        # Inserting before start_scan is a flow bug; a failed start already failed the run.
+        assert self.scan_reference, "insert_results on a batched scan requires a successful start_scan first."
```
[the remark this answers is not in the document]
mivds: @AikidoSec ignore: pins an invariant, not a runtime guard. A batched flow calls start_scan before insert_results, and a failed start already fails the run, so scan_reference is always set here.

Niels-b (approved):

m1n0 (approved):

PR #2841, opened 2026-08-25
fix(sql-dialect): escape regex patterns as string literals

m1n0:
Chain: **this** → #2843 → #2842. Merge bottom-up.

`REGEXP_LIKE` hand-wrapped the pattern in quotes instead of going through
`literal_string()`. On data sources whose SQL parser consumes backslashes in string
literals, `^-?\d+$` reaches the regex engine as `^-?d+$` — matches nothing, so every
non-null row is reported invalid with no error. An apostrophe in a pattern breaks the
query outright.

Five dialects override the builder and needed the same fix:

| dialect | rendered | note |
|---|---|---|
| redshift | `{expr} ~ …` | parser eats backslashes; needs the doubled form [AWS documents](https://docs.aws.amazon.com/redshift/latest/dg/pattern-matching-conditions-posix.html) |
| postgres | `{expr} ~ …` | backslash-safe already, now quote-safe |
| duckdb | `REGEXP_MATCHES(…)` | backslash-safe already, now quote-safe |
| bigquery | `REGEXP_CONTAINS(…)` | its `literal_string()` is triple-quoted, so the pattern reaches the engine as the old `r''` prefix delivered it; apostrophe now escaped |
| sqlserver | `PATINDEX('%…%', …) > 0` | quote-safe; keeps the full override because it rewrites the pattern text |

In soda-extensions: `soda-db2z` already escapes quotes; `soda-db2` and `soda-dremio` carry
copies of the old base implementation and should be deleted so they inherit the fix.

Verified on live Snowflake — [without the fix](https://github.com/sodadata/soda-core/actions/runs/32833001202)
fails on 3.10–3.14 with `assert 3 == 1`, [with it](https://github.com/sodadata/soda-core/actions/runs/32833010800)
passes. `invalid_count: 3` on a 4-row table with one invalid row, which is the report on
SCS-1230: "showing all rows failed".

Tests: `…_escaped_metacharacter` (`^1\.5$` against `1.5` / `1x5` — an escaped metacharacter
rather than `\d`, which POSIX engines read as a literal `d` and so cannot distinguish
"unsupported" from "backslash eaten") and `…_apostrophe` (`^it's$`, exercising `''` and
`\'` live). Per-dialect unit tests for each override.

v3 escapes user regexes via `data_source.escape_regex()`; v4 did not, so a SodaCL
`valid regex: \d+` that worked breaks when migrated to a v4 contract on Snowflake. This
restores parity and drops the dead base `escape_regex()` hook.

**Behaviour change:** a regex field is a regex, not a SQL literal. Doubled-backslash
workarounds must revert to single backslashes. SQL-level escape sequences in a pattern are
no longer interpreted by the data source — matches v3; see SCS-728, SAS-7560.

**Release note:** Fixed: regex patterns in `valid_format` / `invalid_format` /
`missing_format` were not escaped for the data source, so on Snowflake, Databricks and
Redshift a backslash in a pattern never reached the regex engine and every non-null row
could be reported invalid. If you worked around this by doubling backslashes in your
contract, revert those patterns to single backslashes.

Refs SCS-1413, SCS-1230.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -264,7 +264,11 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
 
     def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
         expression: str = self.build_expression_sql(matches.expression)
-        return f"REGEXP_CONTAINS({expression}, r'{matches.regex_pattern}')"
+        # literal_string() below is BigQuery's own triple-quoted form, which escapes
+        # the backslash as \\ -- so the pattern reaches the regex engine exactly as the
+        # r'' prefix used to deliver it, and an apostrophe is escaped as \' instead of
+        # terminating the literal. See SCS-1413.
+        return f"REGEXP_CONTAINS({expression}, {self.literal_string(matches.regex_pattern)})"
 
     def _build_percentile_within_group_sql(self, percentile_within_group: PERCENTILE_WITHIN_GROUP) -> str:
         """BigQuery has no ordered-set aggregates; render as
--- a/soda-bigquery/tests/unit/test_bigquery_dialect.py
+++ b/soda-bigquery/tests/unit/test_bigquery_dialect.py
@@ -5,7 +5,7 @@ from soda_bigquery.common.data_sources.bigquery_data_source import (
 )
 from soda_core.common.metadata_types import SqlDataType
 from soda_core.common.sql_ast import CREATE_TABLE, CREATE_TABLE_COLUMN
-from soda_core.common.sql_dialect import FROM, RANDOM, SELECT
+from soda_core.common.sql_dialect import COLUMN, FROM, RANDOM, REGEX_LIKE, SELECT
 
 
 def test_random():
@@ -398,3 +398,17 @@ def test_time_delta_single_week_still_floors_by_seven_days():
     )
     assert "WEEK" not in sql
     assert sql.endswith("DAY) / 7) AS INT)")
+
+
+def test_regex_like_pattern_goes_through_literal_string():
+    """BigQuerySqlDialect overrides _build_regex_like_sql, so it needs its own copy of
+    the base-dialect guarantee. Its literal_string() is the triple-quoted form, which
+    escapes the backslash as `\\\\` -- so the pattern reaches the regex engine exactly as
+    the old r'' prefix delivered it -- and escapes an apostrophe as `\\'` rather than
+    letting it terminate the literal. See SCS-1413.
+    """
+    sql_dialect = BigQuerySqlDialect()
+    assert (
+        sql_dialect.build_expression_sql(REGEX_LIKE(COLUMN("c"), r"^1\.5$")) == "REGEXP_CONTAINS(`c`, '''^1\\\\.5$''')"
+    )
+    assert sql_dialect.build_expression_sql(REGEX_LIKE(COLUMN("c"), "^it's$")) == "REGEXP_CONTAINS(`c`, '''^it\\'s$''')"
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -416,9 +416,6 @@ class SqlDialect:
         string_literal: str = value.replace("'", "''")
         return string_literal
 
-    def escape_regex(self, value: str):
-        return value
-
     def apply_default_add_semicolon(self, add_semicolon: Optional[bool]) -> bool:
         return add_semicolon if add_semicolon is not None else self.USES_SEMICOLONS_BY_DEFAULT
 
@@ -1250,7 +1247,13 @@ class SqlDialect:
 
     def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
         expression: str = self.build_expression_sql(matches.expression)
-        return f"REGEXP_LIKE({expression}, '{matches.regex_pattern}')"
+        # The pattern is a string literal like any other, so it goes through
+        # literal_string(). Data sources whose parser consumes backslashes in
+        # string literals (Snowflake, Databricks, Redshift) would otherwise
+        # hand the regex engine `d` where the user wrote `\d`, silently
+        # matching nothing, and an apostrophe in a pattern would break the
+        # query outright.
+        return f"REGEXP_LIKE({expression}, {self.literal_string(matches.regex_pattern)})"
 
     def _build_lower_sql(self, lower: LOWER) -> str:
         return f"LOWER({self.build_expression_sql(lower.expression)})"
--- a/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
+++ b/soda-duckdb/src/soda_duckdb/common/data_sources/duckdb_data_source.py
@@ -121,7 +121,7 @@ class DuckDBSqlDialect(SqlDialect, sqlglot_dialect="duckdb"):
 
     def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
         expression: str = self.build_expression_sql(matches.expression)
-        return f"REGEXP_MATCHES({expression}, '{matches.regex_pattern}')"
+        return f"REGEXP_MATCHES({expression}, {self.literal_string(matches.regex_pattern)})"
 
     def create_schema_if_not_exists_sql(self, prefixes: list[str], add_semicolon: bool = True) -> str:
         schema_name: str = prefixes[0]
--- a/soda-duckdb/tests/unit/test_duckdb_dialect.py
+++ b/soda-duckdb/tests/unit/test_duckdb_dialect.py
@@ -1,4 +1,4 @@
-from soda_core.common.sql_dialect import FROM, RANDOM, SELECT
+from soda_core.common.sql_dialect import COLUMN, FROM, RANDOM, REGEX_LIKE, SELECT
 from soda_core.common.statements.metadata_primary_keys_query import (
     MetadataPrimaryKeysQuery,
 )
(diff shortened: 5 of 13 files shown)


m1n0:
Ordering note vs #2842 (isort `line_length` 88 → 120):

If #2842 merges first, this branch needs one follow-up — the `soda_core.common.sql_dialect` import in `soda-snowflake/tests/unit/test_snowflake_dialect.py` is currently wrapped to satisfy isort at 88, and at 120 isort will want it back on a single line. One-line rebase fix.

If this merges first, #2842 just needs a rebase; its reformat will pick the import up automatically.

No functional interaction either way.

Niels-b:
The "release note" is missing from the description (even though there is a reference to it)

Niels-b (commented):
Reviewed against a fresh checkout of the branch. The core change is right — `REGEXP_LIKE` hand-wrapping the pattern instead of going through `literal_string()` is a real bug, and the live-Snowflake A/B is convincing. One thing needs fixing before merge, plus a correction to the follow-up scope and two nits.

## Redshift isn't fixed by this, and CI says so

`RedshiftSqlDialect` overrides the builder, so the base-class change never runs for it:

```python
# soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source.py:180
def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
    expression: str = self.build_expression_sql(matches.expression)
    return f"{expression} ~ '{matches.regex_pattern}'"
```

The [Redshift job on `7af47b8`](https://github.com/sodadata/soda-core/actions/runs/32834397201/job/97760092917) fails on the new test:

```
FAILED test_invalid_check.py::test_invalid_count_valid_regex_escaped_metacharacter
AssertionError: Expected contract verification failed
```

`assert_contract_fail` failing means the contract *passed* — `^1\.5$` arrived as `^1.5$`, `.` matched `1x5`, `invalid_count == 0`. That is SCS-1413 reproducing on the branch that fixes it, and both the PR table and the customer-facing release note name Redshift.

The upside: that failure is direct live evidence that Redshift really does consume backslashes in string literals, confirming the `standard_conforming_strings is off` comment on its own `escape_string`. So the fix is provably a one-liner:

```python
return f"{expression} ~ {self.literal_string(matches.regex_pattern)}"
```

I'd pull that into this PR rather than leave it to the follow-up — Redshift is already in the advertised scope, and the test that covers it is already here.

## The follow-up section is missing five dialects

It says soda-db2 and soda-dremio "are the only dialects that still won't escape an apostrophe in a pattern". Five in-repo dialects also bypass the base:

| dialect | line | rendered |
|---|---|---|
| redshift | `redshift_data_source.py:182` | `{expr} ~ '{pattern}'` |
| postgres | `postgres_data_source.py:110` | `{expr} ~ '{pattern}'` |
| duckdb | `duckdb_data_source.py:124` | `REGEXP_MATCHES(..., '{pattern}')` |
| bigquery | `bigquery_data_source.py:267` | `REGEXP_CONTAINS(..., r'{pattern}')` |
| sqlserver | `sqlserver_data_source.py:321` | `PATINDEX('%{pattern}%', ...)` |

Postgres, DuckDB and BigQuery are backslash-safe by accident (standard-conforming literals; BigQuery's `r''` prefix), which is why only Redshift went red — but an apostrophe in a pattern still breaks the query outright on all of them. `soda-db2z` is the one that already escapes quotes (`pattern.replace("'", "''")`), so it can come off the list.

As merged today the fix covers Snowflake, Databricks (+ sparkdf / hive), Trino, Athena and Oracle. Worth saying that explicitly if the overrides aren't converted here. For BigQuery, `literal_string()` works too (`'''…'''` with `\\` / `\'`), which would let the `r` prefix go.

## Nits

- `test_regex_like_escapes_pattern` guards base-class behaviour but lives only in soda-snowflake. A twin in soda-core against the default dialect (`'^it''s$'`) would guard the base builder in the core unit suite, and per-dialect ones for redshift/postgres/duckdb would have surfaced the override gap before CI did.
- The new test's docstring says "Redshift's `REGEXP_LIKE` is POSIX" — Redshift renders `~`, not `REGEXP_LIKE`. And the earlier `^\d$` revision failing on Redshift doesn't actually discriminate between "POSIX has no `\d`" and "the backslash was eaten": both match nothing. The backslash explanation is the one this run proved.

m1n0:
Good catch on the override — that was the whole thing. I'd been chasing the Redshift failure on the assumption the base builder ran for it, which is why nothing added up: I probed `REGEXP_LIKE` against live Redshift and it doesn't resolve there at all, because Redshift renders `~`. All fixed in eef83fe.

**Redshift, plus the other three.** Converted redshift, postgres, duckdb and sqlserver to `literal_string()`. Only Redshift changes on backslashes; postgres and duckdb were backslash-safe already but all four now escape an apostrophe. Verified the rendered SQL per dialect:

| dialect | `^1\.5$` | `^it's$` |
|---|---|---|
| redshift | `"c" ~ '^1\\\\.5$'` | `"c" ~ '^it''s$'` |
| postgres | `"c" ~ '^1\\.5$'` | `"c" ~ '^it''s$'` |
| duckdb | `REGEXP_MATCHES("c", '^1\\.5$')` | `REGEXP_MATCHES("c", '^it''s$')` |

The doubled Redshift form is what AWS documents — [POSIX operators](https://docs.aws.amazon.com/redshift/latest/dg/pattern-matching-conditions-posix.html) says to escape a metacharacter with two backslashes, with `'.*\\\\..*'` as the worked example.

**BigQuery I did not convert.** Its `r''` prefix already makes it backslash-safe, which is why it never went red, but doing the quote correctly needs BigQuery-specific raw/triple-quoted handling rather than the shared `literal_string()`, and I can't validate that from here. It's now called out in the PR body as the one known gap rather than left implied.

**Tests.** Added the base-dialect twin in `soda-tests/tests/unit/test_sql_generation.py` plus per-dialect copies in redshift, postgres and duckdb — exactly your point that snowflake-only coverage is what hid four override gaps. Worth noting for anyone running these locally: a worktree using the main checkout's venv resolves the dialect packages to the *main* checkout, so these tests silently pass against unpatched code unless you put the worktree's `src` dirs on `PYTHONPATH`. Cost me a confusing run.

**Docstring.** Fixed — it no longer says Redshift uses `REGEXP_LIKE`, and I dropped the claim that the earlier `^\d$` failure proved anything about backslashes. You're right that it doesn't discriminate; the escaped-metacharacter form does, and that's now the stated reason for choosing it.

**Follow-up list.** Corrected to the five overrides with db2z removed, and the release note is now visible text in the description instead of only the HTML comment.

m1n0:

m1n0:
The isort ordering note I posted earlier is resolved by the chain: #2842 now sits on top, so its reformat picks up the `test_snowflake_dialect.py` import automatically and no manual unwrap is needed either way.

Niels-b (approved):
Approving — the blocker is properly fixed and `eef83fe` is green, Redshift job included.

All four overriding dialects now route through `literal_string()`, and the details are right: sqlserver still wraps `%…%` before escaping, and Redshift emits the doubled form AWS documents. The unit coverage came back better than I asked for — a base-dialect test plus per-dialect twins in redshift/postgres/duckdb, each asserting both the backslash and the apostrophe form. Follow-up scoping is fixed too (db2z correctly dropped, db2/dremio left to the extensions PR, #2843 removing the seam that let this be re-implemented per dialect).

Two non-blocking things, both in the description rather than the code.

## 1. "Why it went unnoticed" now contradicts the corrected docstring

It still reads:

> `\d` is a PCRE-ism, and Redshift's `REGEXP_LIKE` is POSIX ERE where `\d` is not a digit class on any code path. An earlier revision of this PR used `^\d$` and failed the Redshift job for exactly that reason.

That's the claim the test docstring just corrected. Redshift renders `~`, not `REGEXP_LIKE`, and we now know the earlier `^\d$` revision failed because the backslash was eaten — the override meant the base fix never reached Redshift at all. Worth updating so the body doesn't outlive the correction; it's the section a future reader will trust.

## 2. BigQuery is the same one-liner as the other four

The body says converting it "needs BigQuery-specific quoting (raw/triple-quoted) rather than the shared `literal_string()`" — but `BigQuerySqlDialect` already overrides `literal_string()` (`bigquery_data_source.py:374`) to emit `'''…'''` with `\\` and `\'`:

- `^1\.5$` → `'''^1\\.5$'''` → BigQuery unescapes → `^1\.5$`, identical to what the `r''` form produces today
- `^it's$` → `'''^it\'s$'''` → `^it's$`, which closes the gap

So it's `REGEXP_CONTAINS({expression}, {self.literal_string(matches.regex_pattern)})`, and the bigquery job already runs `test_invalid_count_valid_regex_escaped_metacharacter`, so it validates itself. Fine as a follow-up, but I wouldn't carry it as a documented gap on that reasoning.

Related, and the reason I'd rather see BigQuery converted than deferred: the apostrophe path is unit-only right now. No integration test puts a quote in a pattern, so Databricks' `\'` form and Redshift's `''` form are never exercised against a live engine. Converting BigQuery is what unblocks adding an apostrophe row to the integration test and covering every dialect at once.

m1n0:
Both addressed — thanks, and you were right that I shouldn't have carried BigQuery as a gap.

**BigQuery converted (530491b).** I'd claimed it needed BigQuery-specific quoting rather than the shared `literal_string()`. Wrong: `BigQuerySqlDialect` already overrides `literal_string()` at :374 with the triple-quoted form. I'd grepped `escape_string`, not `literal_string`, and concluded from the miss. Verified the rendering:

- `^1\.5$` → `REGEXP_CONTAINS(`c`, '''^1\\\\.5$''')` → unescapes to `^1\.5$` — byte-identical to what `r''` delivered, so no behaviour change on backslashes
- `^it's$` → `REGEXP_CONTAINS(`c`, '''^it\\'s$''')` → `^it's$` — gap closed

**And the apostrophe integration test you tied to it.** `test_invalid_count_valid_regex_apostrophe` runs `^it's$` against rows `it's` / `its` on every data source, so `''` and `\'` are now exercised live rather than unit-only. That was the actual argument for not deferring, and it only works because BigQuery is converted.

That leaves **sqlserver as the only dialect overriding `_build_regex_like_sql`**, and for a real reason — it rewrites the pattern text before matching. #2843 has been rebased and now folds BigQuery into the syntax-only seam too, so the seam covers redshift/postgres/duckdb/bigquery and sqlserver is the sole documented escape hatch.

**Description contradiction fixed.** The "why it went unnoticed" section no longer claims Redshift is POSIX `REGEXP_LIKE`, and now says plainly that the earlier `^\d$` revision failed because the override meant the base fix never reached Redshift — not because of `\d` semantics. You were right that it's the section a future reader trusts; it had outlived the correction by two commits. I also rewrote the body wholesale rather than patching it further, since it had grown past GitHub's size limit.

#2842 regenerated on the new #2843 head, so the chain is consistent.

PR #2842, opened 2026-08-25
chore(lint): stop isort wrapping imports at 88 when black allows 120

m1n0:
Chain: #2841 → #2843 → **this**. Diff is against #2843.

`[tool.isort] profile = "black"` pins isort's `line_length` to isort's own default of 88;
it does not read `[tool.black] line-length = 120`. So a 100-char import line passes black
locally and fails the pre-commit CI job with a diff that reads as noise.

Sets `line_length = 120` and reformats to match. 140 files, **imports only** — every
changed line is an import statement, a bare imported name, a parenthesis or a blank.

No code lines change: `line_length` under `[tool.isort]` configures isort alone, and isort
only rewrites import blocks. Black's limit doesn't move, and formatters never expand lines
to fill a limit — raising one only un-wraps what the old limit split. `sql_dialect.py` on
main already carries 229 code lines between 89 and 120 chars, which couldn't exist if black
were running at 88.

`origin/main` is 0-files-dirty under `pre-commit run --all-files`, so none of the churn is
pre-existing drift. After the reformat pre-commit converges clean.

Last in the chain deliberately: keeps the two code PRs free of import churn, and picks up
their imports automatically. Regenerated on top of #2843 rather than rebased.

Reviewer notes: any open PR touching an import block will conflict, so worth merging when
the queue is quiet. `git blame --ignore-rev` on the merge commit hides the noise. The
comment above `line_length` is there so the redundant-looking setting isn't tidied away
later, silently restoring the 88/120 split.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,10 @@ build-backend = "setuptools.build_meta"
 
 [tool.isort]
 profile = "black"
+# profile = "black" sets line_length = 88 and does NOT read [tool.black]
+# line-length below, so without this isort wraps imports at 88 while black is
+# happy up to 120 -- a 100-char import line passes black and fails isort.
+line_length = 120
 
 [tool.black]
 line-length = 120
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source.py
@@ -5,12 +5,8 @@ from datetime import datetime
 from typing import Any, Optional, Tuple
 
 import boto3
-from soda_athena.common.data_sources.athena_data_source_connection import (
-    AthenaDataSource as AthenaDataSourceModel,
-)
-from soda_athena.common.data_sources.athena_data_source_connection import (
-    AthenaDataSourceConnection,
-)
+from soda_athena.common.data_sources.athena_data_source_connection import AthenaDataSource as AthenaDataSourceModel
+from soda_athena.common.data_sources.athena_data_source_connection import AthenaDataSourceConnection
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.data_source_impl import DataSourceImpl, FullyQualifiedTableName
 from soda_core.common.logging_constants import soda_logger
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source_connection.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source_connection.py
@@ -12,9 +12,7 @@ from soda_core.common.aws_credentials import AwsCredentials
 from soda_core.common.data_source_connection import DataSourceConnection
 from soda_core.common.logging_constants import soda_logger
 from soda_core.model.data_source.data_source import DataSourceBase
-from soda_core.model.data_source.data_source_connection_properties import (
-    DataSourceConnectionProperties,
-)
+from soda_core.model.data_source.data_source_connection_properties import DataSourceConnectionProperties
 
 logger: logging.Logger = soda_logger
 
--- a/soda-athena/tests/unit/test_athena_catalog_with_slash.py
+++ b/soda-athena/tests/unit/test_athena_catalog_with_slash.py
@@ -11,10 +11,7 @@ affected paths: contract verify queries, column metadata queries, and DDL.
 
 from __future__ import annotations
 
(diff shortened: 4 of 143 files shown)


m1n0:

Niels-b (approved):
Approving the approach — the diagnosis is right and the config change is correct: `line_length` under `[tool.isort]` overrides the black profile's 88 regardless of key order, since isort merges `{**profile, **config_settings}`.

I verified the imports-only claim two ways: line classification of all 1,340 changed lines gives 0 non-import lines with `pyproject.toml` as the only non-`.py` file; and an AST pass that strips every `Import`/`ImportFrom` line range leaves a byte-identical remainder in all 139 `.py` files. Import symbol sets and source order are identical everywhere, `noqa` balance is 2↔2, and the reformat is complete — 0 import lines over 120 and 0 collapsible blocks remain across all 373 files.

Two remarks, the first worth fixing before merge.

**A. `soda-redshift/pyproject.toml` still pins isort to 88.** It's the only sub-package carrying its own `[tool.isort] profile = "black"` (checked all 15 pyprojects at the head), it has no `line_length`, and this PR doesn't touch it — while unwrapping 10 redshift import lines to 92–110 chars:

- `soda-redshift/src/soda_redshift/statements/redshift_metadata_primary_keys_query.py:5` (108)
- `soda-redshift/src/soda_redshift/common/data_sources/redshift_data_source.py:25-26` (106 each)
- `soda-redshift/tests/unit/test_redshift_dialect.py:93,112` (110 each)

Repo-wide pre-commit stays green only because isort's hook is `require_serial: true` — one invocation whose `argv[0]` is a `soda-athena/…` path, so the root config wins. But isort derives `settings_path` from the first file argument and walks up to the nearest pyproject with a non-empty `[tool.isort]`. So a commit staging only redshift files, or `isort soda-redshift/`, or IDE-on-save, re-wraps those 10 lines at 88 — and the next repo-wide run unwraps them again. That's the exact noise-diff ping-pong this PR exists to remove.

Either add `line_length = 120` there, or delete the duplicated `[tool.isort]`/`[tool.black]` block so redshift inherits the root. (soda-extensions#587 needs nothing here — root is the only `[tool.isort]` among its 18 pyprojects.)

**B. The new comment's premise is wrong, and the comment is load-bearing.** It says `profile = "black"` sets "isort's own default of 88" — isort's own default is 79 (`isort/settings.py`); 88 comes from the black profile (`isort/profiles.py`). Since the comment exists precisely so nobody tidies the setting away as redundant, a checkably-wrong premise invites exactly that. Suggest: *"`profile = "black"` sets `line_length = 88` and does NOT read `[tool.black] line-length` below."*

Also, while you're in here: the body recommends `git blame --ignore-rev`, but the repo has no `.git-blame-ignore-revs` and no `blame.ignoreRevsFile`. Worth adding as a post-merge commit in both repos, otherwise every future reader pays that cost by hand.

m1n0:
Both fixed, and the redshift one was a real trap — thanks.

**A. `soda-redshift/pyproject.toml`** — confirmed it's the only sub-package carrying `[tool.isort]` (checked all 15). I **deleted** the duplicated `[tool.isort]`/`[tool.black]` block rather than adding `line_length` to it, so there's one source of truth instead of two that can drift. Verified the fix the way you described the failure: with the pinned isort 5.12.0, `isort --check-only soda-redshift/` on its own now exits 0, where before it would have re-wrapped those 10 lines at 88.

(Worth knowing: my first check *did* report an error — that was uvx pulling isort 8.0.1, which formats differently again. Pinning to 5.12.0 clears it. Unrelated to this PR, but a future bump will not be free.)

**B. Comment premise corrected.** You're right and I checked it: `isort.settings._DEFAULT_SETTINGS['line_length']` is 79, `isort.profiles.black['line_length']` is 88. Now reads *"profile = \"black\" sets line_length = 88 and does NOT read [tool.black] line-length below"*. A comment whose job is to stop someone deleting the setting shouldn't have a falsifiable premise in it.

Also rebased onto the updated #2843, which as a side effect fixes soda-extensions#587: that branch shares this one's name, the resolver auto-matched it, and it was 4 commits behind `main`. Now it carries `main`, so `supported_check_types` is present.

`.git-blame-ignore-revs` — agreed, but the rev doesn't exist until merge. I'll add it in both repos as a post-merge commit rather than guess.

PR #2843, opened 2026-08-25
refactor(sql-dialect): give REGEX_LIKE a syntax-only override seam

m1n0:
Chain: #2841 → **this** → #2842. Diff is against #2841.

#2841 fixed the same one-line bug in six places, which is the tell: the escaping was an
invariant every dialect could re-implement, and five re-implemented it wrong.

Cause is one field — `REGEX_LIKE.regex_pattern` is a bare `str`, so it never passes through
`build_expression_sql`. `LIKE`/`NOT_LIKE` extend `Operator(left, right)` and render both
sides through it, so their pattern is a `LITERAL` and gets escaped in every dialect for
free; `LIKE` has never had this bug. Every other bare-`str` field in the AST is an
identifier or the `SqlExpressionStr` escape hatch — `regex_pattern` is the only one
carrying user data.

Split the two decisions the builder conflated:

```python
def _build_regex_like_sql(self, matches):
    expression = self.build_expression_sql(matches.expression)
    pattern = self.literal_string(matches.regex_pattern)   # once, here
    return self._regex_like_sql(expression, pattern)

def _regex_like_sql(self, expression, pattern):            # dialects override this
    return f"REGEXP_LIKE({expression}, {pattern})"
```

redshift, postgres, duckdb and bigquery become one-liners that cannot forget to escape.
`test_regex_like_seam_escapes_for_dialects_that_only_change_syntax` pins it for future
dialects. sqlserver keeps the full override — it rewrites the pattern text (`a-z`
expansion) and renders `PATINDEX(…) > 0`.

A keyword table wouldn't cover this: two dialects use an infix operator and one isn't a
regex call.

Rejected: typing `regex_pattern` as a `LITERAL` so escaping is inherited structurally.
Cleaner, but soda-extensions reads it as a `str` (`soda-db2` and `soda-dremio` interpolate
it, `soda-db2z` calls `.replace` on it), so it's a cross-repo break needing a coordinated
pin bump. Worth revisiting once those overrides are deleted.

No behaviour change: rendered SQL is identical per dialect, asserted by the per-dialect
tests from #2841.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -23,7 +23,6 @@ from soda_core.common.sql_ast import (
     LITERAL,
     PERCENTILE_WITHIN_GROUP,
     RANDOM,
-    REGEX_LIKE,
     STRING_HASH,
     TIME_DELTA,
     TUPLE,
@@ -262,13 +261,10 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
     def _build_tuple_sql_in_distinct(self, tuple: TUPLE) -> str:
         return f"TO_JSON_STRING(STRUCT({super()._build_tuple_sql(tuple)}))"
 
-    def _build_regex_like_sql(self, matches: REGEX_LIKE) -> str:
-        expression: str = self.build_expression_sql(matches.expression)
-        # literal_string() below is BigQuery's own triple-quoted form, which escapes
-        # the backslash as \\ -- so the pattern reaches the regex engine exactly as the
-        # r'' prefix used to deliver it, and an apostrophe is escaped as \' instead of
-        # terminating the literal. See SCS-1413.
-        return f"REGEXP_CONTAINS({expression}, {self.literal_string(matches.regex_pattern)})"
+    def _regex_like_sql(self, expression: str, pattern: str) -> str:
+        # `pattern` arrives quoted by this dialect's own literal_string(), the
+        # triple-quoted form that escapes the backslash as \\ and an apostrophe as \'.
+        return f"REGEXP_CONTAINS({expression}, {pattern})"
 
     def _build_percentile_within_group_sql(self, percentile_within_group: PERCENTILE_WITHIN_GROUP) -> str:
         """BigQuery has no ordered-set aggregates; render as
--- a/soda-core/src/soda_core/common/sql_dialect.py
+++ b/soda-core/src/soda_core/common/sql_dialect.py
@@ -7,7 +7,7 @@ from abc import abstractmethod
 from datetime import date, datetime, time, timezone
 from numbers import Number
 from textwrap import indent
-from typing import Any, ClassVar, Optional, Tuple
+from typing import Any, ClassVar, Optional, Tuple, final
 
 from soda_core.common.data_source_results import QueryResult
 from soda_core.common.dataset_identifier import DatasetIdentifier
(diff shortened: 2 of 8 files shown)


m1n0:

Niels-b (approved):
Approving — I verified this is a true no-op. I fetched every changed file at `main` and at the head and diffed the trees: `sqlserver_data_source.py` is byte-identical to `main`, and bigquery/duckdb/postgres/redshift each already called `self.literal_string(matches.regex_pattern)`, so rendered SQL is unchanged per dialect. `literal_string` resolves on the instance, so BigQuery's triple-quoted form and Redshift/Snowflake's backslash-doubling `escape_string` still apply through the base method. `escape_regex` has zero callers in either repo. Nice cleanup.

Heads-up before merge: this is `CONFLICTING` against `main`, because #2841 was squash-merged so its commits aren't ancestors of `main`. GitHub picks a pre-#2841 merge base and shows 13 files / +282 folding in already-merged work, when the real net change is 6 files — the sqlserver hunk in particular reads as new. None of the 4 commits `main` is ahead by add a regex override, so the rebase should be purely textual.

Three non-blocking notes.

**1. The seam doesn't quite make the invariant unforgettable, which is the thesis.** `_build_regex_like_sql` and `_regex_like_sql` differ only by the `_build` prefix, and `_build_X_sql` is the convention for every other builder in the class — so the familiar-looking name is the unsafe one, with no `@final` and no guard. sqlserver here, and `Db2ZosSqlDialect` in soda-extensions, both still take it.

A `_rewrite_regex_pattern(pattern) -> str` hook would close it. sqlserver's `a-z`/`A-Z` expansion plus the `%…%` wrap fits there, and the argument order is no obstacle:

```python
def _regex_like_sql(self, expression: str, pattern: str) -> str:
    return f"PATINDEX ({pattern}, {expression} COLLATE SQL_Latin1_General_Cp1_CS_AS) > 0"
```

Then no dialect needs the outer method and it can be `@final` — enforced rather than documented. db2z collapses onto the seam too.

**2. The `_regex_like_sql` docstring is wrong about bigquery.** It says to override `_build_regex_like_sql` "only when the pattern text has to be rewritten (sqlserver) or needs a non-standard literal form (bigquery's raw strings)" — but in this PR bigquery overrides `_regex_like_sql`; its non-standard literal form comes from its own `literal_string()`, not from holding the outer method. db2z is also a third category (whole-predicate rewrite) the guidance doesn't name. Worth fixing since this docstring is the one artifact the next dialect author reads.

**3. sqlserver — the only dialect the seam doesn't protect — has no regex coverage at all.** `test_sqlserver_dialect.py` is 252 lines / 20 tests with no regex or PATINDEX case, and both new integration tests gate on `supports_regex_advanced()`, which sqlserver returns `False` for, so both skip. The skip is *correct* there (PATINDEX reads `^` literally, so `^it's$` wouldn't match `it's` and the assertion would be wrong) — so a unit test is the right shape:

```python
assert dialect.build_expression_sql(REGEX_LIKE(COLUMN("c"), "^it's$")) == (
    """PATINDEX ('%^it''s$%', "c" COLLATE SQL_Latin1_General_Cp1_CS_AS) > 0"""
)
```

m1n0:
Rebased onto `main` — the conflict was exactly what you diagnosed, #2841's commits aren't ancestors so I re-cut from `main` and cherry-picked the one seam commit. Net change is 6 files as you predicted, now 8 with the added tests. All three notes taken.

**1. The seam now actually enforces it.** You were right that `_regex_like_sql` vs `_build_regex_like_sql` differing by a prefix makes the familiar name the unsafe one. Added `_rewrite_regex_pattern(pattern) -> str`, which runs *before* the escaping, and marked `_build_regex_like_sql` `@final`.

sqlserver moves onto both seams — the `a-z`/`A-Z` expansion and the `%…%` wrap go in the rewrite hook, the `PATINDEX(...) > 0` shape in `_regex_like_sql`. Rendering is unchanged; `^it's$` still gives `PATINDEX ('%^it''s$%', [c] COLLATE …) > 0`. So **no dialect overrides the builder any more**, which is what makes `@final` enforcement rather than documentation.

**2. Docstring fixed.** It no longer claims bigquery needs the outer method — its non-standard literal form comes from its own `literal_string()`, which is a different concern. It now names the three shapes actually in use: function call, infix operator, and whole-predicate rewrite (sqlserver, and db2z in extensions).

**3. sqlserver regex coverage added** — your assertion, with `[c]` since that's how sqlserver quotes:

```python
assert dialect.build_expression_sql(REGEX_LIKE(COLUMN("c"), "^it's$")) == (
    "PATINDEX ('%^it''s$%', [c] COLLATE SQL_Latin1_General_Cp1_CS_AS) > 0"
)
```

Plus the `a-z` expansion case, and a guard that a dialect overriding only the rewrite hook still gets the pattern escaped — the other half of why the builder can be final.

1586 unit + integration tests pass, pre-commit clean.

PR #2847, opened 2026-08-27
feat(mysql): dialect seams and shared-suite support for MySQL

Niels-b:
## Description

The soda-core half of MySQL support (`R-647`). The adapter itself lands in
[soda-extensions][ext-pr] — this PR is the shared-code seams it needs, plus two
fixes it surfaced. Five commits, each reviewable on its own.

### `COMBINED_HASH` can group values the data source considers equal

A duplicate check's verdict is computed by the data source: `COUNT(DISTINCT col)`
compares under the column's collation. On MySQL's default `utf8mb4_0900_ai_ci`
that folds case **and** accents, while `COMBINED_HASH` hashes bytes — so the
failed-rows query, which hashes the same way, could not return the rows the
verdict had counted. For a **multi-column** check the verdict *is* the hash, so
the two check shapes disagreed about the same data:

| check | verdict source | byte-exact hash |
| --- | --- | --- |
| one column | `COUNT(DISTINCT col)` — folds | sample can't be returned |
| two+ columns | `COUNT(DISTINCT <hash>)` | **verdict itself is wrong** |

New opt-in `fold_equal_values` flag, honoured by overriding the new per-operand
`_build_hash_operand` hook. The base ignores it. Row identity (`__soda_row_id`)
and the diagnostics-warehouse keys must stay byte-exact, so only the duplicate
path opts in.

**No other dialect changes.** I restored the pre-flag builder from `main` and
compared rendered SQL for 18 dialects across both repos × 5 operand shapes ×
both flag values: **175 identical, 5 differing — all 5 MySQL with the flag on.**

### `normalizes_own_text_ordering`

Reconciliation compensates for a case-insensitive peer by folding the other side
with `LOWER()`. That cannot reproduce an *accent*-insensitive order — `LOWER()`
drops case but keeps accents — so `café` sorts on the wrong side of `cafz` and a
cross-source `rows_diff` reports diffs that aren't there. A source answering
`True` folds its own keys instead. Defaults `False`; no existing source changes.

### `PicklableColumn` keeps a driver's 8th description field

PEP 249 fixes the first seven entries; a driver may append its own. MySQL puts
the charset id there, the only thing distinguishing `BINARY` from `CHAR` (each
pair shares one field-type code). Added with a default, so existing pickled
snapshots still load.

### Reference-check join alias is quoted by the dialect — independent fix

`InvalidReferenceCountQuery.query_join` hardcoded `f'"{alias}".{column}'`. On
BigQuery and Databricks a double-quoted identifier is a **string literal**, so
`"C".country` was a syntax error: any contract combining `column_expression`
with `valid_reference_data` could not run at all there. SQL Server needs
brackets. Unrelated to MySQL — worth reviewing (or reverting) on its own.

## Impact on downstream packages

soda-extensions consumes all four seams; its PR is stacked on this branch via
`SODA_CORE_BRANCH`. No behaviour change for any existing data source — the
byte-identity comparison above is the evidence for the hash, and the other three
are additive (new hook, new default-`False` property, defaulted tuple field).

## Tests

Two new unit files, both independent of a live server:

- `test_combined_hash_seam.py` — pins the **exact pre-flag SQL** for all 13
  dialects shipped here, including their pre-existing quirks (BigQuery
  triple-quotes the sentinel, Redshift concatenates with `||`), reproduced not
  corrected. A shared-builder change moves every dialect together, so the
  previous same-dialect comparison was blind to it: under a mutation swapping
  the NULL sentinel, the old style of assertion passes 26/26 while these fail 16.
  Also that the flag stays inert per dialect, and that the multi-column verdict
  asks for the fold.
- `test_invalid_reference_alias_quoting.py` — drives the real `query_join`
  across the four quoting styles. Previously zero coverage: the only test
  reaching that branch is the feature suite, which runs on the postgres lane —
  the one dialect whose output is unchanged.

| suite | result |
| --- | --- |
| `TEST_DATASOURCE=mysql` | 1502 passed, 35 skipped |
| `TEST_DATASOURCE=postgres` | 1510 passed, 27 skipped |
| `TEST_DATASOURCE=duckdb` | 1498 passed, 35 skipped |
| `TEST_DATASOURCE=sqlserver` (integration) | 172 passed, 16 skipped |

## Checklist

- [x] I verified this PR does not break [soda-extensions][1] — its suites run
      green against this branch locally (packages 775 passed; `all_data_sources`
      149/151 in both transfer modes; cross-source mysql↔postgres/sqlserver 16/16
      live). The extensions PR pins `SODA_CORE_BRANCH=r-647-mysql-support`.

[ext-pr]: https://github.com/sodadata/soda-extensions

### DIFF
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -210,20 +210,23 @@ class DataSourceImpl(ABC):
         return self._can_create_materialized_view
 
     # The capability hooks below are all default-off / behavior-preserving. They are consumed by
-    # soda-reconciliation's cross-source reconciliation (rows_diff / reference_diff) and overridden
-    # only by soda-salesforce, letting a SOQL-first / case-insensitive source opt into the behavior
-    # cross-source recon needs without affecting any conventional SQL source. (get_value_comparator
-    # returns a ValueComparatorProtocol; the others are booleans.)
+    # soda-reconciliation's cross-source reconciliation (rows_diff / reference_diff), and let a
+    # source whose ordering or comparison differs from plain codepoint SQL opt into the behavior
+    # cross-source recon needs without affecting any conventional SQL source. Overridden by
+    # soda-salesforce (SOQL-first, case-insensitive) and soda-mysql (accent-insensitive ordering).
+    # (get_value_comparator returns a ValueComparatorProtocol; the others are booleans.)
     @property
     def normalizes_own_text_ordering(self) -> bool:
-        """Whether this data source must fold its OWN text keys to become codepoint-comparable.
+        """Whether reconciliation should fold THIS source's text keys rather than the peer's.
 
-        Default False. ``orders_text_case_insensitively`` says "my ordering differs, fold the other
-        side to match me" — which only works when the peer's fold can reproduce this source's
-        order. It cannot when the ordering is accent-insensitive: the fold is ``LOWER()``, which
-        drops case but keeps accents, so `café` still sorts on the wrong side of `cafz`. A source in
-        that position returns True here instead, and the merge-join folds *this* side — through its
-        own ``order_by_key_expression`` — leaving the peer's native codepoint order untouched.
+        Cross-source reconciliation merge-joins two streams, so both sides must arrive in the same
+        order. Default False: the peer is folded to match this source, using ``LOWER()``.
+
+        Return True when ``LOWER()`` cannot reproduce this source's order. It only removes case, so
+        a source that also ignores accents orders ``café`` before ``cafz`` while a lowered peer
+        orders it after — and the join reports differences that do not exist. Folding this source
+        instead, through its own ``order_by_key_expression``, gives both sides the peer's plain
+        codepoint order.
         """
         return False
 
--- a/soda-core/src/soda_core/common/data_source_impl.py
+++ b/soda-core/src/soda_core/common/data_source_impl.py
@@ -215,6 +215,19 @@ class DataSourceImpl(ABC):
     # cross-source recon needs without affecting any conventional SQL source. (get_value_comparator
     # returns a ValueComparatorProtocol; the others are booleans.)
     @property
+    def normalizes_own_text_ordering(self) -> bool:
+        """Whether this data source must fold its OWN text keys to become codepoint-comparable.
+
+        Default False. ``orders_text_case_insensitively`` says "my ordering differs, fold the other
+        side to match me" — which only works when the peer's fold can reproduce this source's
+        order. It cannot when the ordering is accent-insensitive: the fold is ``LOWER()``, which
+        drops case but keeps accents, so `café` still sorts on the wrong side of `cafz`. A source in
+        that position returns True here instead, and the merge-join folds *this* side — through its
+        own ``order_by_key_expression`` — leaving the peer's native codepoint order untouched.
+        """
+        return False
+
+    @property
     def orders_text_case_insensitively(self) -> bool:
         """Whether this data source's SQL orders text columns case-insensitively.
 
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -308,6 +308,27 @@ class TUPLE(SqlExpression):
 @dataclass
 class COMBINED_HASH(SqlExpression):
     expressions: list[SqlExpression | str]
+    # Whether the hash should treat two values the data source considers EQUAL as one, rather
+    # than hashing their bytes. Default-off, so every existing caller renders byte-for-byte
+    # unchanged on every dialect.
+    #
+    # The two uses of this node need opposite answers, which is why it is a flag rather than a
+    # dialect-wide choice:
+    #   - duplicate equivalence (the duplicate check's verdict, and the failed-rows query that
+    #     has to return the very rows that verdict counted) must agree with what the data source
+    #     considers a duplicate. On a case/accent-insensitive collation, 'ABC' and 'abc' ARE one
+    #     value, so hashing their bytes made the check report duplicates it could not then show.
+    #   - row identity (`__soda_row_id`) and diagnostics-warehouse keys must stay byte-exact:
+    #     folding there would give two genuinely different rows the same id.
+    #
+    # Byte-exactness is what row identity REQUIRES; it is not on its own enough to make the hash
+    # collision-free, and this flag does not change that either way. Two pre-existing sources of
+    # collision in the base rendering, both measured on MySQL 8.4:
+    #   - the CONCAT_WS separator is not escaped, so ('P','Q||R') and ('P||Q','R') are two rows
+    #     the server calls distinct and one hash group;
+    #   - CAST(x AS VARCHAR) yields NULL for bytes that are not valid text, so four distinct
+    #     VARBINARY values collapsed to two groups — one of them shared with a genuine NULL key.
+    fold_equal_values: bool = False
 
     def __post_init__(self):
         super().__post_init__()
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -4,7 +4,7 @@ import logging
 import weakref
 from dataclasses import dataclass, field
 from numbers import Number
-from typing import Any, Optional, Union
+from typing import Any, Literal, Optional, Union
 
 from soda_core.common.logging_constants import soda_logger
 from soda_core.common.metadata_types import SamplerType, SodaDataTypeName, SqlDataType
@@ -305,18 +305,25 @@ class TUPLE(SqlExpression):
         self.handle_parent_node_update(self.expressions)
 
 
+# Whose equality a hash agrees with. "exact" hashes the bytes; "database-equality" hashes so that
+# two values the data source itself calls equal hash alike — on a case- or accent-insensitive
+# collation those are different answers. Named for the equality, not for a transformation: nothing
+# is normalised, the dialect hashes the collation's own sort key.
+ComparisonMode = Literal["exact", "database-equality"]
+
+
 @dataclass
 class COMBINED_HASH(SqlExpression):
     expressions: list[SqlExpression | str]
-    # When True, two values the data source's own equality considers equal hash alike; when False
-    # their bytes are hashed. Default False, so every existing caller is unchanged.
+    # Whose definition of "equal" the hash agrees with. Defaults to "exact", so every existing
+    # caller is unchanged.
     #
-    # A flag rather than a per-dialect choice because the two callers need opposite answers:
+    # Per-node rather than per-dialect because the two callers need opposite answers:
     #   - a duplicate check must group the way the data source does, or the failed-rows query
     #     cannot return the rows the check's own verdict counted;
     #   - row identity (`__soda_row_id`) and the diagnostics-warehouse keys must stay byte-exact,
     #     or two different rows are given the same id.
-    fold_equal_values: bool = False
+    comparison_mode: ComparisonMode = "exact"
 
     def __post_init__(self):
         super().__post_init__()
--- a/soda-core/src/soda_core/common/sql_ast.py
+++ b/soda-core/src/soda_core/common/sql_ast.py
@@ -308,26 +308,14 @@ class TUPLE(SqlExpression):
 @dataclass
 class COMBINED_HASH(SqlExpression):
     expressions: list[SqlExpression | str]
-    # Whether the hash should treat two values the data source considers EQUAL as one, rather
-    # than hashing their bytes. Default-off, so every existing caller renders byte-for-byte
-    # unchanged on every dialect.
+    # When True, two values the data source's own equality considers equal hash alike; when False
+    # their bytes are hashed. Default False, so every existing caller is unchanged.
     #
-    # The two uses of this node need opposite answers, which is why it is a flag rather than a
-    # dialect-wide choice:
-    #   - duplicate equivalence (the duplicate check's verdict, and the failed-rows query that
-    #     has to return the very rows that verdict counted) must agree with what the data source
-    #     considers a duplicate. On a case/accent-insensitive collation, 'ABC' and 'abc' ARE one
-    #     value, so hashing their bytes made the check report duplicates it could not then show.
-    #   - row identity (`__soda_row_id`) and diagnostics-warehouse keys must stay byte-exact:
-    #     folding there would give two genuinely different rows the same id.
-    #
-    # Byte-exactness is what row identity REQUIRES; it is not on its own enough to make the hash
-    # collision-free, and this flag does not change that either way. Two pre-existing sources of
-    # collision in the base rendering, both measured on MySQL 8.4:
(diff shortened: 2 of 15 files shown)


soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -295,6 +295,27 @@ def __post_init__(self):
...
+    #     folding there would give two genuinely different rows the same id.
+    #
+    # Byte-exactness is what row identity REQUIRES; it is not on its own enough to make the hash
+    # collision-free, and this flag does not change that either way. Two pre-existing sources of
+    # collision in the base rendering, both measured on MySQL 8.4:
+    #   - the CONCAT_WS separator is not escaped, so ('P','Q||R') and ('P||Q','R') are two rows
+    #     the server calls distinct and one hash group;
+    #   - CAST(x AS VARCHAR) yields NULL for bytes that are not valid text, so four distinct
+    #     VARBINARY values collapsed to two groups — one of them shared with a genuine NULL key.
+    fold_equal_values: bool = False
```
paulteehan: interesting ...  so I guess for Salesforce this is True because abc and ABC are considered equivalent and need to map to the same hash?

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -214,6 +214,19 @@ def can_create_materialized_view(self) -> bool:
     # only by soda-salesforce, letting a SOQL-first / case-insensitive source opt into the behavior
     # cross-source recon needs without affecting any conventional SQL source. (get_value_comparator
     # returns a ValueComparatorProtocol; the others are booleans.)
+    @property
```
mivds, line 218: Comment above is no longer accurate

soda-core/src/soda_core/common/data_source_impl.py
```diff
@@ -214,6 +214,19 @@ def can_create_materialized_view(self) -> bool:
...
+    def normalizes_own_text_ordering(self) -> bool:
+        """Whether this data source must fold its OWN text keys to become codepoint-comparable.
+
+        Default False. ``orders_text_case_insensitively`` says "my ordering differs, fold the other
+        side to match me" — which only works when the peer's fold can reproduce this source's
+        order. It cannot when the ordering is accent-insensitive: the fold is ``LOWER()``, which
+        drops case but keeps accents, so `café` still sorts on the wrong side of `cafz`. A source in
+        that position returns True here instead, and the merge-join folds *this* side — through its
+        own ``order_by_key_expression`` — leaving the peer's native codepoint order untouched.
```
mivds: Can we clarify this to something understandable? I don't get it 😅

soda-core/src/soda_core/common/sql_ast.py
```diff
@@ -295,6 +295,27 @@ def __post_init__(self):
...
+    #   - row identity (`__soda_row_id`) and diagnostics-warehouse keys must stay byte-exact:
+    #     folding there would give two genuinely different rows the same id.
+    #
+    # Byte-exactness is what row identity REQUIRES; it is not on its own enough to make the hash
+    # collision-free, and this flag does not change that either way. Two pre-existing sources of
+    # collision in the base rendering, both measured on MySQL 8.4:
+    #   - the CONCAT_WS separator is not escaped, so ('P','Q||R') and ('P||Q','R') are two rows
+    #     the server calls distinct and one hash group;
+    #   - CAST(x AS VARCHAR) yields NULL for bytes that are not valid text, so four distinct
+    #     VARBINARY values collapsed to two groups — one of them shared with a genuine NULL key.
```
mivds: This sounds like convoluted bug description getting fixed in this PR. Suggest removing it, or at the very least translating this to something humans can understand

soda-core/src/soda_core/common/sql_dialect.py
```diff
@@ -1448,6 +1458,25 @@ def supports_row_sampling(self) -> bool:
         """
         return True
 
+    def returns_native_boolean_values(self) -> bool:
+        """Whether this source's driver hands back a real ``bool`` for a canonically BOOLEAN column.
+
+        True for every dialect whose driver does. MySQL is the exception: its BOOLEAN is an alias
+        for TINYINT(1) and PyMySQL returns ``1``/``0``, so a consumer testing ``value is True``
```
mivds: Can't we handle this at the connection level? Potentially a normalizing wrapper around the the data returned by the driver? That would be simpler than having that complexity impact the higher layers.

From the description we're also encoding a driver-specific setting in the dialect, which is design-wise not correct. If we swap to a different MySQL driver that does properly return `bool`, the dialect would have to change.

soda-tests/tests/feature/test_column_expression.py
```diff
@@ -1,9 +1,31 @@
...
 from helpers.mock_soda_cloud import MockResponse
+from helpers.test_fixtures import test_datasource
 from helpers.test_table import TestTableSpecification
 from soda_core.contracts.contract_verification import CheckOutcome
 
+# column_expression carries customer-supplied SQL, so these fixtures are inherently
+# dialect-flavoured. The values below are the Postgres/DuckDB spellings the file was written
+# with; a dialect that spells casts or JSON access differently overrides them rather than
+# skipping the file. Plain constants, not a dict, so the contract f-strings need no subscript.
+if test_datasource == "mysql":
```
mivds: Can't we use the dialect to generate this, rather than hard-coding here? It's also leaking `mysql` stuff in places it doesn't belong

soda-tests/tests/unit/test_combined_hash_seam.py
mivds, line 1: Wouldn't this be more suited as a dialect test in the respective data source packages? I see the appeal of having a central place to test `COMBINED_HASH`, but I'm not sure grouping that way is the best option

soda-core/src/soda_core/contracts/impl/check_types/duplicate_check.py
```diff
@@ -344,10 +344,17 @@ def _get_id_properties(self) -> dict[str, any]:
 
     def sql_expression(self) -> SqlExpression:
         filter_expr: SqlExpression = SqlExpressionStr.optional(self.check_filter)
+        # fold_equal_values so a multi-column duplicate check agrees with the single-column one
+        # above, which uses COUNT(DISTINCT(column)) and therefore folds under the data source's
+        # own text comparison. Without it the two answered differently about the same data on a
+        # case-insensitive collation, and the failed-rows query — which hashes the same way —
+        # could not return the rows this count reported. Identity on byte-exact dialects, so
+        # their SQL is unchanged.
+        combined_hash = COMBINED_HASH(self.column_expressions, fold_equal_values=True)
```
paulteehan: I see what you mean about this one being trickier than expected!

soda-tests/src/helpers/snapshot_connection.py
```diff
@@ -21,8 +21,13 @@
 # A picklable replacement for DBAPI cursor.description column entries.
 # DB drivers like psycopg3 use C-extension Column objects that can't be pickled.
 # This namedtuple preserves both attribute access (.name) and index access ([0]).
+# PEP-249 fixes the first seven entries; a driver may append its own. `driver_extra` carries an
+# eighth so a snapshot does not silently drop it — MySQL puts the charset id there, which is what
+# tells a binary column from a text one.
 PicklableColumn = namedtuple(
-    "PicklableColumn", ["name", "type_code", "display_size", "internal_size", "precision", "scale", "null_ok"]
+    "PicklableColumn",
+    ["name", "type_code", "display_size", "internal_size", "precision", "scale", "null_ok", "driver_extra"],
```
paulteehan, line 29: claude warns that once this code goes live, any old code that runs with newly pickled objects will fail, because the new objects will have too many columns.   i don't think is going to actually happen as long as the code is rebased.  if we get weird pickle errors in CI snapshots this might be why.

soda-tests/tests/unit/test_combined_hash_seam.py
Niels-b: I'm a bit on the fence here. Having it unit-testable in soda-core also seems to make sense to me. But having it per dialect seems to be the way we usually do things. 

Splitting this one up is something we can still do as a follow-up (rather keep this PR diff clean). I will note this down in my TODO list

paulteehan (approved):
Straightforward, didn't see any isses myself, though claude found a couple

claude reports there are issues relating to concat seperatiors and hash equivalence in several other dialects which should be fixed, though outside the scope of this PR, @Niels-b maybe you could investigate?  Detail:


> On BigQuery and Redshift the CONCAT_WS/|| separator renders as ''||'', which SQL parses as '' || '' — the empty string — so the pinned goldens in soda-tests/tests/unit/test_combined_hash_seam.py:106-109 and :121-124 show a multi-column hash with no separator at all, meaning ('a','bc') and ('ab','c') collide and a duplicate check reports duplicates that don't exist. Separately, the note at soda-core/src/soda_core/common/sql_dialect.py:1432 records that the SQL Server family folds by default and doesn't override _build_hash_operand, so on its case-insensitive default collation a single-column duplicate check (server verdict, folds) and a multi-column one (hash verdict, byte-exact) disagree about the same data — the same class of bug fold_equal_values fixes for MySQL.

soda-core/src/soda_core/common/sql_ast.py
Niels-b: Good question, and the answer is no — but only because Salesforce can't get here. Its own README says `duplicate_diff` "needs a per-row hash SOQL cannot express", so it never renders a `COMBINED_HASH`, never overrides `_build_hash_operand` and never sees the flag. It does override `orders_text_case_insensitively` and `get_value_comparator`, so your premise about its equality is right; the hash just never happens.

Your question points at something real, though. MySQL is the **only** dialect anywhere that overrides `_build_hash_operand` — and it is not the only case-insensitive engine we support. Our own SQL Server test container, stock:

```
server collation: SQL_Latin1_General_CP1_CI_AS
'abc' = 'ABC' -> EQUAL
```

`CI` is SQL Server's default. So a multi-column duplicate check there looks like it has the same defect MySQL had before this PR: the server's `COUNT(DISTINCT ...)` folds case, the byte hash doesn't, and the check reports duplicates it can't show — or, for a multi-column key, silently reports zero. Synapse and Fabric inherit the same dialect.

I've only verified the collation and the absence of an override, not the end-to-end verdict, and SQL Server is deliberately untouched on this branch — so it's going on a ticket rather than into this PR.

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: Fixed. It listed only soda-salesforce after soda-mysql had started overriding these hooks; it now names both and says what each opts into.

soda-core/src/soda_core/common/data_source_impl.py
Niels-b: Rewritten. The old version explained this hook by quoting a *different* hook's slogan, which meant reading two docstrings to understand one.

It now leads with the mechanism — reconciliation merge-joins two streams, so both must arrive in the same order — then the default, then when to return True and why `LOWER()` isn't enough: it drops case but keeps accents, so `café` lands on the wrong side of `cafz`.

soda-core/src/soda_core/common/sql_ast.py
Niels-b: Cut from 22 lines to 8: what the flag does, and why it is a flag rather than a dialect-wide choice. You were right about what made it convoluted — roughly half of it described two collision bugs that have nothing to do with this flag.

To be clear about what happened to them, since deleting them shouldn't make them invisible: both are **pre-existing and still unfixed**, and are going on a ticket.

- The `CONCAT_WS` separator is not escaped, so `('P','Q||R')` and `('P||Q','R')` are two rows the server calls distinct and one hash group.
- `CAST(x AS VARCHAR)` returns NULL for bytes that aren't valid text, which collapsed four distinct VARBINARY values into two groups — one shared with a genuine NULL key.

Both measured on MySQL 8.4, both affecting every data source.

soda-core/src/soda_core/common/sql_dialect.py
Niels-b: Done, at the cursor, as you suggested — the hook you're commenting on is gone.

`FieldMetadataCursor` now normalises `FIELD_TYPE.TINY` with `get_column_length() == 1` to a real `bool`, reading it off `self._result.fields`. `returns_native_boolean_values` is reverted, so soda-core carries no MySQL-shaped concept, and `BooleanDwhValueMapper` is removed from the extension too (see soda-extensions#592).

One fix did stay in soda-core, and it is not MySQL plumbing: `profiling/boolean.py` tested `row[0] is True`, which is False for **any** driver returning `1`, so the true bucket was silently dropped. Reproduced before any of my changes — `{0: 1, -1: 1}` instead of `{0: 1, 1: 1}`. Fixed unconditionally rather than gated on MySQL, since it's wrong for every such driver.

Worth flagging one limit of doing it at the cursor: it cannot distinguish intent. MySQL sends one field type for `BOOLEAN` and for a genuine `TINYINT(1)`, and the wire carries no "declared BOOLEAN" bit — so a customer's real `TINYINT(1)` column is normalised to bool as well. Unavoidable at the protocol level; say the word if you'd rather have it back as a dialect decision.

soda-tests/tests/feature/test_column_expression.py
Niels-b: Moved. The spellings now live in a `column_expressions` dict on `DataSourceTestHelper` that each data source's helper overrides, and the test file has no `test_datasource` import or conditional left.

Not generated from the dialect, though I tried: `CAST` has an AST node, but JSON path access has none, so three of the five have nothing to generate from. Adding a JSON-path node to soda-core would be the way to close that, and is bigger than this PR.

Your underlying point is the more useful one — that was the third place this PR leaked a `test_datasource == "mysql"` conditional into shared test code. All three went in the same commit.

soda-core/src/soda_core/contracts/impl/check_types/duplicate_check.py
Niels-b: It was. The part that took longest wasn't the folding, it was proving the boundary: which hashes must stay byte-exact (`__soda_row_id`, diagnostics keys — fold them and two different rows get one id) versus which must fold (the duplicate verdict, and the failed-rows query that has to return the rows that verdict counted). There's a test file pinning exactly that boundary, because neither direction is visible in the shape of the SQL.

soda-tests/src/helpers/snapshot_connection.py
Niels-b: Your instinct is right, and it's measurable rather than a hope.

All **329** pickles in the extensions snapshot store load clean under the current definition, and none of them contains a `PicklableColumn` payload at all — they hold `SnapshotEntry` + `QueryResult`, so column descriptions aren't in the recorded set today.

And the failure mode can't occur even when they are, because the field carries a default. A 7-field payload recorded before `driver_extra` existed, loaded against the current 8-field definition:

```
PicklableColumn(name='id', type_code=23, display_size=None, internal_size=4,
                precision=None, scale=None, null_ok=True, driver_extra=None)
```

`defaults=(None,)` fills the eighth. So an unrebased branch reading new snapshots is the only direction that could bite, and that's the direction rebasing already fixes.

PR #2850, opened 2026-09-01
fix(athena): convert tz timestamps in the driver, report unparseable freshness values truthfully

m1n0:
## What

Freshness checks on Athena columns that are `timestamp with time zone` at query time (e.g. Iceberg `timestamptz`) failed with a misleading error:

```
Freshness column 'sys_update_time_utc' returned value 'None' of data type 'NoneType' which is not a datetime or datetime-compatible type.
```

The data was fine — the driver was handing the engine a raw string.

## Root cause

- `MAX()` over such a column has result type `timestamp with time zone`. pyathena only maps that type to a converter since **3.9.0**; on older versions (pinned at 2.25.2 in the contract launcher until v0.1.63) the value falls through to the default converter and reaches the engine as the raw string `'2026-09-01 12:00:00.000 UTC'`.
- The engine's `convert_str_to_datetime` can't parse the zone suffix and returns `None`, and the error message then printed the post-conversion `None` — reading as if the database returned NULL.
- The schema check passes throughout because Athena metadata types are normalized by truncating at the first `(` (`timestamp(6) with time zone` → `timestamp`), so nothing pointed at the type.

## Changes

1. **`LenientTypeConverter` now maps `timestamp with time zone` itself**, so the fix holds on any supported pyathena (`>=2.2.0,<4.0`): named zones, offset zones (pyathena's own converter drops these silently, producing naive datetimes), values with and without fractional seconds. Sub-microsecond fractions (`timestamp(7..9)`) are truncated instead of failing the whole fetch. An unrecognized zone raises a `ValueError` naming the value rather than guessing.
2. **The freshness check reports unparseable strings truthfully**: the error now names the actual received value (`... returned value '2026-09-01 12:00:00.000 UTC' of data type 'str' which could not be parsed as a datetime`) instead of claiming `'None'`/`'NoneType'`.
3. **`soda-athena` now declares `tzdata`**, so region-named zones resolve through `zoneinfo` on images that ship no system IANA database.

## Validation

- New unit tests for both (converter flavors incl. offsets/nanos/unknown zone; freshness message honesty). Full unit suite green (1405 passed).
- Validated against live Athena on pyathena **2.25.2** and **3.35.4**: every timestamp flavor now returns a correct timezone-aware datetime on both.

```customer-facing-release-note
Freshness checks on Athena timestamp-with-time-zone columns now evaluate correctly.
```

https://claude.ai/code/session_01KKHGpDcwQVq3UgxRfxpvfF

### DIFF
--- a/soda-athena/pyproject.toml
+++ b/soda-athena/pyproject.toml
@@ -11,6 +11,8 @@ dependencies = [
     "soda-core==4.23.0",
     "PyAthena>=2.2.0,<4.0",
     "setuptools>=60",
+    # ZoneInfo needs an IANA database; slim images often ship without one.
+    "tzdata",
 ]
 
 [project.entry-points."soda.plugins.data_source.athena"]
--- a/soda-athena/src/soda_athena/common/data_sources/athena_data_source_connection.py
+++ b/soda-athena/src/soda_athena/common/data_sources/athena_data_source_connection.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import logging
+import re
 from abc import ABC
-from datetime import datetime, timezone, tzinfo
+from datetime import datetime, timedelta, timezone, tzinfo
 from typing import Literal, Optional, Union
+from zoneinfo import ZoneInfo
 
 import pyathena
 from pyathena.converter import DefaultTypeConverter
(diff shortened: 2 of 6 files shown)


Niels-b:
We should definitely run the extensions tests (especially cross source) with these changes. There is some risk with changing timestamp handling.

Niels-b (commented):
Thanks for doing this work. Looks good to me 

Claude did flag two interesting things though:
1. an Unknown Timezone will fail the code; if the TZ is unknown, the scan will still fail because we raise `ValueError` for that. 
          Do we want this? Maybe we can default to UTC in that case so the whole scan doesn't fail? No strong opinion here, just flagging this.
2. `tzdata` is not a dependency for Athena, which may cause confusing errors when using `ZoneInfo`.

Also, we should definitely run soda-extensions CI (see earlier comment). Time values are a tricky one 😅

m1n0:
Both addressed:

- tzdata is now a declared dependency of soda-athena (5addb48) — without it, a perfectly resolvable region zone on a slim image would misreport as "unrecognized".
- On the unknown-TZ ValueError: kept the raise, deliberately. With tzdata in place everything Athena can actually render resolves (IANA names, offsets, UTC aliases), so only genuinely corrupt values reach that path — and defaulting those to UTC would silently compute freshness off by the true offset, which is the same failure class this PR removes (pyathena's own tz converter does exactly that for offset zones: gettz('+02:00') → None → naive datetime). If you'd rather degrade than fail there, say so and I'll switch it to warn + UTC.

Extensions CI is running against this branch, all datasources, cross-source live: https://github.com/sodadata/soda-extensions/actions/runs/33620522395

m1n0:
Core CI is fully green (live Athena job included). The first extensions dispatch died on CI infra, not tests — live cross-source starts DB2/Oracle/MySQL/SQL Server on one runner and filled its disk before any test ran (ENOSPC at uv install, all 11 modules alike). Fix: sodadata/soda-extensions#600. Rerun from that branch, all datasources + cross-source live: https://github.com/sodadata/soda-extensions/actions/runs/33622409051

m1n0:
Extensions validation is green: all adapter tests and all 15 cross-source modules, live, against this branch — https://github.com/sodadata/soda-extensions/actions/runs/33622409051. The earlier red attempts were CI-side only: runner disk exhaustion (fixed in sodadata/soda-extensions#600) and the known Fabric shared-warehouse concurrent-DDL flake, which passed once rerun in isolation. Ready to merge from my side.

m1n0:
@Niels-b all addressed 🙏

Niels-b (approved):
Thanks for addressing all these points and fixing the issue in the live extensions run 🙏 .

PR #2851, opened 2026-09-02
feat(review): add the de-review skill and an advisory Claude review workflow

m1n0:
Captures the engine review conventions as a skill the team and CI share — the invariants that must hold, the checks that matter per kind of change, and the failure shapes agent-written PRs produce here. Distilled from this repository's own review history.

Adds:

- `.claude/skills/de-review/SKILL.md` — usable locally as `/de-review`
- `.github/workflows/claude-de-review.yml` — advisory, **does not gate merge**
- a review checklist in the PR template
- a `.gitignore` negation so `.claude/skills/` is tracked while local settings stay ignored

Same shape as the review workflows already running in `sodadata/soda`: label-triggered with `agent-review`, `@claude-de` to re-run, action pinned to the same SHA.

## This repository is public

Two consequences, both handled:

**The review output is world-visible.** The skill carries explicit rules against naming private repositories, internal packages, customers or ticket IDs, and requires every finding to be provable from this repository alone. The workflow prompt repeats them. The generator that produces this file fails the build if a private reference appears in it.

**Fork PRs do not get secrets.** GitHub does not pass secrets to `pull_request` runs from a fork, so the `agent-review` label silently does nothing on roughly half the PRs here — which are exactly the outside contributions this is most useful for. The `@claude-de` comment path covers those, because `issue_comment` runs in the base repo. It is gated to `OWNER`/`MEMBER`/`COLLABORATOR` so an outside contributor cannot trigger a secret-bearing run on their own code, and the job reads the diff only. **It must never gain a build, install or test step.**

Net effect: on fork PRs the review is a deliberate maintainer action rather than automatic. Worth agreeing that is the trade we want.

## Before this does anything

`CLAUDE_CODE_OAUTH_TOKEN` needs to be reachable from this repository. `sodadata/soda` has it; this repo does not have it as a repo secret, and I cannot see whether the org secret's visibility covers us.

An `agent-review` label is also needed.

## Source of truth

The skill is generated. Its source lives in soda-punch and is synced with `just de-review-sync` — edit it there, not here. See sodadata/soda-punch#30.

https://claude.ai/code/session_013CLHBhYAG2CeRten4xSyEv

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,3 +9,33 @@ Fixes #
 - [ ] Tests added or updated to cover the change
 - [ ] Documentation updated where relevant
 - [ ] PR title follows the conventional commit format (e.g. `fix(sqlserver): ...`)
+
+## Review checklist
+
+<!-- Keep the sections that apply to this change and delete the rest.
+     The full version, with the reasoning behind each line, is the `de-review`
+     skill: .claude/skills/de-review/SKILL.md -->
+
+**Dialect or SQL rendering**
+- [ ] Every dialect that needs this overrides it; the ones inheriting the base are still correct
+- [ ] Quoting and escaping go through the dialect seam, not the call site
+- [ ] Warehouse types outside the mapping are handled explicitly, not defaulted to `TABLE`
+- [ ] Rendered SQL changed → snapshots re-recorded, not the job retried
+- [ ] A CI lane actually exercises this warehouse
+
+**Core seam or base class**
+- [ ] The change sits at the right altitude — base or dialect, not copied into a leaf
+- [ ] Every caller is identified
+- [ ] Downstream consumers cannot misread the result
+- [ ] Nothing here duplicates a sibling connector or a base implementation
+
+**Check semantics and evaluation**
+- [ ] Every field that makes two instances different is in `_get_id_properties()`
+- [ ] The exit code is right: `1` checks failed, `3` engine error, `4` results not sent
+- [ ] Nothing re-implements what `check_collections` already provides
+- [ ] The check cannot pass while reporting the wrong number
+
+**CI, workflow, packaging**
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -9,3 +9,33 @@ Fixes #
...
+     skill: .claude/skills/de-review/SKILL.md -->
+
+**Dialect or SQL rendering**
+- [ ] Every dialect that needs this overrides it; the ones inheriting the base are still correct
+- [ ] Quoting and escaping go through the dialect seam, not the call site
+- [ ] Warehouse types outside the mapping are handled explicitly, not defaulted to `TABLE`
+- [ ] Rendered SQL changed → snapshots re-recorded, not the job retried
+- [ ] A CI lane actually exercises this warehouse
+
+**Core seam or base class**
```
mivds, line 26: Claude lingo, not even sure what it means really 😅

.github/pull_request_template.md
```diff
@@ -9,3 +9,33 @@ Fixes #
...
+- [ ] The change sits at the right altitude — base or dialect, not copied into a leaf
+- [ ] Every caller is identified
+- [ ] Downstream consumers cannot misread the result
```
mivds, line 29: These are quite vague / non-sensical

.github/pull_request_template.md
```diff
@@ -9,3 +9,33 @@ Fixes #
 - [ ] Tests added or updated to cover the change
 - [ ] Documentation updated where relevant
 - [ ] PR title follows the conventional commit format (e.g. `fix(sqlserver): ...`)
+
+## Review checklist
```
mivds, line 13: Would probably drop this. I'm not sure PR description is the correct place for this type of thing. On top of that, since this is a public repo where we may get user contributions having this kind of extensive checklist is a no-go IMO

PR #2852, opened 2026-09-03
fix(cloud): report failed remote verification as ERROR instead of a passing UNKNOWN

m1n0:
## Description

Reported internally on 2026-09-02: when a contract is verified on a Runner with `--use-runner` and something goes wrong, the result says everything is fine. `has_errors`, `is_failed` and `get_errors()` come back false, and the CLI exits 0.

Root cause. `verify_contract_on_runner` leaves the result's status at its initial `UNKNOWN` on every path that ends without an outcome from Cloud, and `has_errors` is `status is ERROR`. The remote-state mapper knows four of Cloud's seven final states, so `canceled` and `timedOut` fell through to `UNKNOWN`, and `failed` mapped to check failures. The local error logs never reached `log_records` either, because capture only started after polling. On top of that, the runner loop in the session wrapped the call in a bare `except:` that dropped the item, so an exception left an empty session, which also reads as a pass.

What a CI caller saw before, and sees now. Exit codes via `interpret_contract_verification_result`:

| condition | before | after |
|---|---|---|
| insufficient permissions | 0 | 3 |
| 200 without a scan id | 0 | 3 |
| Cloud state `canceled` or `timedOut` | 0 | 3 |
| exception on the path | 0, empty session | 3, result kept with its logs |
| Cloud state `failed` | 1, read as check failures | 3 |
| contract upload failed | AttributeError, the method returned `[]` | 4 |
| permission query or verify command could not be sent | AttributeError on `None`, or 0 | 4 |
| permission query or verify command non-200, client poll timeout | 4, no errors on the result | 4, errors on the result |

Changes:

- Every path without an outcome returns `ERROR`. The send flag stays where Cloud could not be reached or refused the request, so those still exit 4.
- `canceled`, `timedOut` and `failed` map to `ERROR` and log why. `failed` no longer reads as check failures. It is also how a runner reports an engine that errored before producing results.
- `Logs()` capture starts before the first Cloud call, so `get_errors()` reports the failure. Every ERROR-mapped state logs a line, so `get_errors()` is never empty on exit 3.
- A failed upload returns a result instead of `[]`.
- An exception on the path is caught inside the capture window: the result keeps its dataset identity, scan id and the records logged before the raise. `_execute_on_runner` keeps an ERROR placeholder for parse and construct failures, the way the local path does.
- The log replay is guarded. A hiccup fetching logs does not change the outcome Cloud already reported.
- The status poll survives a transient request failure, an unknown state string, and a body without a state. All three are treated as not final and polling continues with the usual wait.
- A permission query that could not be sent takes the send-flag branch instead of raising on `None`.
- Non-JSON bodies from the command, status and logs endpoints no longer raise.

The existing `test_verify_contract_on_runner_permission_check` pinned the silent behaviour by asserting `log_records is None`. It now asserts the error. New tests cover each path and the exit code it maps to.

Unchanged on purpose: the client-side poll timeout keeps exit 4.

Docs: sodadata/docs#1045 adds the exit-code table to the core CLI reference.

## Checklist

- [x] Documentation updated where relevant
- [x] PR title follows the conventional commit format

## Review checklist

**Check semantics and evaluation**
- [x] The exit code is right: `1` checks failed, `3` engine error, `4` results not sent

**Core seam or base class**
- [x] Every caller is identified: `ContractImpl.verify_on_runner` is the only call site, reached from `_execute_on_runner`, plus the deprecated `verify_contract_on_agent` and `verify_on_agent` aliases
- [x] Downstream consumers cannot misread the result

```customer-facing-release-note
Verifying a contract on a runner now reports an error when the scan cannot complete.
```

https://claude.ai/code/session_01ChwRG6RQYKhsg5QFJhtdCN

### DIFF
--- a/soda-core/src/soda_core/common/soda_cloud.py
+++ b/soda-core/src/soda_core/common/soda_cloud.py
@@ -679,6 +679,45 @@ class SodaCloud:
             status=CheckCollectionStatus.UNKNOWN,
         )
 
+        # Capture from the first Cloud call, not from the poll onwards: the errors logged on
+        # the failure paths (permissions, upload, command, poll) have to land in
+        # ``log_records`` for ``get_errors()`` to report them.
+        logs: Logs = Logs()
+        try:
+            verification_result.status = self._verify_contract_on_runner(
+                verification_result=verification_result,
+                contract_yaml_str_original=contract_yaml_str_original,
+                contract_local_file_path=contract_local_file_path,
+                dataset_identifier=dataset_identifier,
+                variables=variables,
+                blocking_timeout_in_minutes=blocking_timeout_in_minutes,
+                publish_results=publish_results,
+                verbose=verbose,
+            )
+        finally:
+            logs.close()
+            verification_result.log_records = logs.get_log_records()
+
+        return verification_result
+
+    def _verify_contract_on_runner(
+        self,
+        verification_result: ContractVerificationResult,
+        contract_yaml_str_original: str,
+        contract_local_file_path: Optional[str],
+        dataset_identifier: DatasetIdentifier,
+        variables: dict[str, str],
+        blocking_timeout_in_minutes: int,
+        publish_results: bool,
+        verbose: bool,
+    ) -> CheckCollectionStatus:
+        """Drive one remote verification and return its status.
+
+        Every path that ends without an outcome from Cloud returns ERROR: the result starts
+        out UNKNOWN, and UNKNOWN reads as a pass to ``has_errors``, ``is_failed`` and the CLI
+        exit code. ``sending_results_to_soda_cloud_failed`` is set on top of that only where
+        Cloud could not be reached or refused the request, which turns exit code 3 into 4.
+        """
         can_publish_and_verify, reason = self.can_publish_and_verify_contract(
             dataset_identifier.data_source_name, dataset_identifier.prefixes, dataset_identifier.dataset_name
         )
@@ -688,12 +727,13 @@ class SodaCloud:
                 verification_result.sending_results_to_soda_cloud_failed = True
             else:
                 logger.error(f"Skipping contract verification because of insufficient permissions: {reason}")
-            return verification_result
+            return CheckCollectionStatus.ERROR
 
         file_id: Optional[str] = self._upload_contract_yaml_file(contract_yaml_str_original)
         if not file_id:
-            logger.critical("Contract wasn't uploaded so skipping sending the results to Soda Cloud")
-            return []
+            logger.error("Contract wasn't uploaded, skipping remote contract verification")
+            verification_result.sending_results_to_soda_cloud_failed = True
+            return CheckCollectionStatus.ERROR
 
         verify_contract_command: dict = {
             "type": "sodaCoreVerifyContract" if publish_results else "sodaCoreTestContract",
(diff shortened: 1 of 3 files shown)


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -679,6 +679,45 @@ def verify_contract_on_runner(
...
+                contract_local_file_path=contract_local_file_path,
+                dataset_identifier=dataset_identifier,
+                variables=variables,
+                blocking_timeout_in_minutes=blocking_timeout_in_minutes,
+                publish_results=publish_results,
+                verbose=verbose,
+            )
+        finally:
+            logs.close()
+            verification_result.log_records = logs.get_log_records()
```
Niels-b, line 699: The `finally` populates `verification_result`, and then the exception propagates — so on the raising path this object is never returned and the records are discarded.

`_execute_on_runner` replaces it with `ContractImpl.build_error_result(...)`, whose placeholder carries only `str(exc)` in a fresh `Logs`, plus a `Contract` with `data_source_name=None, dataset_prefix=[], dataset_name=""` (`check_collections/base.py:1145+`). So a consumer of `verify_contract()` loses both the diagnostics this block just captured and the dataset identity. The `init_log_records` prepend at `contract_verification_impl.py:446` is skipped too, since it's assigned after the call inside the `try`.

Catching here instead would keep both — set `status = ERROR`, `logger.error(..., exc_info=True)` inside the capture window, return the populated result — with `_execute_on_runner`'s net staying as the backstop for the `ContractYaml.parse` / `ContractImpl(...)` failures that `test_execute_on_runner_isolates_exceptions_as_error_results` covers. It would fold in the `_replay_remote_scan_logs` case below as well.

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -709,36 +749,51 @@ def verify_contract_on_runner(
...
         if not scan_id:
-            logger.warning("Did not receive a Scan ID from Soda Cloud")
-            return verification_result
+            logger.error("Did not receive a Scan ID from Soda Cloud")
+            return CheckCollectionStatus.ERROR
+        verification_result.scan_id = scan_id
 
         scan_is_finished, contract_dataset_cloud_url, scan_status = self._poll_remote_scan_finished(
             scan_id=scan_id, blocking_timeout_in_minutes=blocking_timeout_in_minutes
         )
```
Niels-b, line 769: One path still aborts via an exception rather than a mapped status, which is the one gap left in the PR's thesis.

`_poll_remote_scan_finished` calls `RemoteScanStatus.from_value` (`soda_cloud.py:1263`), which raises `ValueError` on any state string it doesn't know — and the mapping isn't total against the boundary: Cloud's `ScanStateDTO` includes `started` (internal `ScanState.STARTED`, mapped at `PublicApiScanDTOConverterModule.java:109`), which `RemoteScanStatus` has no member for. Same shape for a transient `requests` error inside the loop: `if not response: continue` only catches a falsy `Response`, not a raise.

I could not demonstrate this firing for contract verification specifically — `ScanState.java:26-28` notes *"a contract verification reports no scan start at all, so its results are the first state its scan ever receives"* — so treat it as hardening rather than a live break. Catching `ValueError` and treating an unknown state as non-final would be a couple of lines in the same spirit as the rest of the PR.

soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -709,36 +749,51 @@ def verify_contract_on_runner(
...
 
-        verification_result.status = _map_remote_scan_status_to_contract_verification_status(scan_status)
+        self._replay_remote_scan_logs(scan_id)
```
Niels-b, line 771: This is unguarded and now runs *before* the status is computed and returned, so a hiccup fetching the logs turns a successful verification into exit 3.

`_replay_remote_scan_logs` goes through `_execute_rest_get` → `requests.get`, which — unlike `_execute_command`, which swallows and returns `None` — has no try/except and raises on a connection reset or timeout. That propagates out of `_verify_contract_on_runner`, past the `finally`, into the new `except Exception` at `contract_verification_impl.py:451`, producing an ERROR placeholder and exit 3 for a scan Cloud already reported as `completed`.

Not a regression — before this PR the same raise was swallowed by the bare `except:` and exited 0 — but it does turn a silent wrong-pass into a loud wrong-fail on a path where the answer was already known. The replay is cosmetic; it shouldn't be able to invalidate an outcome Cloud already gave us. A `try/except Exception` + `logger.warning` around this call keeps the intent without the false alarm.

Niels-b (approved):
Thanks for fixing this. Should we make a note about this in the release notes / update the documentation with the updated exit code cases?

Claude still found some things I think are worth addressing, which it should have commented ;)

soda-core/src/soda_core/common/soda_cloud.py
m1n0: Done in fac1efdc. Caught inside the capture window, status ERROR, and the exception text goes into the error line so get_errors() carries it. The result keeps its dataset identity and scan_id. The session-level net stays for parse and construct failures. Test: test_verify_contract_on_runner_keeps_result_when_the_path_raises.

soda-core/src/soda_core/common/soda_cloud.py
m1n0: Done in fac1efdc. An unknown state, a body without one, a falsy response and a raise from the status request all log a warning and keep polling with the usual wait, so `started` or a connection reset no longer abort the run. Tests for the unknown state and for the transient raise.

soda-core/src/soda_core/common/soda_cloud.py
m1n0: Guarded in fac1efdc. Warning, and the outcome Cloud reported stands. Test: test_verify_contract_on_runner_survives_log_fetch_failure.

m1n0:
Docs: sodadata/docs#1045 adds the exit-code table the core CLI page never had and names the runner states behind 3 and 4. The release-note block is in the description.

One more gap on a second pass, also in fac1efdc: the permission query returns None when the request itself fails, and the old code dereferenced status_code on it. It now takes the send-flag branch, exit 4.

PR #2853, opened 2026-09-05
fix(bigquery): make time partition filters eligible for partition elimination

mivds:
## Problem

The BigQuery dialect's `sql_expr_timestamp_coerce` hook wraps the partition expression in `timestamp(...)`. BigQuery only prunes partitions through a small set of functions on the partitioning column, and the `timestamp()` constructor is not one of them — the wrapped filter full-scans, and on tables with `require_partition_filter` the query is rejected with *"Cannot query over table ... without a filter over column(s) ... that can be used for partition elimination"*. This is the v4 port of the same defect fixed in v3 soda-library (sodadata/soda-library#790), where it was customer-reported. Two adjacent seams share the type problem: aware window bounds render with a UTC offset (which cannot coerce to a DATETIME column — the very reason the wrap existed), and the typed anchors/bucket arithmetic are TIMESTAMP-typed, so a DATETIME partition expression already fails inside `TIMESTAMP_DIFF`.

## Change

Compare everything in DATETIME. `sql_expr_timestamp_coerce` becomes `CAST(... AS DATETIME)` — the one conversion BigQuery accepts for partition elimination (verified against live BigQuery with dry-run byte estimates on a `require_partition_filter` table, nested casts included); it compiles for DATE, DATETIME and TIMESTAMP expressions alike and converts TIMESTAMP→DATETIME at UTC, preserving window semantics exactly.

Literals render **untyped**, offset-less, UTC: `literal_datetime` drops the offset (an offset-carrying string cannot coerce to DATETIME) and `literal_timestamp_typed` becomes an untyped string too — BigQuery coerces string literals inside `DATETIME_DIFF`/`DATETIME_ADD` and against raw TIMESTAMP or DATETIME columns alike, which keeps user-authored custom metric SQL (`col BETWEEN ${soda.PARTITION_START_TIME} ...` on TIMESTAMP columns) compiling unchanged; a typed `DATETIME` literal broke exactly that pattern in the v3 CI.

The time-bucket delta measures elapsed `MICROSECOND`s floor-divided by the unit size instead of `DATETIME_DIFF` at DAY/HOUR: those parts count *calendar-boundary crossings* (a 2-hour span across midnight counts as 1 DAY), which mis-buckets backfill grids anchored off midnight — confirmed with live constant queries and caught by the v3 2AM CI tests. The microsecond form matches the base dialect's `EXTRACT(EPOCH ...)` elapsed semantics; `ADD_INTERVAL` stays exact arithmetic on `DATETIME_ADD`.

**Pairs with** sodadata/soda-extensions#604, which routes the MM bulk builders through the coerce hook — with this change alone, an MM `TIME_DELTA` over a raw TIMESTAMP expression would stop compiling, so the two should land together.

## Testing

Dialect unit pins updated/added (coerce hook, offset stripping, untyped anchors, microsecond-based delta; written first and watched fail). SQL generated by the paired MM builders + this dialect validated against live BigQuery via dry runs: accepted on a `require_partition_filter` TIMESTAMP-partitioned table and pruned on DATE/TIMESTAMP partitions for regular scan, guard, freshness lookback, backfill CTE, and the profiling window filter with offset-carrying bounds. DIFF boundary-vs-elapsed semantics and untyped-string coercion confirmed with live constant queries. The v3 twin of this change passes the previously failing BigQuery CI suite locally.

### DIFF
--- a/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
+++ b/soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
@@ -1,5 +1,6 @@
 import logging
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from typing import Optional
 
 from soda_bigquery.common.data_sources.bigquery_data_source_connection import (
@@ -269,14 +270,19 @@ class BigQuerySqlDialect(SqlDialect, sqlglot_dialect="bigquery"):
         quantile_number: int = int(percentile_within_group.percentile * 1000)
         return f"APPROX_QUANTILES({expression_sql}, 1000)[{quantile_number}]"
 
-    # Singular unit names for TIMESTAMP_DIFF/TIMESTAMP_ADD. WEEK is deliberately
-    # absent: BigQuery's TIMESTAMP functions only accept MICROSECOND..DAY parts
-    # (WEEK is a DATE_DIFF/DATETIME_DIFF-only part), so weekly arithmetic
-    # renders day-denominated in both builders below.
+    # Singular unit names for DATETIME_ADD. WEEK is deliberately absent: weekly
+    # arithmetic renders day-denominated in both builders below, keeping
+    # cross-dialect bucket indexes identical.
     _TIME_BUCKET_UNIT_NAMES: dict = {
         "days": "DAY",
         "hours": "HOUR",
         "seconds": "SECOND",
+    }
+
+    _TIME_BUCKET_UNIT_MICROSECONDS: dict = {
+        "days": 86_400_000_000,
+        "hours": 3_600_000_000,
+        "seconds": 1_000_000,
     }
 
     def _build_time_delta_sql(self, time_delta: TIME_DELTA) -> str:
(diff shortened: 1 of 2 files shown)


soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -269,16 +270,21 @@ def _build_percentile_within_group_sql(self, percentile_within_group: PERCENTILE
...
     _TIME_BUCKET_UNIT_NAMES: dict = {
         "days": "DAY",
         "hours": "HOUR",
         "seconds": "SECOND",
     }
 
+    _TIME_BUCKET_UNIT_MICROSECONDS: dict = {
+        "days": 86_400_000_000,
+        "hours": 3_600_000_000,
+        "seconds": 1_000_000,
```
Niels-b, line 285: NIT / personal preference, feel free to ignore

Can we make this a formula? E.g `hours: 60 * 60 * 1_000_000`

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -287,20 +293,21 @@ def _build_time_delta_sql(self, time_delta: TIME_DELTA) -> str:
...
-        unit_name: str = self._TIME_BUCKET_UNIT_NAMES[unit]
-        sql: str = f"TIMESTAMP_DIFF({end_sql}, {start_sql}, {unit_name})"
-        if count != 1:
-            sql = f"CAST(FLOOR({sql} / {count}) AS INT)"
-        return sql
+        # Elapsed MICROSECONDs floor-divided by the unit size: DATETIME_DIFF at
+        # DAY/HOUR counts calendar-boundary crossings (a 2h span across midnight
+        # is 1 DAY), which would mis-bucket grids anchored off midnight.
+        unit_microseconds: int = self._TIME_BUCKET_UNIT_MICROSECONDS[unit] * count
+        sql: str = f"DATETIME_DIFF({end_sql}, {start_sql}, MICROSECOND)"
```
Niels-b, line 300: Drive-by: this PR invalidates the class docstrings of the two nodes it re-renders, over in `soda-core/src/soda_core/common/sql_ast.py` — worth fixing here, since they're the first thing you read when adding a dialect override:

- `sql_ast.py:506-507` (`TIME_DELTA`): _"Snowflake overrides with a TIMESTAMPDIFF-seconds form; BigQuery with TIMESTAMP_DIFF."_
- `sql_ast.py:530` (`ADD_INTERVAL`): _"Snowflake overrides with TIMESTAMPADD; BigQuery with TIMESTAMP_ADD."_

Both now render `DATETIME_DIFF` / `DATETIME_ADD` on BigQuery.

Niels-b (approved):
LGTM. Claude picked up one small comment about docs. And I left one NIT (feel free to ignore)

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
mivds: Done in edf49c85 — `"days": 24 * 60 * 60 * 1_000_000`, `"hours": 60 * 60 * 1_000_000`.

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
mivds: Fixed in edf49c85. `TIME_DELTA` now reads "BigQuery with a DATETIME_DIFF-microseconds form" and `ADD_INTERVAL` "BigQuery with DATETIME_ADD".

PR #2854, opened 2026-09-05
fix(bigquery): resolve INFORMATION_SCHEMA regions from dataset metadata

mivds:
## Problem

Table-metadata queries — dataset discovery (`discover_qualified_objects`) and table-existence checks (`verify_if_table_exists`) — on a BigQuery data source without a configured `location` silently return nothing when the data lives outside the US multi-region.

The region used to qualify `INFORMATION_SCHEMA` views is detected with `SELECT @@location`. That query references no tables, so BigQuery cannot infer a location for the job and runs it in its default location — the US multi-region ([BigQuery locations docs](https://docs.cloud.google.com/bigquery/docs/locations)). `@@location` reports the job's own location, so the detection is circular: it answers `US` regardless of where the datasets actually live, and the wrong region's `INFORMATION_SCHEMA` views are legitimately empty. Contract verification is unaffected because its queries are fully qualified; only the region-scoped metadata paths break. This also affects downstream consumers of these APIs, such as diagnostics-warehouse table-existence checks.

## Change

Regions are now read from dataset metadata (`datasets.get` / `datasets.list` — project-scoped REST calls that span all regions) instead of from a probe query:

- When the scope names a dataset (e.g. `verify_if_table_exists` with `[project, dataset]` prefixes), that dataset's own region is used — one lookup, one query.
- Otherwise the project's datasets are listed and the metadata query runs once per distinct region, merging the results — projects with datasets in multiple regions now work as well.
- A configured `location` still pins the single region and skips all metadata lookups, preserving support for accounts without `bigquery.datasets.get` (region views are permission-filtered and remain readable to them).

Region resolution moved from query construction (`create_metadata_tables_query`) to execution time (`BigQueryMetadataTablesQuery.execute`), where the scope — and therefore the dataset — is known. `get_location()` and the `SELECT @@location` probe are removed. The test-helper snapshot passthrough that kept that probe out of query snapshots is replaced by `_snapshot_pin_metadata_lookups()`: dataset metadata is a REST call, which the snapshot wrapper (SQL-only) cannot intercept, so replay would otherwise open a real connection. The BigQuery override pins the region through the configured-`location` path rather than a monkeypatch.

The equivalent fix for the v3 engine is sodadata/soda-library#789.

## Follow-up

The identical probe lives in soda-extensions metric monitoring (`metadata_providers.py`), where it silently empties row count, modification time and history backfill for non-US datasets. Tracked as OBSL-1045, implemented in sodadata/soda-extensions#609 — draft until this merges and the `soda-core` floor there is bumped.

## Testing

Unit tests with a mocked BigQuery client cover: configured-location pinning (no metadata calls), known-dataset resolution, project-wide listing with region de-duplication, per-dataset caching, per-region query fan-out and result concatenation, and removal of the probe. Full `soda-bigquery` + `soda-tests` unit suites pass (1435 tests); not yet verified against a live multi-region BigQuery project.

```customer-facing-release-note
BigQuery dataset discovery now finds datasets in every region.
```


soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
```diff
@@ -81,25 +123,50 @@ def _create_data_source_connection(self) -> DataSourceConnection:
...
-        return location
+    def region_for_dataset(self, project_id: Optional[str], dataset_id: str) -> str:
+        """Get the BigQuery location of a dataset from its metadata (a REST call, not a query job)."""
+        configured_location = self.data_source_model.connection_properties.location
+        if configured_location:
+            return configured_location
+        project_id = project_id or self.data_source_connection.project_id
+        cache_key = (project_id, dataset_id)
+        if cache_key not in self._dataset_region_cache:
+            dataset = self.data_source_connection.client.get_dataset(f"{project_id}.{dataset_id}")
```
Niels-b, line 134: Snapshot replay is no longer hermetic for the BigQuery metadata paths.

The `_snapshot_passthrough_queries` override this PR deletes existed precisely to keep the region lookup out of snapshots. Its replacement is a `client.get_dataset(...)` REST call, which the snapshot layer can't intercept — it only wraps `execute_query`.

In replay mode `SnapshotDataSourceConnection.__getattr__` (`soda-tests/src/helpers/snapshot_connection.py:441-459`) proxies `.client` to `_real`, and when `_real` is `None` it calls the fallback factory to **open a real connection**. So a replay-mode BigQuery test that reaches `discover_qualified_objects` / `verify_if_table_exists` now needs live credentials and does live REST work — against the wrapper's stated contract at `snapshot_connection.py:336` (_"In replay mode: returns cached results without any database interaction"_).

CI won't surface it: the bigquery leg always has credentials. Could we re-add an equivalent seam in `BigQueryDataSourceTestHelper` that pins `regions_in_scope` / `region_for_dataset` instead of the SQL?

Niels-b (commented):
I like the approach and idea in general, but this one may require some follow ups. One for the snapshots and one for soda-extensions (see [here](https://sodadata.slack.com/archives/C02J6Q493PY/p1789127793361879?thread_ts=1789121184.026789&cid=C02J6Q493PY) )

soda-bigquery/src/soda_bigquery/common/data_sources/bigquery_data_source.py
mivds: Confirmed — `regions_in_scope` / `region_for_dataset` reach `data_source_connection.client`, which the wrapper only proxies via `__getattr__`, so replay opens a real connection. `test_discovery.py` and `query_existing_test_views` both reach it, so the bigquery replay leg was doing live REST work.

Seam re-added one level up in 8aa68b9: `DataSourceTestHelper._snapshot_pin_metadata_lookups()`, called from both snapshot install paths, next to `_snapshot_passthrough_queries`. The BigQuery override sets the connection's configured `location` from `BIGQUERY_LOCATION` — the production escape hatch that already short-circuits both methods before any REST call — so the pinning runs through real code instead of a monkeypatch, and the client is untouched because it is constructed before the helper mutates the model. Pinned by a unit test asserting neither `get_dataset` nor `list_datasets` is called.

Niels-b (approved):
Thanks for fixing the snapshot mechanism here 🙌

PR #2855, opened 2026-09-07
fix(databricks): enable Delta column mapping on tables Soda creates

m1n0:
Delta rejects `CREATE TABLE` when a column name contains one of ` ,;{}()\n\t=` unless the table has column mapping enabled. Our diagnostics tables mirror source column names verbatim, so the Linklaters dataset with columns like `First Name` fails the whole failed-rows stage with `DELTA_INVALID_CHARACTERS_IN_COLUMN_NAMES`. Their source table reads fine, it already has column mapping on. Quoting is not the issue, the SQL parses.

Reported here: https://sodadata.slack.com/archives/CAXRR8SS3/p1788538370535779

## Change

New `_build_create_table_properties_sql` hook on `SqlDialect`, rendered after the column list in `CREATE TABLE` and before `AS` in a CTAS. Both paths create diagnostics tables. Empty by default, Databricks returns the TBLPROPERTIES that turn column mapping on.

## Always, not only when needed

Every table we create on Databricks gets column mapping, so no dataset can slip through later.

The alternative was emitting the properties only when a column name actually contains a restricted character, to leave every other table on its current Delta protocol version. @Niels-b measured that on the CI workspace and the premise was wrong: a table created with no table properties at all is already reader 3 / writer 7, because that runtime represents Delta features as table features rather than the legacy monolithic versions. The `TBLPROPERTIES` here adds the columnMapping feature and bumps no protocol, so there is no older-client risk to be conservative about. His caveat stands, one workspace and one runtime, and the 3/7 default comes from DBR plus catalog defaults rather than a guarantee.

The `minReaderVersion` 2 / `minWriterVersion` 5 pair is a floor, not an exact value. It is a no-op on a table-features runtime and only does work on an older DBR where column mapping needs it declared. Keeping it as a defensive shim.

Two more, out of scope here:

- `DatabricksHiveSqlDialect` inherits the properties. On a hive_metastore warehouse that defaults to a non-Delta provider the `delta.*` properties would land on a non-Delta table. Neither of us has one to test against.
- `drop_column_supported` returns `False` with a note that it needs column mapping mode. New tables qualify once this lands.

## Tests

Unit tests for both SQL paths, full unit suite green.

Verified live on Databricks by @Niels-b in both DDL modes, `CREATE TABLE` and in-source-transfer CTAS, along with `INSERT`, `ALTER TABLE ADD COLUMN` with a restricted-character name, `CREATE VIEW` + select, and the metadata query returning every name intact. The regression test covering every DWH object family is sodadata/soda-extensions#605, which merges after this.

```customer-facing-release-note
Failed rows are now stored on Databricks for datasets whose column names contain spaces or other characters Delta restricts.
```

https://claude.ai/code/session_016xsqPCEQpGnYcsnQD9sysw

### DIFF
--- a/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
+++ b/soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
@@ -98,6 +98,17 @@ class DatabricksSqlDialect(SqlDialect, sqlglot_dialect="databricks"):
     # Databricks SQL rejects DROP TABLE ... CASCADE with PARSE_SYNTAX_ERROR
     # (CASCADE is valid on DROP SCHEMA only) — same as sparkdf/trino/athena.
     SUPPORTS_DROP_TABLE_CASCADE: bool = False
+    # Delta refuses to create a table whose column names contain one of " ,;{}()\n\t=" unless the
+    # table has column mapping enabled. Soda's diagnostics tables mirror source column names
+    # verbatim, so a dataset with a column like "First Name" made the whole failed-rows stage fail
+    # with DELTA_INVALID_CHARACTERS_IN_COLUMN_NAMES, even though the source table itself is
+    # readable (it has column mapping on). Every table Soda creates here declares it.
+    DELTA_TBLPROPERTIES: str = (
+        "TBLPROPERTIES ("
+        "'delta.columnMapping.mode' = 'name', "
+        "'delta.minReaderVersion' = '2', "
+        "'delta.minWriterVersion' = '5')"
+    )
 
     SODA_DATA_TYPE_SYNONYMS = (
         (SodaDataTypeName.TEXT, SodaDataTypeName.VARCHAR, SodaDataTypeName.CHAR),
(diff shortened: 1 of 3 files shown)


soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
```diff
@@ -98,6 +98,17 @@ class DatabricksSqlDialect(SqlDialect, sqlglot_dialect="databricks"):
...
+    # Delta refuses to create a table whose column names contain one of " ,;{}()\n\t=" unless the
+    # table has column mapping enabled. Soda's diagnostics tables mirror source column names
+    # verbatim, so a dataset with a column like "First Name" made the whole failed-rows stage fail
+    # with DELTA_INVALID_CHARACTERS_IN_COLUMN_NAMES, even though the source table itself is
+    # readable (it has column mapping on). Every table Soda creates here declares it.
+    DELTA_TBLPROPERTIES: str = (
+        "TBLPROPERTIES ("
+        "'delta.columnMapping.mode' = 'name', "
+        "'delta.minReaderVersion' = '2', "
+        "'delta.minWriterVersion' = '5')"
```
mivds, line 110: Should be fine, but this pins us to Databricks Runtime 10.4 LTS or later. 10.4 was released on March 18, 2022 and is no longer supported, so probably safe.

mivds (approved):

Niels-b (approved):
Thank you so much for looking into this 🙌 !

Seems like the right fix to me.
As a follow up, we may want to add a test for this in soda-extensions, just so we test the full flow and for other datasources as well.

soda-databricks/src/soda_databricks/common/data_sources/databricks_data_source.py
m1n0: good find, indeed that should be safe though

PR #2856, opened 2026-09-08
Bump soda-core v4.23.1

mivds:
Bump versions to v4.23.1


Niels-b (approved):
We can use [SKIP-TESTS[ to skip the CI.
Otherwise LGTM 👍 . Thanks 🙌

PR #2857, opened 2026-09-08
fix(failed_rows): coerce rows_tested_query result to an int row count

m1n0:
## What

A `failed_rows` check's `rows_tested_query` result is now coerced to an int before it becomes the `check_rows_tested` measurement. A value that is not a number is dropped, with an error naming the fix, and the check carries on unmeasured.

## Why

`RowsTestedQuery.execute()` took `query_result.rows[0][0]` verbatim. If the query returns rows rather than a count, that first cell can be anything, and it reached Soda Cloud as `checkRowsTested`. The API types that field as an integer, so a text value failed the JSON parse of the **whole** `sodaCoreInsertScanResults` body:

```
Status Code: 400
Message: 'Could not parse request from json body' | Response Code: 'invalid_request'
```

Because the rejection happens at body-parse time, the blast radius is the entire scan rather than the one bad check:

- no check results and no logs in the Soda Cloud UI
- failed-rows extraction skipped, `Skipping failed rows extraction: No soda cloud response available`
- launcher exits 4

And nothing in the engine output pointed at the field. The only clue was the printed summary table:

```
failed_rows_count:   9918
failed_rows_percent: 0
dataset_rows_tested: 463455012
check_rows_tested:   YARDI_EMEA_SCHRODERS     <-- a string
```

The value never passed through `convert_db_value`, because `RowsTestedQuery` builds its `Measurement` directly instead of going down the aggregation-metric path that applies it.

Found on a support ticket where the customer's `rows_tested_query` was missing its `select count(*) from ( ... )` wrapper, so it returned the query's rows and the first column was `SOURCEDB`.

## Behaviour change

A non-numeric `rows_tested_query` result used to kill the scan. It now logs a runtime error and leaves `check_rows_tested` unmeasured, which is the same shape the code already produces for a query returning NULL. `failedRowsPercent` is 0 and the check still evaluates on `failed_rows_count`. The scan reports one runtime error instead of vanishing.

`int()` truncation is deliberate: a row count is integral, and `Decimal`/`float` counts from a warehouse should land as ints.

## Test

`test_failed_rows_rows_tested_query_returning_non_numeric_is_rejected` pins the whole chain: the text value never reaches the wire, the rest of the diagnostics survive, the check still fails on its own count, and the error tells the user to wrap the query in `SELECT COUNT(*)`.

## Follow-ups, not in this PR

- Nothing stops a future non-numeric value from any other check type reaching a numeric field in `diagnostics.v4` and taking down a scan the same way. A wire-level type guard in `_build_v4_diagnostics_check_type_json_dict` would close that off, but it touches every check type and deserves its own review.
- `datasetRowsTested` is an `Integer` on the Cloud side. The customer that hit this is already at 463,809,660 rows, so a dataset past 2.1B rows produces the identical 400.

```customer-facing-release-note
A `rows_tested_query` returning a non-numeric value no longer fails the whole scan.
```

https://claude.ai/code/session_01Rmhr4oTPpTjrsvFKC7FRvK

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
@@ -418,6 +418,6 @@ class RowsTestedQuery(Query):
                 f"Could not read a row count from the rows tested query: expected a number, got "
                 f"{type(value).__name__} {value!r}. Use a query that returns a single count, for "
                 f"example 'SELECT COUNT(*) FROM (<your query>)'. Continuing without "
-                f"check_rows_tested.\n{self.sql}"
+                f"check_rows_tested.\nExecuted SQL:\n{self.sql}"
             )
             return None
--- a/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
+++ b/soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
@@ -391,6 +391,33 @@ class RowsTestedQuery(Query):
         if not query_result.rows:
             metric_value = None
         else:
-            metric_value = query_result.rows[0][0]
+            metric_value = self._to_row_count(query_result.rows[0][0])
         metric_impl: MetricImpl = self.metrics[0]
         return [Measurement(metric_id=metric_impl.id, value=metric_value, metric_name=metric_impl.type)]
+
+    def _to_row_count(self, value: any) -> Optional[int]:
+        """Coerce the first cell of the rows_tested_query result to an int row count.
+
+        This value reaches Soda Cloud as ``checkRowsTested``, which the API types as an
+        integer. A non-numeric value is rejected at JSON-parse time, which fails the whole
+        ``sodaCoreInsertScanResults`` body: every check result and every log line of that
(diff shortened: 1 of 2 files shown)


soda-core/src/soda_core/contracts/impl/check_types/failed_rows_check.py
```diff
@@ -391,6 +391,33 @@ def execute(self) -> list[Measurement]:
...
+        if value is None:
+            return None
+        try:
+            return int(value)
+        except (TypeError, ValueError, OverflowError):
+            logger.error(
+                f"Could not read a row count from the rows tested query: expected a number, got "
+                f"{type(value).__name__} {value!r}. Use a query that returns a single count, for "
+                f"example 'SELECT COUNT(*) FROM (<your query>)'. Continuing without "
+                f"check_rows_tested.\n{self.sql}"
```
mivds: The SQL suffix here seems slightly out of place. Suggest giving it a header:


```suggestion
                f"check_rows_tested.\nExecuted SQL:\n{self.sql}"
```

mivds (approved):

Niels-b (approved):

PR #2858, opened 2026-09-11
fix(checks): skip non-numeric check values instead of failing the scan

m1n0:
## What

A check value that is not a number is now skipped, and the check is reported NOT_EVALUATED. It used to either reach Soda Cloud and fail the upload of the whole scan, or abort the verification.

Three places turn a user-controlled warehouse value into a check value. They now share `MetricImpl._convert_db_value_to_number`:

- **metric check `query`** took `rows[0][0]` verbatim, so text went onto the wire as `diagnostics.value`.
- **metric check `expression`** and the **`aggregate` function** called `float()`, so text raised `ValueError` out of the aggregation query and aborted every check sharing it.

Finite numbers pass through unchanged. Numeric text, as produced by a `CAST` to a string type, is parsed, which is what `float()` already did on the expression and aggregate paths. Anything else, NaN and infinity included, logs an error naming the fix, the value becomes `None`, and the check is not evaluated, exactly as for an unmeasured metric.

As a backstop for check types that build their own value, the evaluate loop in `check_collections` applies the same rule to any `threshold_value` that is not a finite number, logging an error the way the unsupported-check path next to it does. Both spots use `is_finite_number`, which also treats a value `math.isfinite` cannot convert, a signalling `Decimal('sNaN')` or an int beyond float range, as not finite rather than raising.

## Why

`diagnostics.value` is a double in the Soda Cloud API, so one text value fails the JSON parse of the **whole** `sodaCoreInsertScanResults` body:

```
Status Code: 400
Message: 'Could not parse request from json body' | Response Code: 'invalid_request'
```

The scan then has no results and no logs in Soda Cloud, and the launcher exits 4. Nothing in the engine output names the field. The only tell is `NOT_EVALUATED` on that check in the summary table. A query ending in `CAST(failed_rows AS STRING)` is enough to hit it.

On the expression and aggregate paths the failure is earlier and just as wide: `max` on a text column raises inside `AggregationQuery.execute`, which has no per-metric error handling.

NaN and infinity fail on the client instead. They pass as numbers, but `requests` serializes the body with `allow_nan=False`, so it raises `InvalidJSONError` before anything is sent, and `_execute_cqrs_request` turns that into a critical log. The whole upload is lost with nothing naming the field. `AVG` or `MAX` over a double column holding one NaN is enough.

`MetricQuery` never called `convert_db_value`, unlike the aggregation path. It does now, and `MetricImpl` owns the identity default so that holds for any metric handed to it.

Same class as #2857, which covered `rows_tested_query`.

## Behaviour change

- Numeric text from a metric query, like `'12'`, is evaluated as a number. It used to be not evaluated and then fail the upload.
- Other text, `nan`/`inf` text or dates from a metric query log an error and leave the check not evaluated. They used to fail the whole upload.
- Text from a metric expression or an aggregate function logs an error and leaves the check not evaluated. It used to abort the verification.
- NaN or infinity, native or `Decimal`, on any of these paths or from any check type, logs an error and leaves the check not evaluated. It used to lose the whole upload client-side.
- Finite numbers pass through unchanged, `Decimal` and `bool` included, so downstream consumers that build on `MetricQuery` keep the types they get today.
- A not-evaluated check still goes out with the placeholder value the payload already uses for unmeasured checks. That part is unchanged.

## Test

- `test_udf_metric_check.py`: numeric text evaluates. Non-numeric text from a query, and text from an expression, leave the check not evaluated with the error, while the other check in the contract still evaluates and uploads.
- `test_aggregate_check.py`: `max` on a text column is not evaluated, and `avg` on another column in the same query still passes.
- `test_metric_check.py`: the conversion table. The existing `MagicMock` metric now needs an identity `convert_db_value`, since `MetricQuery` delegates to it.
- `test_udf_metric_check.py` also runs a NaN metric query end to end on DuckDB and PostgreSQL, the two warehouses with a shared NaN cast.
- `test_non_numeric_check_value.py`: the evaluate-loop backstop and `is_finite_number`.
- `MockSodaCloud` now serializes request bodies the way `requests` does, so a payload production would refuse to send fails in tests instead of being recorded as sent. It used to record the raw dict, which is why this class was invisible. No existing test trips it.

All new tests fail on main. Full suite run locally on DuckDB, the other warehouses are CI.

## Not in this PR

- The numeric fields inside `diagnostics.v4` have no guard. Their one user-controlled source, `rows_tested_query`, was fixed in #2857.

```customer-facing-release-note
Metric and aggregate checks that return text no longer fail the whole scan.
```

https://claude.ai/code/session_017Rcd96rg659FFESH7LtPcB

### DIFF
--- a/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import math
+import reprlib
 from abc import ABC
 from dataclasses import dataclass
 from enum import Enum
@@ -1608,6 +1610,33 @@ class MetricImpl:
     def sql_condition_expression(self) -> Optional[SqlExpression]:
         pass
 
+    def convert_db_value(self, value: any) -> any:
+        return value
+
+    def _convert_db_value_to_number(self, value: any, source: str, hint: str) -> Optional[Number]:
+        """Read a warehouse value as a number, or skip it.
+
+        Numbers pass through unchanged, and numeric text, as produced by a CAST to a string type,
+        is parsed. Anything else is skipped with an error: the value becomes None, so the check
+        using it is not evaluated. Soda Cloud types check values as numbers, and a single value it
+        cannot parse rejects the results of the whole scan.
+        """
+        if value is None or isinstance(value, Number):
+            return value
+        if isinstance(value, str):
+            try:
+                number: float = float(value)
+            except ValueError:
+                pass
+            else:
+                if math.isfinite(number):
+                    return number
+        logger.error(
+            f"{source} returned {type(value).__name__} {reprlib.repr(value)}, not a number, "
+            f"so the check is not evaluated. {hint}"
+        )
+        return None
+
 
 class AggregationMetricImpl(MetricImpl):
     def __init__(
@@ -1641,9 +1670,6 @@ class AggregationMetricImpl(MetricImpl):
         """
         Used in extensions
         """
-
-    def convert_db_value(self, value: any) -> any:
-        return value
 
     def get_short_description(self) -> str:
         return self.type
--- a/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
+++ b/soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import math
 import reprlib
 from abc import ABC
 from dataclasses import dataclass
@@ -15,6 +14,7 @@ from soda_core.common.consistent_hash_builder import ConsistentHashBuilder
 from soda_core.common.data_source_impl import DataSourceImpl
 from soda_core.common.exceptions import InvalidRegexException, SodaCoreException
 from soda_core.common.logs import Logs
+from soda_core.common.number_conversions import is_finite_number
 from soda_core.common.soda_cloud import SodaCloud
 from soda_core.common.sql_dialect import *
 from soda_core.common.yaml import ContractYamlSource, DataSourceYamlSource, SodaCloudYamlSource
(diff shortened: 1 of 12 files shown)


soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
```diff
@@ -1608,6 +1610,33 @@ def __eq__(self, other):
...
+
+    def _convert_db_value_to_number(self, value: any, source: str, hint: str) -> Optional[Number]:
+        """Read a warehouse value as a number, or skip it.
+
+        Numbers pass through unchanged, and numeric text, as produced by a CAST to a string type,
+        is parsed. Anything else is skipped with an error: the value becomes None, so the check
+        using it is not evaluated. Soda Cloud types check values as numbers, and a single value it
+        cannot parse rejects the results of the whole scan.
+        """
+        if value is None or isinstance(value, Number):
```
Niels-b: **Non-finite floats still take down the whole upload.**

`math.isfinite` is applied only on the string branch below (line 1632). A native `float('nan')` / `float('inf')` — or `Decimal('NaN')` — satisfies `isinstance(value, Number)` here and returns unchanged, and `_skip_non_numeric_threshold_value` (`soda-core/src/soda_core/check_collections/base.py:61`) lets it through for the same reason.

Reachable on the paths this PR touches: Spark/Databricks `double('NaN')`, BigQuery `IEEE_DIVIDE(0, 0)`, Snowflake `'NaN'::FLOAT` — and `avg` / `max` over a double column holding one of those propagates it.

The consequence is the failure this PR exists to prevent, with worse diagnostics than the 400. soda-core pins `requests>=2.32.3` (`soda-core/pyproject.toml:12`), and requests serializes the command body with `complexjson.dumps(json, allow_nan=False)`, so it raises `InvalidJSONError` before the request ever leaves. The bare `except Exception` in `_execute_cqrs_request` (`soda_cloud.py:1402-1407`) turns that into a `logger.critical` and a `None` response: the whole `sodaCoreInsertScanResults` is lost, with no 400 and nothing naming the field.

Two lines to close it — gate the `Number` early return on finiteness here and in `base.py`. `math.isfinite` returns `False` for `Decimal('NaN')` / `Decimal('Infinity')` and `True` for `bool`, so it composes with what is already there.

If you would rather not widen the PR, this belongs under **Not in this PR**: the description currently says "`nan`/`inf` text ... log an error", which reads as if nan/inf is covered generally.

Niels-b (approved):
Looks good, Claude still had one finding worth fixing I think.

soda-core/src/soda_core/contracts/impl/contract_verification_impl.py
m1n0: Right, and the chain holds end to end: requests 2.33 does `complexjson.dumps(json, allow_nan=False)`, `_execute_cqrs_request` swallows the `InvalidJSONError`, and DuckDB hands back `nan` for `AVG`/`MAX` over a double column with one NaN in it.

Fixed in 277cfdea, with one change to the suggestion. A bare `math.isfinite` raises on `Decimal('sNaN')` and on ints past float range, and in `base.py` that runs for every check, so it would swap a skip for a crash. Both spots go through `is_finite_number` now, which treats those as not finite.

The mock also never serialized request bodies, which is why this was invisible to the suite. It now serializes like requests does. Nothing existing trips it, and the new NaN test fails on the previous commit. Description updated too.

PR #2859, opened 2026-09-14
feat(cloud): send the dataset's columns in the scan results metadata

mivds:
## Problem

Soda Cloud fills a dataset's column list from the `metadata` bucket of the scan results payload. Contract verification never sends that bucket, so a dataset whose only activity is contract verification shows no columns in Soda Cloud. Datasets that reach v4 without an onboarding run (for example after a migration) stay empty forever.

## Change

When a contract has a `schema` check, the engine already reads the dataset's full column list from the warehouse. Verification now forwards those columns in the payload's `metadata` bucket: one entry for the contract's dataset, with the column name and the lowercase source data type per column. No extra warehouse query is made. Contracts without a schema check, and schema checks that did not get their columns, send no `metadata` at all, so Cloud never receives an empty list it could mistake for a schema without columns.

The result object gets a generic `dataset_columns` field that the payload builder reads, so other check collections can supply columns later without touching the wire code. The Cloud side ingests the bucket behind a feature flag; older Cloud versions accept and ignore the field as they already do for v3 scans.

Ref SCS-1415.

## Testing

Unit tests for the payload shape, the qualified-name equality with the checks, the empty and missing type cases and multi-result batches. Integration tests run a contract with and without a schema check against DuckDB and assert the captured payload. Verified end to end against a local Soda Cloud and a Postgres data source: a dataset went from no columns to the full typed column list after one verification, and a dropped column was marked deleted on the next run.

```customer-facing-release-note
Contract verifications with a schema check now populate the dataset's column list in Soda Cloud.
```


soda-core/src/soda_core/common/soda_cloud.py
```diff
@@ -1850,6 +1850,42 @@ def _build_token_usage_dicts(contract_verification_result: ContractVerificationR
...
+        if dataset_qualified_name in seen_dataset_qualified_names:
+            continue
+        schema: list[dict] = [
+            # The bare type name, already lowercased by SqlDataType, without the
+            # precision/length parameters: the same columnName / sourceDataType spelling
+            # capture-schema uses. The primary-key flag it also sends is left out here on
+            # purpose — contract verification has no reason to restate it.
+            {"columnName": column.column_name, "sourceDataType": column.sql_data_type.name}
+            for column in result.dataset_columns
+            if column.sql_data_type and column.sql_data_type.name
```
m1n0: Dropping a column with no type name soft-deletes it in Cloud, since the backend does a full column sync and anything missing from `schema` gets marked deleted. I'd send it name-only instead. `updateColumnSourceDataType` ignores a null type, so the column survives and keeps whatever type it already had. Rare, because the schema query always builds a `SqlDataType`, but this fails the wrong way.

m1n0 (commented):
Cleaner than what I had in #2782, closing mine in favor of this.

m1n0 (approved):
Comment above is non-blocking.

soda-core/src/soda_core/common/soda_cloud.py
mivds: Checked both sides. The full column sync is as you describe, so leaving a column out does soft-delete it. But name-only is not safe either: `SchemaElementCoreDTO.sourceDataType` is `@NotEmpty`, and the only reason that does not reject the batch today is that `MetadataCoreDTO.schema` has no `@Valid`. One typeless column would then fail the whole scan results upload.

The case is also unreachable: every producer of the schema check's actual columns builds a `SqlDataType` (the dialect metadata query, Databricks, Salesforce). The `Optional` on `ColumnMetadata.sql_data_type` is for the contract's expected columns, which never flow in here.

So the column stays out, but it is now logged as an error naming the column and the dataset. Results still upload on every path and the scan is not quietly green. Raising instead would drop the results, and on a multi-contract run the logs too. Pushed in c4d59550.

mivds: Went with your suggestion after all: a column without a type name now goes out as `{"columnName": ...}` with no `sourceDataType`, so Cloud keeps whatever type it has. Profiling already sends through this flow, so the field is not going to turn strict. Force-pushed as d29d406d, replacing my earlier commit.

PR #2860, opened 2026-09-14
test(snapshot): let the metadata pin target any impl of the helper's type

mivds:
## Problem

Snapshot recordings that use a BigQuery diagnostics warehouse fail their replay verification since #2854. The recording holds an `INFORMATION_SCHEMA.TABLES` query at operation 2; the replay issues the `COLUMNS` query in its place. The recon cell fails on the same cause: it finds no `rrd_` table.

Since #2854 the BigQuery tables query resolves its region from dataset metadata, a REST call on the client. The snapshot wrapper records and replays SQL only, which is why that PR added `_snapshot_pin_metadata_lookups` to the test helper. The pin only reaches the helper's own data source impl.

The diagnostics warehouse is a different impl. The scan builds it from a yaml file, and the test suite wraps its connection separately. Nothing pins it, so the wrapper's attribute proxy opens a real connection for the lookup. During recording the diagnostics dataset exists and the region resolves. During replay the dataset has been cleaned up, live BigQuery reports it missing, no region is in scope and the TABLES query is skipped. The log shows it right before the mismatch:

```
BigQuery dataset 'soda_diagnostics_bigquery_20260914_3b5de693' does not exist, so no region is in scope for it.
```

Re-recording does not help. The recordings are correct; replay diverges because it asks live BigQuery about a dataset that is gone.

## Change

`_snapshot_pin_metadata_lookups` takes the impl it applies to. The two existing callers pass the helper's own impl, so their behaviour does not change. The code that wraps the diagnostics warehouse connection can now pin that impl through a helper of the matching type.

Setting `location` in the diagnostics warehouse yaml would also skip the lookup, but the pin's own comment explains why that was rejected for the source side, and the same holds for the target: the model is frozen, and it also pins the real client.

## Testing

`soda-bigquery/tests/unit` passes (68 tests). The pinning test pins an impl the helper does not own and asserts the pinned region comes back without a client call. Not yet verified end to end: that needs a snapshot recording with a bigquery diagnostics warehouse run against this branch.


soda-bigquery/tests/unit/test_bigquery_region_resolution.py
```diff
@@ -157,15 +157,29 @@ def test_pinned_region_resolution_touches_no_bigquery_client(self, monkeypatch):
...
 
-        helper._snapshot_pin_metadata_lookups()
+        helper._snapshot_pin_metadata_lookups(impl)
 
         assert impl.regions_in_scope(project_id=PROJECT) == ["europe-west1"]
         assert impl.region_for_dataset(PROJECT, "dataset_a") == "europe-west1"
         impl.data_source_connection.client.list_datasets.assert_not_called()
         impl.data_source_connection.client.get_dataset.assert_not_called()
 
+    def test_pins_an_impl_the_helper_does_not_own(self, monkeypatch):
```
Niels-b: Unsure if this test really tests anything new 🤔 .

soda-tests/src/helpers/data_source_test_helper.py
```diff
@@ -1201,13 +1201,16 @@ def _snapshot_passthrough_queries(self) -> dict:
...
-        reads over its client's REST API bypasses it entirely, so in replay mode it
-        would open a real connection and hit the network. Override in subclasses to
-        make those answers deterministic in both snapshot modes.
+        The snapshot wrapper records and replays SQL only. Metadata a data source reads over its client's
+        REST API bypasses it entirely, so in replay mode it would open a real connection and hit the
+        network. Override in subclasses to make those answers deterministic in both snapshot modes.
+
+        The impl is a parameter because the helper's own impl is not the only one that gets a snapshot
+        connection: a scan builds the diagnostics-warehouse target from a yaml file, and the DWH snapshot
+        interceptor in soda-extensions pins that impl through the target's helper.
```
m1n0: soda-core is public, so I'd keep the extensions internals out of this. Something like "other impls get a snapshot connection too, e.g. one a scan builds from a yaml file" says the same. The comment in `test_pins_an_impl_the_helper_does_not_own` has the same issue.

m1n0 (approved):

Niels-b (approved):

soda-bigquery/tests/unit/test_bigquery_region_resolution.py
mivds: Agreed. The existing test above already covers it: the helper is built with `__new__` and has no impl of its own, so it already proves the pin works on an impl the helper does not own. Removed.

soda-tests/src/helpers/data_source_test_helper.py
mivds: Reworded the docstring along those lines. The test with the same comment is gone (see the other thread), and the commit message and PR body no longer mention it either.

PR #2863, opened 2026-09-15
fix(discovery): skip Databricks metric-view materializations

mivds:
## Description

**Problem.** Databricks materializes metric views through a managed Lakeflow pipeline whose backing objects are named `__materialization_mat_<pipeline id>___metric_view_mat_...` ([Databricks docs](https://docs.databricks.com/aws/en/uc-semantics/metric-views/materialization)). They are created in the customer's own schemas, so `soda data-source discover` lists them like any other table and reports them to Soda Cloud as discovered datasets. One customer's daily discovery picked up ~2,900 of them in a single schema, which dragged their "Onboarded vs Discovered" coverage widget from 77% to 43%. Onboarding rules cannot help: they decide what gets onboarded, not what counts as discovered.

**Change.** Discovery skips these objects for Databricks data sources. The existing system filter works per schema (`is_system_schema`), which cannot catch objects sitting next to the customer's tables, so the dialect gains a name-level counterpart, `is_system_table_name`, applied to every discovered object type. The base dialect flags nothing, so other data sources are unaffected. Hive metastore catalogs inherit the rule.

Same fix for the v3 engine: sodadata/soda-library#796.

Refs: SCS-1451

## Checklist

- [ ] Documentation updated where relevant (no user-facing option changes; docs unaffected)
- [x] PR title follows the conventional commit format (e.g. `fix(sqlserver): ...`)

**Testing.** Unit tests for the dialect rule and for the discovery filter (table and view objects); full `soda-tests/tests/unit` suite green locally.

```customer-facing-release-note
Databricks data sources no longer discover metric-view materialization tables.
```


m1n0:
One question, applies to soda-library#796 too. This stops new materialization tables from being reported, but what happens to the ~2,900 that are already discovered in the customer's org? If Cloud keeps datasets that discovery no longer reports, their coverage widget stays at 43% after the upgrade and we still need a cleanup on the Cloud side. Do you know if the backend drops them, or should we plan a one-off cleanup for SCS-1451?

m1n0 (approved):
Looks good. The filter sits in discover_qualified_objects so it also reaches the columns fallback and the failed rows DWH lookup, both harmless since we never name our own tables like that. Cleanup question is in the comment above, not blocking.

mivds:
> One question, applies to soda-library#796 too. This stops new materialization tables from being reported, but what happens to the ~2,900 that are already discovered in the customer's org? If Cloud keeps datasets that discovery no longer reports, their coverage widget stays at 43% after the upgrade and we still need a cleanup on the Cloud side. Do you know if the backend drops them, or should we plan a one-off cleanup for SCS-1451?

Discovered datasets that no longer exist in the warehouse are automatically removed on the next discovery scan. If they've already been onboarded they stay, but that shouldn't be an issue in this case.

PR #2864, opened 2026-09-18
feat(soda_cloud): send a product-token User-Agent on every Soda Cloud request

mivds:
## Summary

Soda Cloud reads the `User-Agent` product tokens (RFC 9110 section 10.1.5) to tell which client sent a request. soda-core only set the header on the command/query path; the log batch uploads and the REST GETs built their own header dict and went out as `python-requests/x.y`. The header also pinned a stale constant, so it reported `SodaCore/4.0.0.b1` on every release.

This PR:

- Adds `soda_core.common.user_agent` with `user_agent()` and `register_user_agent_product(name, version)`. soda-core is always the first token; packages that ride on soda-core append theirs, so a request reads e.g. `SodaCore/4.24.1 SodaExtensions/4.24.1`.
- Reports the maintained `SODA_CORE_VERSION` from `soda_core.__version__` instead of the stale `soda_core.common.version` (removed; nothing else used it).
- Adds `SodaCloud.request_headers(headers)` which merges the default headers under request-specific ones, and uses it on every request site that sets its own `Authorization`/`Content-Type`. `SodaCloud.headers` is now a property so products registered after construction are picked up.

Motivation: SCS-1465. soda-extensions' failed-row sample upload hits the shared `/api/scan/upload` endpoint, and the backend cannot tell it apart from an old soda-library client. The companion soda-extensions PR registers `SodaExtensions/<version>`; the backend PR reads the tokens. Behaviour of the requests themselves is unchanged.

Companion PRs: soda-extensions (registers its token), soda (parses all tokens, uses it in the upload gate).

## Test plan

- [x] New `soda-tests/tests/unit/test_soda_cloud_user_agent.py`: default token, registration and replacement, invalid tokens, header merging, log-batch and REST GET requests carry the header
- [x] Existing `test_soda_cloud.py` expectations point at `SODA_CORE_VERSION`
- [x] Full `soda-tests/tests/unit` suite passes locally (1501 passed)
- [ ] CI green


m1n0 (commented):
Would it make sense to make this overridable so that commercial extensions can override?

m1n0 (approved):
Every Cloud call carries the header now, including log batches and REST GETs. `headers` is now a property returning a fresh dict, so `soda_cloud.headers["X"] = ...` silently does nothing. No caller does it today, but a docstring line would save someone. Needs to be released before extensions#622 can merge.

PR #2865, opened 2026-09-20
fix(synapse): keep the paginator column cache on the impl, reset per test

mivds:
## Problem

`SynapseSqlDialect` caches a table's column list for its ROW_NUMBER paginator, so a `rows_diff` scan asks `information_schema.columns` once instead of once per page. The cache lives on the dialect for as long as the dialect does. In a scan that is one scan. In the test suites it is the whole pytest session: the data source helper is session-scoped, and so are its impl and dialect.

So only the first test that paginates a given table issues the columns lookup. Every later test on the same table skips it. Snapshot replay then depends on test order: a snapshot recorded by a test that ran first expects the lookup, and replaying it after another test on the same table fails with a mismatch at that operation.

This is what has kept the soda-extensions `cross-source summary` job red on main since 2026-09-15 (run [34954445302](https://github.com/sodadata/soda-extensions/actions/runs/34954445302)). The narrowed `target=bigquery` re-recording ran `test_dwh_cross_source_recon[synapse-bigquery]` first, so it recorded the lookup. The full synapse row replays that cell after `synapse-athena`, which has already warmed the cache. Every PR since inherits the failure.

## Change

Two things.

The cache moves from the dialect to `SynapseDataSourceImpl`, inside the resolver that issues the query. Dialects are stateless SQL renderers everywhere else in the codebase, and the other caches of warehouse-derived facts (BigQuery's dataset region, the connection's session timezone) already live on the impl or the connection. The dialect just calls its resolver.

The synapse test helper clears that cache in `test_method_ended`, the same hook the base helper uses to reset its table name cache so each test's snapshot is self-contained. Every test now issues the same metadata queries whatever ran before it, so a narrowed re-record of any cell replays in the full row. Scans never call the reset; they build a fresh impl.

## Caveat for rollout

Existing synapse snapshots that were recorded with a warm cache (a `rows_diff` test on a table another test paginated earlier in the session) lack the lookup and will mismatch once this is on main, until they are re-recorded. In soda-extensions that means the synapse rows of the cross-source lane and the reconciliation suite. Plan: merge, then dispatch `nightly-cross-source-record.yaml` with `source=synapse, target=all` and the nightly snapshot recording for synapse from main. The regular weekly and nightly recordings would also catch up on their own.

## Testing

`soda-synapse/tests/unit`: 19 passed. Two new tests on a real impl built from the helper's real config, with only `execute_query` stubbed and a real `QueryResult` fed through the real metadata parser: the impl resolves a table's columns once (and each table separately), and `clear_column_names_cache()` makes it query again. The wiring in `test_method_ended` is guarded by the snapshot replay lanes themselves, which is where this bug surfaced.


m1n0 (approved):
No production change that I can find. The cache was already per instance, same key, same lifetime, and nothing in extensions touches it. Please dispatch the ext synapse re-record right after this merges instead of waiting for the nightly, otherwise more ext lanes go red.

Niels-b (commented):
Not sure about this one. I think it has something to do with the snapshotting system in general. This fix touches production code, which I'm not a fan of to do just for the snapshot system.

Also, we have no way of testing this now as the snapshots are re-recorded as of this morning.

mivds:
> Not sure about this one. I think it has something to do with the snapshotting system in general. This fix touches production code, which I'm not a fan of to do just for the snapshot system.
> 
> Also, we have no way of testing this now as the snapshots are re-recorded as of this morning.

This PR does 2 things (maybe I should have split them):

1. Clear the synapse column cache in between tests
   * Testing only change
   * Indeed related to snapshots and intended to fix https://github.com/sodadata/soda-extensions/actions/runs/34954445302/job/104333009319
   * Snapshots currently have a dependency on tests always being executed in the same order. This PR fixes that
2. It moves the per-dataset cache from dialect to the data source object
   * This is purely a refactor

I had Claude do 2 while working on 1, simply because I noticed it from its explanation. The dialects are generally stateless and just generate SQL. The Synapse version was maintaining a cache and that made it generate different SQL depending on its state. Might have been better suited as a separate PR

Niels-b:
If it's code that's used in testing only; all good 👍 . 

Just in general; for snapshots I tried to touch as little "production" code as possible, so seeing changes in a `DataSourceImpl` raises some flags for me 😅 . 

Thank you for fixing this and looking into it!

Niels-b (approved):

PR #2866, opened 2026-09-21
feat(mariadb): add MariaDB v4 data source support (PLATL-716)

paulteehan:
Registers `soda-mariadb`'s test helper in `DataSourceTestHelper.create`. Four lines, mirroring the `mysql` arm directly above.

**Pairs with sodadata/soda-extensions#623**, which adds the `soda-mariadb` package. That dispatch is a hardcoded if/elif chain, so without this entry the helper is unreachable from the shared suites — including the cross-source DWH matrix, which needs MariaDB as both source and target.

## Merge order

**This one first, or together.** On its own it is inert: nothing in soda-core imports `soda_mariadb`, and the new branch is only reachable when `TEST_DATASOURCE=mariadb`, which no soda-core lane sets.

## Note for whoever picks this up

This branch is **named identically to #623's head branch**, so soda-extensions CI resolves it automatically and needs no `SODA_CORE_BRANCH=` marker in that PR's body.

That convenience has a sharp edge worth knowing: extensions CI then builds against *this* branch, whatever state it is in. When it drifted 5 commits behind soda-core main it stopped carrying `soda_core.common.user_agent`, and **every lane on #623 failed** on that import until it was rebased. Keep it current with main while #623 is open.


m1n0 (approved):
Import target exists at ext head (`soda_mariadb/test_helpers/mariadb_data_source_test_helper.py:8`), no core lane can reach the arm, soda-tests only. Merge before #623.

PR #2867, opened 2026-09-21
fix(metric): warn when a metric query result is not scalar

m1n0:
## What

`MetricQuery.execute` reads a metric query result as `rows[0][0]`. Extra rows and extra columns were dropped without a word. It now logs one warning when the result is not a single cell:

```
Metric query returned 3 rows, expected 1 and 2 columns, expected 1. Using the first column of the first row. Which row comes first is nondeterministic without an ORDER BY.
Metric query:
SELECT ...
```

The row part and the column part each appear only when they apply, and the ORDER BY sentence only with extra rows. The SQL is included the same way the no-rows warning next to it does.

## Why

A query that forgot its aggregate or its `LIMIT 1` still evaluates. It uses whichever row the warehouse returns first, and without an `ORDER BY` that row can change between runs, so a check can flip between pass and fail with nothing in the logs to explain it.

`MetricQuery` also runs the reconciliation `metric_diff` query mode in soda-extensions, so that path gets the warning with no change there.

## Behaviour change

None beyond the log line. The value still comes from the first cell, and no check outcome changes. The column count is read from the first row, not from cursor metadata.

## Tests

Multi-row, multi-column and both at once, at two levels:

- unit, in `test_metric_check.py`: the exact warning, one warning only, the SQL in it, the value from `rows[0][0]`, and no warning on a scalar result
- integration, in `test_udf_metric_check.py`: the check still passes on the first cell and the warning reaches the verification logs

Follows #2858 for guarding this spot. Linear: DET-193

## Checklist

- [ ] Documentation updated where relevant
- [x] PR title follows the conventional commit format (e.g. `fix(sqlserver): ...`)

```customer-facing-release-note
Metric checks now warn when their query returns more than one row or column, since only the first value is used.
```


Niels-b (approved):
Seems like a good thing to add, to make users more aware.

Not sure about the log level. Could also be `info`. No strong opinion though.

mivds (approved):

PR #2870, opened 2026-09-22
test(schema): let the metadata tests run on a source with its own columns

Niels-b:
## Description

The nightly Salesforce live-org lane in `soda-extensions` has been failing since #2859 merged:

```
FAILED soda-tests/tests/integration/test_schema_check.py::test_schema_check_sends_the_dataset_columns_as_metadata
FAILED soda-tests/tests/integration/test_schema_check.py::test_metadata_carries_the_whole_table_not_the_declared_columns
```

([nightly run](https://github.com/sodadata/soda-extensions/actions/runs/35682382255))

The two new tests assert the dataset column list that now travels to Cloud in the `metadata` bucket. Both verify a contract over the fixture's six columns, then compare the bucket against that same set as a literal — which assumes the dataset has exactly the columns the fixture declared, in declaration order.

A Salesforce object has neither. Describe reports every object's nine standard fields (`Id`, `OwnerId`, `IsDeleted`, `Name`, `CreatedDate`, …) alongside the fixture's own, in the org's order, so the schema check fails on nine extra columns *and* on column order — `assert_contract_pass` raises before a single metadata assertion runs. The `soda-extensions` nightly clones soda-core `main` unpinned, so both tests hit that lane the night they merged.

**To be clear: the feature itself works correctly on Salesforce.** The connector always builds a `SqlDataType` with a clean lowercase, parameter-free name (`id`, `reference`, `string`, `double`, `datetime`, `date`), so `sourceDataType` is well-formed and the new "no data type name" error branch cannot fire. Only the tests' preconditions are unsatisfiable there.

### What changed

The `metadata` bucket stays the subject of both tests — the schema check's own strictness is covered by the rest of the module.

- Both contracts declare `allow_extra_columns` + `allow_other_column_order`, on every data source, so a dataset carrying columns no fixture declared (in the source's own order) no longer fails the contract before the metadata assertions run.
- The literal expected column sets are replaced by an assertion that the bucket mirrors what the schema check measured — column names *and* type names, from `SchemaCheckResult.actual_columns`. That is the stronger statement of the two: it holds on any source, and it pins the type spelling that was previously only checked loosely for lowercase/no-parens.
- `test_metadata_carries_the_whole_table_not_the_declared_columns` keeps its point via `len(metadata_column_names) > 2` — more than the two the contract declared, without pinning how many more.

The alternative was adding both tests to the Salesforce exclude manifest, which would have left the new Cloud metadata path with no Salesforce coverage at all. This way it's covered on the source where it's most interesting: 15 actual columns against 6 declared.

### Testing

- **Postgres** — full `test_schema_check.py`: 15 passed, 1 skipped (pre-existing Oracle skip).
- **Salesforce, live org** — the two tests pass. Confirmed genuinely against the org (Metadata API traffic to `SODATEST_schema_401c8b9c__c`, `Found 15 actual columns`, `CheckOutcome.PASSED`), not skipped.

## Checklist

- [x] Documentation updated where relevant — n/a, test-only
- [x] PR title follows the conventional commit format


mivds (approved):
Approving, but I think we could simplify this by using the salesforce path for all data sources. Point of the tests is to validate the `metadata`, not the schema check itself. Having `{allow_extra_columns: true; allow_other_column_order: true}` is perfectly fine

Niels-b:
Ok, thanks for the feedback! I didn't want to touch the original flow too much, but that simplifies it indeed!

Niels-b:
Done in a100c680 — the tolerances now apply on every data source and the gate is gone, along with the f-string contracts it needed. What's left in place of the literal column sets is the assertion that the bucket mirrors what the schema check measured (names and type names), which is the stronger statement anyway.

Re-ran after the change: Postgres full module 15 passed / 1 skipped, and both tests pass against the live Salesforce org.

